PR #20, opened 2025-10-31
Fix formatting and update YAML instructions in kernels.md

jnwei:


seth-hazelnutt:
Addressed @jessica-fletchirrel's review feedback.

PR #24, opened 2025-11-03
Fix `presets` field in ModelUpdate

### LINKED ISSUE #23: Without evoformer_attn/evoformer_attn.so  and model.aux_heads.pae.linear.weight".
Dear team，

I'm following the deployment tutorial and encountered two errors when running inference:

First Error:
After l installation, when running:

run_openfold predict --query_json=examples/example_inference_inputs/query_ubiquitin.json

I got the  extension error:

[ERROR] Failed for query_id(s) query1: **/home/yl/.cache/torch_extensions/py310_cu126/evoformer_attn/evoformer_attn.so**
Second Error:
Based on community suggestions, I tried using the runner.yml config:

run_openfold predict --query_json=examples/example_inference_inputs/query_ubiquitin.json --runner_yaml=runner.yml
But now I'm getting a different error about unexpected keys in the state_dict:

text
**Unexpected key(s) in state_dict: "model.aux_heads.pae.layer_norm.weight", "model.aux_heads.pae.layer_norm.bias", "model.aux_heads.pae.linear.weight"**
My runner.yml content:

model_update:
  preset: 
    - predict
    - pae_enabled  # <-- Required for OpenFold3 current model release which has PAE output
    - low_mem 
  custom:
    settings:
      memory:
        eval:
Environment:
 Linux with A100 500G
Python 3.10
CUDA 12.6
Any insights, workarounds, or pointers to the correct configuration would be immensely helpful. Thank you so much for your time and support!

jnwei:
The documentation and examples for the cuequivariance kernel had a typo for `presets` in the `model_update` section. Because of the misspelling, these preset selections were ignored.

The `ModelUpdate` config Model is updated to forbid extra fields, which will cause an error for misspelled fields in the future.

The miniforge version is also pinned to 25.3.1 in this PR to fix the CI testing

Fixes #23

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ ENV PATH=/opt/conda/bin:$PATH
 
 # Copy and install dependencies with aggressive cleanup
 COPY environments/production.yml /opt/openfold3/environment.yml
-RUN mamba env update -n base --file /opt/openfold3/environment.yml --prune \
+RUN mamba env update -n base --file /opt/openfold3/environment.yml \
     && mamba clean --all --yes \
     && conda clean --all --yes
 


jessica-fletchirrel (approved):
Looks good.

Verified the following:

**Typo fix (preset -> presets):** The field rename on ModelUpdate is correct and complete. All example YAMLs (cuequivariance.yml, affinity.yaml, low_mem.yml, all training YAMLs) and docs/source/kernels.md already use `presets`. The validator defaults are consistent with the new field name.

**extra="forbid" guard:** Adding PydanticConfigDict(extra="forbid") to ModelUpdate is the right defensive measure — users who still have the old `preset:` key in their YAML will now get a clear validation error rather than silent no-op behavior.

**Config logging move (setup vs run):** Moving _log_experiment_config() and _log_model_config() from run() to setup() in InferenceExperimentRunner is a correctness improvement. Config is now written before any inference work starts, so it is available even if run() fails.

**experiment_config.json write consolidation:** Removing the json.dump(expt_config.model_dump_json(...)) call from run_openfold.py eliminates a double-serialization bug (the old code wrote a JSON-encoded string into a JSON file). The canonical write is now _log_experiment_config() in InferenceExperimentRunner, which uses model_dump_json + write_text correctly. The `import json` removal from run_openfold.py is safe — json is still imported in experiment_runner.py where it is used for the WandB config store.

**Miniforge pin:** Pinning to 25.3.1-0 instead of latest is a standard CI stability practice.

**Build/lint:** ruff check passes with no issues. All 15 TCE tests pass.

PR #28, opened 2025-11-06
Support dummy sequence MSA

haewonc:
Support dummy main MSA consisted of query sequence
Figure: Complex with and without dummy MSA (PDB: 3WBM)
no_vs_dummy

### DIFF
--- a/environments/development.txt
+++ b/environments/development.txt
@@ -1,4 +1,4 @@
-pytest
+pytest==8.4.2  # until v9.0.1 is available due to unittest bug: pytest-dev/pytest#13895
 ruff
 awscli
 aria2


openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1044,3 +1044,39 @@ def preprocess_colabfold_msas(
...
+
+def augment_main_msa_with_query_sequence(
+    inference_query_set: InferenceQuerySet,
+    compute_settings: MsaComputationSettings,
+) -> InferenceQuerySet:
+    output_directory = compute_settings.msa_output_directory
+    for query_name, query in inference_query_set.queries.items():
+        for chain in query.chains:
+            if (
+                chain.molecule_type == MoleculeType.PROTEIN
```
gnikolenyi, line 1057: Instead of hard-coding the moltypes, could you make them dependent on MSASettings.moltypes (should be in projects.of3_all_atom.config.dataset_config_components), which we use elsewhere to determine which molecule types are expected to have MSAs.

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1044,3 +1044,39 @@ def preprocess_colabfold_msas(
...
+            ) and chain.main_msa_file_paths is None:
+                dummy_msa_file_path = (
+                    output_directory
+                    / "dummy"
+                    / f"{get_sequence_hash(chain.sequence)}.npz"
+                )
+                dummy_msa_file_path.parent.mkdir(exist_ok=True, parents=True)
+                dummy_msa = ">query\n" + chain.sequence
+                msas_preparsed = {"dummy": parse_a3m(dummy_msa).to_dict()}
+                np.savez_compressed(dummy_msa_file_path, **msas_preparsed)
```
gnikolenyi: This is nice and should be the default. Could you please also add an option to save the raw a3m msa files.

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1044,3 +1044,39 @@ def preprocess_colabfold_msas(
...
+                chain.main_msa_file_paths = [dummy_msa_file_path]
+
+                chain_ids = ",".join(chain.chain_ids)
+                warnings.warn(
+                    (
+                        f"Expected MSA file for chain {chain_ids} of "
```
gnikolenyi: nit: wording, maybe say "A dummy MSA with only the query sequence will be used for this chain."

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1044,3 +1044,39 @@ def preprocess_colabfold_msas(
...
+) -> InferenceQuerySet:
+    output_directory = compute_settings.msa_output_directory
+    for query_name, query in inference_query_set.queries.items():
+        for chain in query.chains:
+            if (
+                chain.molecule_type == MoleculeType.PROTEIN
+                or chain.molecule_type == MoleculeType.RNA
+            ) and chain.main_msa_file_paths is None:
+                dummy_msa_file_path = (
+                    output_directory
```
gnikolenyi: Could you please also update the documentation in the main inference doc in the MSA-free section and in the outputs section that this dummy dir will also be created?

openfold3/projects/of3_all_atom/config/dataset_config_components.py
```diff
@@ -100,6 +100,7 @@ class MSASettings(BaseModel):
         "nt_hits",
         "concat_cfdb_uniref100_filtered",
         "colabfold_main",
+        "dummy",
```
jnwei: nit: could you add a small comment explaining what `dummy` is used for? That way a user who is reading this out of context of this PR will know what this is for.

jnwei:
This is great Minji!

Could you please add a test for testing feature generation using the dummy msa file? I think you could do it by extending / copying this test: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_colabfold_msa.py#L225

openfold3/core/data/tools/colabfold_msa_server.py
haewonc: Oh, moltypes is attribute of MSASettings not MSAComputationSettings. 
So I think it's not accessible yet in the DataModule prepare_data level.
Is there any way to cleanly address this?
But I think it's okay to compute & save for both. Later if moltypes don't include RNA or PROTEIN, this field will be ignored. How do you think?

PR #38, opened 2025-11-12
Fix: Handle empty pdb70.m8 template file from ColabFold

### LINKED ISSUE #36: Queries with no template hits stump the pipeline
Hi Team,

Some queries get no template hits. However, Colabfold still includes an empty pdb70.m8 file in the tar.gz archive. The code at line 759-761 in openfold3/core/data/tools/colabfold_msa_server.py 

```python
template_alignments = pd.read_csv(
    self.output_directory / "raw/main/pdb70.m8", sep="\t", header=None
)
m_with_templates = set(template_alignments[0])
```

always attempts to read the template file causing a downstream pandas error:
`pandas.errors.EmptyDataError: No columns to parse from file`

The fix could be a check on whether pdb70 file contains data - if it's empty, find a way to handle that gracefully and alert the user that no templates will be used.
Will submit a PR to tackle this issue. Thanks!

qurat-ul-ain95:
**Problem**

The ColabFold server may include an empty `pdb70.m8` file in the tar.gz archive even when templates are not available or when no template hits are found.When the ColabFold API returns an empty `pdb70.m8` template file (0 bytes), the code fails with:
```
pandas.errors.EmptyDataError: No columns to parse from file
```

This occurs because the code at line 759-761 (before fix) unconditionally attempts to read the template file:
```python
template_alignments = pd.read_csv(
    self.output_directory / "raw/main/pdb70.m8", sep="\t", header=None
)
m_with_templates = set(template_alignments[0])
```
When the file is empty, pandas cannot parse it and raises `EmptyDataError`.

**Changes**
- Prevents pandas EmptyDataError when ColabFold returns empty template file by
1. Check if pdb70.m8 exists and has content before reading
2. Create empty DataFrame if pdb70.m8 empty
3.  Add defensive check to prevent accessing empty DataFrame
4. Log warning when templates are not available

Also, added a test `test_empty_m8_file_handling` in `test_colabfold_msa.py`.

**Solution**

1. *Verify that `pdb70.m8` exists and has content (> 0 bytes) before attempting to parse it.

2. If the file is empty or missing, create an empty DataFrame and empty set, allowing the pipeline to proceed without templates.

3. Add `len(template_alignments) > 0` check when processing templates to prevent accessing empty DataFrames.

4. Inform users when templates are not available so they understand why template processing is skipped. (Please see attached image).

You can use this query from casp16 to test no template return:
[T1201.json](https://github.com/user-attachments/files/23508718/T1201.json)

Alerts: 
image

image

closes #36


openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -709,10 +709,21 @@ def query_format_main(self):
...
+        if template_alignments_file.exists() and template_alignments_file.stat().st_size > 0:
+            template_alignments = pd.read_csv(
+                template_alignments_file, sep="\t", header=None
+            )
+            m_with_templates = set(template_alignments[0])
+        else:
+            # pdb70.m8 downloaded by Colabfold returned empty - No template alignments available 
+            # Create empty DataFrame with expected column structure (at least column 0)
+            # to match the structure when file is read with header=None.
+            logger.warning(f"Colabfold returned no templates for at least one of the sequences in the batch \
```
jnwei: In order for the `pdb70.m8` file to be empty, wouldn't all of the queries have to have no template alignments?

openfold3/tests/test_colabfold_msa.py
```diff
@@ -277,6 +285,52 @@ def test_features_on_multiple_queries_with_same_name(
                 "Expected all test sequences to be present in the mapping file"
             )
 
+    @patch(
+        "openfold3.core.data.tools.colabfold_msa_server.query_colabfold_msa_server",
+        side_effect=_construct_dummy_a3m,
+    )
+    def test_empty_m8_file_handling(
```
jnwei, line 328: Could we expand this test to include the call to `preprocess_colabfold_msas` and verify that the generated inference_query_set has the expected None / empty fields  in the Chain for `template_alignment_file_path` and `template_entry_chain_ids`?

jnwei (changes requested):
Overall this is great @qurat-ul-ain95! Thank you for your contribution. 

I have one suggestion for expanding the coverage of the test to verify that the updated has the correct fields `inference_query_set`

openfold3/core/data/tools/colabfold_msa_server.py
qurat-ul-ain95: ah yes, you're right. I'll change the wording to be more clear.

jnwei (approved):
LGTM! I'm working out some things on our CI testing infrastructure, I'll merge this in once I've restored the CI

PR #44, opened 2025-11-18
feat: enable cuequivariance kernels on the base image

avelinoapheris:
* Modify the base image to install the cuequivariance kernels. This also involve upgrading the base cuda version.
* Add checks to ensure that the cuequivariance kernels are not loaded when using a setup without GPUs.


docker/Dockerfile.cueq
```diff
@@ -35,7 +35,7 @@ RUN /opt/openfold3/scripts/install_third_party_dependencies.sh
 
 # Install the package
 WORKDIR /opt/openfold3
-RUN python3 setup.py install
+RUN pip install '.[cuequivariance]' --no-build-isolation && python setup.py install
```
jandom: Yeah, I've also found that even with --no-build-isolation, you **still** need to run `python setup.py install`

openfold3/core/model/layers/triangular_multiplicative_update.py
```diff
@@ -30,7 +30,10 @@
 from openfold3.core.model.primitives import LayerNorm, Linear
 from openfold3.core.utils.tensor_utils import permute_final_dims
 
-cueq_is_installed = importlib.util.find_spec("cuequivariance_torch") is not None
+cueq_is_installed = (
+    importlib.util.find_spec("cuequivariance_torch") is not None
+    and torch.cuda.is_available()
+)
```
jandom: Can we mutualize this into a helper function? it seems to be duplicated 2-3 times in this PR

Dockerfile
```diff
@@ -45,7 +45,7 @@ ENV TORCH_CUDA_ARCH_LIST="8.0;8.6;9.0"
 #     python3 -c "import deepspeed; print('DeepSpeed ops loaded successfully')"
 
 # Runtime stage - use devel image for full CUDA support
-FROM nvidia/cuda:12.1.1-cudnn8-devel-ubuntu22.04 AS runtime
+FROM nvidia/cuda:12.6.1-cudnn-devel-ubuntu22.04 AS runtime
```
jandom: This has me worried the most, we're adding multiple base build images support – is this strictly necessary for cuequivariance or just nice to have?

jandom (commented):
Great contribution @avelinoapheris – couple of questions/comments below

Many thanks!

openfold3/core/model/layers/triangular_multiplicative_update.py
avelinoapheris: Done :)

Dockerfile
avelinoapheris: I am afraid we do, we were getting crashes in our environment without it, but i see you added the option of changing the base image through env variables. I modified the Dockerfile to install or not cuequivariance using another env variable, and update the documentation explaining how to do it.

openfold3/core/utils/tensor_utils.py
jnwei, line 1: Thank you for adding this helper function to test for cuequivariance installation. 

Would it be possible to move `is_cuequivariance_available` to a new module `cueq_utils` in [`core.kernels`](https://github.com/aqlaboratory/openfold-3/tree/main/openfold3/core/kernels)

avelinoapheris: Done :)

docker/Dockerfile
```diff
@@ -71,7 +77,6 @@ ENV TORCH_CUDA_ARCH_LIST="8.0;8.6;9.0"
 #     python3 -c "import deepspeed; print('DeepSpeed ops loaded successfully')"
 
 # Devel stage - use devel image for full CUDA support
-ARG CUDA_BASE_IMAGE_TAG=12.1.1-cudnn8-devel-ubuntu22.04
```
jandom, line 74: good catch

openfold3/core/model/layers/attention_pair_bias.py
```diff
@@ -203,7 +203,9 @@ def forward(
         #  and expects batch and seq dims to exist
         #  Current reshape function only expects missing batch dim
         batch_dims = a.shape[:-2]
-        reshape_for_ds_kernel = use_deepspeed_evo_attention and len(batch_dims) == 1
+        reshape_for_ds_kernel = (
+            use_deepspeed_evo_attention or use_cueq_triangle_kernels
+        ) and len(batch_dims) == 1
```
jandom, line 208: This is good

docker/Dockerfile
```diff
@@ -61,7 +63,11 @@ RUN /opt/openfold3/scripts/install_third_party_dependencies.sh
 
 # Install the package
 WORKDIR /opt/openfold3
-RUN uv pip install --no-deps --editable .
+RUN if [ "$INSTALL_CUEQ" = "true" ]; then \
+        uv pip install --no-build-isolation --editable ".[cuequivariance]"; \
+    else \
+        uv pip install --no-deps --editable "."; \
+    fi
```
jandom, line 70: Sorry, this maybe a naive question – 
1) could we not just always install the cuequivariance and 
2) why is --no-build-isolation required? (separate issue)

avelinoapheris: 1. The main reason to use `pip install .[cuequivariance]` instead of just `pip install cuequivariance` is to avoid to have to have different requirements inside the dockerfile and in pyprojects.toml. There should be only one place for that, plus there are other requirements a part from the  like `cuequivariance` package, like `torch>=2.7`

2.  If i don't use `--no-build-isolation` pip  creates its own environment, and either does not recognise that the packages already installed by the current environment or it doesn't link the new packages correctly.

PR #45, opened 2025-11-18
fix: small fixes used when finetuning a model

avelinoapheris:
* Allow training without templates. The current setup will crash if the templates directories are not given, whoever we find use cases where we will like to train without them.
* Allows running a validation dataset whose cache do not have "token_count" or "use_metric" fields. This is useful if one wants to run .validate on the training dataset.
* Fix a bug that requires the "ema" weights to be present when loading a checkpoint manually. There are cases where we want to start from a public checkpoint with only the weights of the model.
*  Expose all the arguments for the lightning checkpoint callback.


openfold3/core/data/primitives/structure/template.py
```diff
@@ -200,7 +203,7 @@ def sample_templates(
     else:
         k = np.min([np.random.randint(0, l), n_templates])
 
-    if k > 0:
+    if (k > 0) and (template_cache_directory is not None):
```
jnwei: Could the parentheses around the individual statements be removed? I believe this will cause an issue for our ruff formatter.

openfold3/entry_points/validator.py
```diff
@@ -81,8 +81,12 @@ def _maybe_download_parameters(target_path: Path) -> None:
 class CheckpointConfig(BaseModel):
     """Settings for training checkpoint writing."""
 
+    model_config = PydanticConfigDict(extra="allow")
```
jnwei: Is `extra="allow"` required to support backwards compatibility with old versions of the code?

We generally prefer to set `extra="forbid"` to help catch instances where a field might be ignored if it is not recognized by the Model (e.g. in the case of a typo or a new field).

jnwei (changes requested):
Thank you for adding the fixes for supporting fine tuning and enabling computation of metrics on the training set.

One suggestion: would it be possible to add a test to check that the case where `dataset_config.dataset_paths.template_structure_array_directory` is None works as intended? Perhaps a test that is a variation of the tests that are used to test training dataset runner yamls would be suitable, e.g. https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_entry_points.py#L63

openfold3/entry_points/validator.py
avelinoapheris: Fair enough. I removed it, but added a couple of extra fields that are available in the `ModelCheckpoint` callback.

openfold3/core/data/primitives/structure/template.py
avelinoapheris: Done

openfold3/entry_points/validator.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from turtle import save
```
jnwei: This seems to be a stray addition? Can we remove it if it is unused.

openfold3/projects/of3_all_atom/config/dataset_configs.py
```diff
@@ -80,12 +80,13 @@ def _validate_exactly_one_path_exists(
...
         ):
             which_paths_exist = [p is not None for p in path_values]
-            if sum(which_paths_exist) != 1:
+            if sum(which_paths_exist) > 1:
                 existing_paths = [
                     p for p, b in zip(path_values, which_paths_exist, strict=True) if b
                 ]
                 raise ValueError(
-                    f"Exactly one path in set of {group_name} should exist."
+                    "If there is a template folder, "
```
jnwei: Rather than reuse the `_valiadate_paths` check here, I would suggest creating a separate validation for `_validate_template_paths` with the intended behavior, perhaps throwing a warning instead of an error message when the template path is not set.

openfold3/entry_points/validator.py
avelinoapheris: Yes, sorry, removed

openfold3/projects/of3_all_atom/config/dataset_configs.py
avelinoapheris: Good idea. I fixed it.

avelinoapheris: I also add a validation and test to check the correct extensions.

openfold3/entry_points/experiment_runner.py
```diff
@@ -393,7 +393,8 @@ def manual_load_checkpoint(self):
         self.lightning_module.load_state_dict(
             state_dict, strict=self.ckpt_load_settings.strict_loading
         )
-        self.lightning_module.ema.load_state_dict(ckpt["ema"])
+        if "ema" in ckpt:
+            self.lightning_module.ema.load_state_dict(ckpt["ema"])
```
jandom, line 397: This is good

PR #69, opened 2025-12-06
feat(inference): Add support for custom residue numbering (resolves #58)

### LINKED ISSUE #58: Allow to specify the number of the first residue of a chain
Something that I find really inconvenient in AF3 is the lack of an option to specify what number should be assigned to the first residue of a given chain (if the user does not want it to be the default "1"). I have tried convincing the AF3 team to introduce this but they did not want to do so, as in principle this can of course also be done using postprocessing scripts. Nonetheless, I do think that implementing this directly would make people's life easier, so hopefully there is room for such an enhancement in OF3?

Once again using the example examples/example_inference_inputs/query_multimer.json as a template, what I am suggesting is something like this:

```
{
    "queries": {
        "7cnx": {
            "chains": [
                {
                    "molecule_type": "protein",
                    "chain_ids": [
                        "A",
                        "C"
                    ],
                    "sequence": "MLNSFKLSLQYILPKLWLTRLAGWGASKRAGWLTKLVIDLFVKYYKVDMKEAQKPDTASYRTFNEFFVRPLRDEVRPIDTDPNVLVMPADGVISQLGKIEEDKILQAKGHNYSLEALLAGNYLMADLFRNGTFVTTYLSPRDYHRVHMPCNGILREMIYVPGDLFSVNHLTAQNVPNLFARNERVICLFDTEFGPMAQILVGATIVGSIETVWAGTITPPREGIIKRWTWPAGENDGSVALLKGQEMGRFKLG",
					"starting_residue_number:": "102"
                },
                {
                    "molecule_type": "protein",
                    "chain_ids": [
                        "B",
                        "D"
                    ],
                    "sequence": "XTVINLFAPGKVNLVEQLESLSVTKIGQPLAVSTGHHHHHHG",
					"starting_residue_number:": "22"
                }
            ]
        }
    }
}
```

which would generate PDBs where chains A and C would contain residues M102-G354, and chains B and D would contain residues X22-G63.

Thank you!!!

chodec:
This Pull Request implements full support for **custom residue numbering** in the inference output. This feature allows users to define specific residue numbering in the input JSON, resolving **Issue #58**.

### Summary of Changes

The primary goal was to allow users to define specific residue numbering in the input JSON, rather than relying on the default numbering starting from 1. This includes support for non-sequential numbers and PDB-style **insertion codes** (e.g., '103A').

The implementation required coordinated changes across three key modules:

1.  **Input Definition (`inference_query_format.py`):**
    * Added two optional string fields to the `Chain` class: **`starting_residue_number`** (for simple offset) and **`residue_ids`** (for explicit lists).
2.  **Data Processing (`inference.py`):**
    * Implemented logic that ensures **`residue_ids` takes precedence** over `starting_residue_number`. If a valid explicit list is provided, it is used; otherwise, a sequential list is generated based on the start number. The final list is stored in the data batch.
3.  **Output Writing (Post-processing in `writer.py`):**
    * Implemented the static method **`OF3OutputWriter._renumber_atom_array`**. This method executes *after* model inference but *before* writing the PDB/mmCIF file.
    * It uses **regular expressions (`re`)** to safely parse string IDs (e.g., separating `'103A'` into the integer ID 103 and the insertion code 'A').
    * The new IDs are applied directly to the Biotite **`AtomArray`'s `res_id` and `ins_code`** annotations. This ensures the output structure reflects the desired numbering without affecting core model calculations.

---

### Related Issues

Resolves: **#58**

---

### Testing and Validation

**Note on Testing:** Due to local environment configuration issues (missing model checkpoints), an end-to-end test run was not possible to perform.

However, the logic has been manually validated to ensure:

* **Priority and Consistency:** The implementation correctly prioritizes `residue_ids` and handles sequence length mismatch by defaulting to standard numbering (1, 2, 3...).
* **Parsing Robustness:** The regex parsing logic in `writer.py` correctly extracts insertion codes, which is critical for PDB compliance.

### DIFF
--- a/examples/example_inference_inputs/query_multimer.json
+++ b/examples/example_inference_inputs/query_multimer.json
@@ -8,7 +8,8 @@
                         "A",
                         "C"
                     ],
-                    "sequence": "MLNSFKLSLQYILPKLWLTRLAGWGASKRAGWLTKLVIDLFVKYYKVDMKEAQKPDTASYRTFNEFFVRPLRDEVRPIDTDPNVLVMPADGVISQLGKIEEDKILQAKGHNYSLEALLAGNYLMADLFRNGTFVTTYLSPRDYHRVHMPCNGILREMIYVPGDLFSVNHLTAQNVPNLFARNERVICLFDTEFGPMAQILVGATIVGSIETVWAGTITPPREGIIKRWTWPAGENDGSVALLKGQEMGRFKLG"
+                    "sequence": "MLNSFKLSLQYILPKLWLTRLAGWGASKRAGWLTKLVIDLFVKYYKVDMKEAQKPDTASYRTFNEFFVRPLRDEVRPIDTDPNVLVMPADGVISQLGKIEEDKILQAKGHNYSLEALLAGNYLMADLFRNGTFVTTYLSPRDYHRVHMPCNGILREMIYVPGDLFSVNHLTAQNVPNLFARNERVICLFDTEFGPMAQILVGATIVGSIETVWAGTITPPREGIIKRWTWPAGENDGSVALLKGQEMGRFKLG",
+                    "starting_residue_number": "100" 
                 },
                 {
                     "molecule_type": "protein",
@@ -16,7 +17,8 @@
                         "B",
                         "D"
                     ],
-                    "sequence": "XTVINLFAPGKVNLVEQLESLSVTKIGQPLAVSTGHHHHHHG"
+                    "sequence": "XTVINLFAPGKVNLVEQLESLSVTKIGQPLAVSTGHHHHHHG",
+                    "residue_ids": ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28A", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "40", "41", "42"]
                 }
             ]
         }
--- a/openfold3/__init__.py
+++ b/openfold3/__init__.py
@@ -24,11 +24,11 @@ from . import hacks  # noqa: F401
 if version.parse(gemmi.__version__) >= version.parse("0.7.3"):
     gemmi.set_leak_warnings(False)
 
-if importlib.util.find_spec("deepspeed") is not None:
-    import deepspeed
-
-    # TODO: Resolve this later
-    # This is a hack to prevent deepspeed from doing the triton matmul autotuning
-    # This has weird effects with hanging if libaio is not installed and can
-    # cause restart errors if run is preempted in the middle of autotuning
-    deepspeed.HAS_TRITON = False
+#if importlib.util.find_spec("deepspeed") is not None:
+#    import deepspeed
+# 
+#     # TODO: Resolve this later
+#     # This is a hack to prevent deepspeed from doing the triton matmul autotuning
+#     # This has weird effects with hanging if libaio is not installed and can
+#     # cause restart errors if run is preempted in the middle of autotuning
+#    deepspeed.HAS_TRITON = False
--- a/openfold3/core/runners/writer.py
+++ b/openfold3/core/runners/writer.py
@@ -20,6 +20,7 @@ from pathlib import Path
 
 import numpy as np
 import torch
+import re
 import torch.distributed as dist
 from biotite import structure
 from pytorch_lightning.callbacks import BasePredictionWriter
@@ -119,6 +120,70 @@ class OF3OutputWriter(BasePredictionWriter):
             make_ost_compatible=make_ost_compatible,
         )
 
+    @staticmethod
+    def _renumber_atom_array(
+        atom_array: structure.AtomArray,
+        custom_residue_ids: list[str]
+    ) -> structure.AtomArray:
+        """
+        Applies custom residue numbering to the Biotite AtomArray using the 
+        list of string IDs provided by the user (via the feature dict).
+
+        It handles the splitting of complex PDB IDs (e.g., '102A' -> res_id=102, 
+        ins_code='A') and maps them from residue-level to atom-level annotations.
+        """
+        
+        # Get indices where each residue starts in the AtomArray (e.g., [0, 2, 4, 6, 8]).
+        residue_start_indices = structure.get_residue_starts(atom_array)
+        
+        # Initialize new annotation arrays for all atoms.
+        arr_len = len(atom_array)
+        
+        # IMPORTANT: Initializing with zeros is acceptable, as all elements 
+        # must be overwritten by the loop for a successful renumbering.
+        new_res_id = np.zeros(arr_len, dtype=np.int32) 
+        new_ins_code = np.full(arr_len, '', dtype=object) 
+
+        # Pre-calculate end indices for slicing. The end index of residue N is
+        # the start index of residue N+1, or the total array length for the last residue.
+        # Example: [0, 2, 4, 6, 8] -> [2, 4, 6, 8, 10]
+        end_indices = np.append(
+            residue_start_indices[1:], 
+            arr_len                    
+        )
+        
+        # Iterate over all residue start indices (i is the residue index, not atom index).
+        for i in range(len(residue_start_indices)):
+            
+            if i >= len(custom_residue_ids):
+                # Safety break: Should be prevented by featurizer
+                break
+            
+            # Start and End indices for the atom array slice corresponding to this residue.
+            start_index = residue_start_indices[i] 
+            end_index = end_indices[i]             
+            
+            res_id_str = custom_residue_ids[i]
+            
+            # Parse the string ID to separate number and insertion code.
+            # (\d+) captures digits (res_id), ([A-Z]?) captures optional char (ins_code)
+            match = re.match(r"(\d+)([A-Z]?)", res_id_str)
+            if match:
+                num = int(match.group(1))
+                code = match.group(2) if match.group(2) else ''
+            else:
+                num, code = 0, ''
+            
+            # Apply the new res_id and ins_code to all atoms in this residue range
+            new_res_id[start_index:end_index] = num
+            new_ins_code[start_index:end_index] = code
+            
+        # Overwrite the AtomArray annotations.
+        atom_array.set_annotation("res_id", new_res_id)
+        atom_array.set_annotation("ins_code", new_ins_code)
+        
+        return atom_array
+    
     def get_pae_confidence_scores(self, confidence_scores, atom_array):
         pae_confidence_scores = {}
         single_value_keys = [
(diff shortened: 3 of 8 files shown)


openfold3/core/runners/writer.py
```diff
@@ -208,12 +271,17 @@ def write_confidence_scores(
         elif out_fmt == "npz":
             np.savez_compressed(out_file_full, **full_confidence_scores)
 
+# openfold3/core/runners/writer.py
```
jnwei: Please remove this stray comment

jnwei (changes requested):
Hi @chodec, 

Thank you for working on this PR. The custom numbering feature is a tricky feature to get right, but as #58 indicates, it would be very useful for researchers.

I took a first pass over this PR today, and I have a few suggestions / questions:

- First, a check of my understanding: In the current implementation, a new method for creating a custom residue ID list (`get_custom_residue_ids`) in the `InferenceDataset` class. It looks like the intention is for the new residue IDs to be read by the output writer, but I don't see where the new residue ids are added to the batch?

- I would recommend that the custom `residue_id` list be created upon construction of the `Chain` class in `inference_query_format.py` rather than being generated in the `InferenceDataset`. This way, the logic around parsing the residue ids can be kept in one place, rather than adding extra logic to the InferenceDataset.
   - For an example of how to use pydantic validators to create the `residue_id` list given an input that is either a full list or an int, you might be able to borrow the logic used in `InferenceExperimentSettings` to generate random seeds from a list or a initial integer seed [here](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/entry_points/validator.py#L257)
   - The `InferenceDataset` can then be used to create batch features of the custom residue list if it is provided in a chain.
   
- Could you please add unit tests to test the creation of the custom residue numbering? I think it could be helpful to have two tests:
      - One test for generating the optional `residue_id` list in the `Chain` class, perhaps added [here](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_structure_from_query.py)
      - One test for writing the outputs, which could be added [here](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_writer.py)). 
- I would guess that some of the examples you used for manual validation of the numbering might be suitable test cases.

Also assigning @ljarosch to review, as he has more experience working with biotite and renumbering chains and may have additional suggestions regarding the organization.

Please let us know if you have any questions, and thank you again for your work on this issue!

chodec:
Hi @lucajovine  really sorry for the late reply. Between the holidays, a work crunch, and now my university exams, I completely lost track of this. I’m just heading out for a vacation now, but I’ll jump back into it as soon as I’m back. Thanks for your patience

openfold3/projects/of3_all_atom/config/inference_query_format.py
```diff
@@ -12,14 +12,15 @@
...
+from typing import Annotated, Any, NamedTuple, List, Optional, Union
+from pydantic import model_validator, Field
 
 from pydantic import (
     BaseModel,
     BeforeValidator,
     DirectoryPath,
     FilePath,
-    field_serializer,
+    field_serializer
```
jnwei, line 23: nit: The imports seem to have been separated into two groups from pydantic. Can these import statements be merged together?

A linting tool should be able to fix the import statements. For this project, we use [`ruff` ](https://docs.astral.sh/ruff/) with the settings in `pyproject.toml`.

openfold3/tests/__init__.py
jnwei, line 1: Please revert these changes to `tests/__init__.py` the final submission

openfold3/__init__.py
jnwei, line 1: Please revert these changes to `openfold3/__init__.py`

openfold3/projects/of3_all_atom/config/inference_query_format.py
```diff
@@ -76,6 +79,42 @@ def serialize_enum_name(self, v: MoleculeType, _info):
...
+                    "starting_residue_number must be convertible to an integer."
+                )
+
+            res_ids = [str(start_num + i) for i in range(sequence_length)]
+            data["residue_ids"] = res_ids
+            # Ensure consistency: clear starting_residue_number after use
+        else:
+            # Default to numbering starting from 1
+            res_ids = [str(1 + i) for i in range(sequence_length)]
+            data["residue_ids"] = res_ids
```
jnwei, line 115: I think it would be preferable to leave `residue_ids` as None if there is no custom residue numbering / start number provided by the user. 

I see that later, this custom `residue_ids` field is passed to the batch, which is then used to trigger the `renumber_residue_ids` later. If `residue_ids` is left blank, then default behavior for inference would remain the same (i.e. it would not perform the renumbering step).

examples/example_inference_inputs/query_multimer.json
jnwei, line 1: Would you mind making a new copy of `query_multimer.json` with the desired residue_ids. Perhaps you could save it as `examples/example_inference_inputs/query_multimer_custom_numbering.json`

openfold3/core/runners/writer.py
```diff
@@ -119,6 +120,70 @@ def write_structure_prediction(
             make_ost_compatible=make_ost_compatible,
         )
 
+    @staticmethod
+    def _renumber_atom_array(
```
jnwei, line 124: @ljarosch Do you have any comments about this method? I assume we have other parts of the data pipeline that would also require renumbering / reannotation of atom arrays. Would it make sense to add this function as one of the pipelines?

openfold3/tests/test_structure_from_query.py
```diff
@@ -87,8 +90,8 @@
         "non_canonical_peptide",
     ],
 )
-def test_structure_from_query(query: Query, ground_truth_file: Path):
-    """Tests that the generated structure and reference molecules matches gt."""
+def test_structure_with_ref_mols_from_query(query, ground_truth_file):
```
jnwei, line 93: The changes to `test_structure_with_ref_mols_from_query` seem unrelated to this PR? could we revert these changes?

openfold3/tests/test_structure_from_query.py
```diff
@@ -134,3 +137,44 @@ def test_smiles_with_explicit_hydrogen():
...
+    base_params = {
+        "molecule_type": MoleculeType.PROTEIN,
+        "chain_ids": ["A"],
+        "sequence": "AAA" 
+    }
+    
+    chain_default = Chain.model_validate(base_params)
+    assert chain_default.residue_ids == ['1', '2', '3']
+    
+    params_start = base_params.copy()
```
jnwei, line 156: I would recommend creating a separate test for each test case, rather than updating the base_params. This way, if a test fails, it is easy to see which specific use case has failed.

Perhaps an organization that is something like this (pseudocode, not tested)

```
class TestCustomResidueIDGeneration:
  def base_params(self):  # could potentially be a pytest.fixture instead
       return {
        "molecule_type": MoleculeType.PROTEIN,
        "chain_ids": ["A"],
        "sequence": "AAA" 
    }

  def test_base_definition(self):
    chain_default = Chain.model_validate(self.base_params())
    assert chain_default.residue_ids == ['1', '2', '3']

  def test_residue_id_starting_number(self):
         params_start = self.base_params().update({'starting_residue_number': "100"})
         assert chain_start.residue_ids == ['100', '101', '102']
```

openfold3/tests/test_writer.py
```diff
@@ -233,3 +249,87 @@ class DummyMock:
 
         assert writer.failed_count == 1
         assert writer.success_count == 0
+
+    def test_renumber_atom_array_with_insertion_codes(self):
```
jnwei, line 253: Would it be possible to explicitly write out the atom array? Similar to the dummy_array created here? https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/conftest.py#L10-L24

I find that it is much easier to check the input example when large inputs such as atom_arrays are written explicitly rather than constructed piece by piece.

jnwei (changes requested):
@chodec Thank you very much for your work on this PR and for refactoring the residue id construction into the Chain definition. I also really appreciate the additional tests and attention to detail, especially for getting parsing the insertion codes.

I can help run an end to end test for this PR some time next week. But for now, I wanted to provide some suggestions for organization of the tests and default behavior.

@ljarosch Please take a quick look, especially at the `_renumber_atom_array` logic in OF3OutputWriter.

jnwei:
Hi @chodec ! Circling back here, I wanted to check on your plans for this PR.

Worth flagging that a related PR was recently opened by @CesarPuentes. Perhaps Cesar could provide some assistance if you would like. I know this fix has been long awaited and it would be great to see the changes merged in.

Thanks!

lucajovine:
Thanks again to both @chodec and @CesarPuentes - hopefully this will finally make it through, one way or the other! Happy to test when merged, of course.

chodec:
Hey @jnwei sorry I'm totally overwhelmed (I believe you are all overwhelmed) but I'm finishing degree and work is just insane, but I'll look at it this week. Thanks for reminding!

jandom:
@chodec is there anything we can do to help with this PR? I think we're 80/90% there and your contribution would be really valuable to folks in the community

chodec:
Hey @jnwei, while re-testing this I ran into something I want to double-check with you before going further.

There's a shape mismatch between how the custom residue IDs are built and how they're read. get_custom_residue_ids() returns a nested list — one entry per chain — but _renumber_atom_array() treats it as a flat list and walks through it residue by residue across the whole atom array.

There's another wrinkle: the atom array is built by expanding each chain once per entry in chain_ids. So for a multimer with copied chains, the flat list has to be longer than the number of chains, or the writer runs off the end.

And separately, I noticed custom_residue_ids doesn't make it through the batch collator (openfold_batch_collator → dict_multimap) — it gets mangled before the writer ever sees it.

So the question is really: should this be a flat, chain-expanded list that matches the atom array's residue order, or should the writer handle it per chain? Happy to adjust the implementation either way once we settle that.

openfold3/tests/test_writer.py
```diff
@@ -234,3 +249,87 @@ class DummyMock:
...
+            atom_array=atom_array_short,
+            custom_residue_ids=custom_ids_short
+        )
+        
+        expected_res_id_short = np.repeat([10, 11, 0, 0, 0], atoms_per_residue)
+        np.testing.assert_array_equal(renumbered_array_short.res_id, expected_res_id_short)
+
+
+    def test_end_to_end_custom_ids_write(self, tmp_path):
+            """Ověřuje celý tok zápisu custom ID (včetně inzerčních kódů) do mmCIF souboru."""
```
jnwei, line 294: Just saw this when rereviewing the current changes, but please update this docstring to use English

jnwei (commented):
Thanks for raising this issue. 

I actually just thought of another approach for storing the custom residue ids -- perhaps we could add them as an annotation to the atom_arrays themselves? As an author id or a new field? 

We are already safely passing the atom arrays through the model for inference. Then at the writer stage, we could load the custom residue information from the atom array, and assign it to the final residue numbers.

What do you think? I believe this will sidestep the issues you are currently seeing with the residue id mismatch, and also help avoid how to pass a new data type through the batch.

PR #70, opened 2025-12-09
build(docker): layering and caching improvements

jandom:
**Summary**

Slight change to the main Dockerfile, ~~should probably do the Blackwell and test docker images too, for consistency.~~
The Blackwell Dockerfile is left untouched as its own thing for now.

We continue to use dockerhub for stable/production docker image distribution. 
This PR introduces also a GHCR (github container registry) which is easy to use with GH runners (self-hosted or otherwise). We use GHCR to store a special tag called `cache`, eg `cache-12.1.1-cudnn8-devel-ubuntu22.04`, which are **non-runnable** but are used to cache all the intermediate builder layers. Combined with improvements to dockerfile structure, this cache results in massively faster build times (because docker can check that it has the layer in GHCR already, doesn't rebuild but pulls instead). 

***Results*** 

No-op, rebuild from cache: 1.2s
Change a single source: 2.4+/-0.2s (N=3)

**Changes**

Improve docker image build speeds by installing a bare-bones version of the repo in the builder stage and then copying over the sources in the runtime stage. Now editing the sources will still give you a sub 10 seconds rebuild, when you have the cache warm. 

See detailed notes in the PR comments.

**Related Issues**

- disparate docker images 
- questions around how to package the code 

**Testing**

- Not sure how to test this, the container runs and I could import openfold, training would be ideal

Tests are broken with 

```
FAILED openfold3/tests/test_utils.py::TestUtils::test_chunk_layer_dict - AssertionError: tensor(False) is not true
FAILED openfold3/tests/test_utils.py::TestUtils::test_chunk_layer_tensor - AssertionError: tensor(False) is not true
```

but it's also broken on main, so not related to my changes

**Other Notes**

Manually triggered the gh workflow 

https://github.com/aqlaboratory/openfold-3/actions/runs/20098817240 ✅ 

***Dockerhub cache***

https://github.com/aqlaboratory/openfold-3/actions/runs/20101218016 ✅  but slow 1h 17min
https://github.com/aqlaboratory/openfold-3/actions/runs/20133611865 ✅  23mn looks like some cache is working!

***GHCR cache***
https://github.com/aqlaboratory/openfold-3/actions/runs/20135233502 ✅  GHCR cache, 50 min
https://github.com/aqlaboratory/openfold-3/actions/runs/20136728861 ✅ GHCR cache, 15 min – of which 14 s (!!!) were building the image

***GHCR cache + parallel re-usable workflows ***

https://github.com/aqlaboratory/openfold-3/actions/runs/20190734384/attempts/1 ✅ 32min cold, 12 min warm

** Final version ** 

We spin up N aws-runners, which can run multiple builds (different base images etc), here CUDA 12.1.1 only but can easily be extended.

Screenshot 2025-12-13 at 12 11 30

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -23,7 +23,6 @@ jobs:
       matrix:
         include:
           - cuda_base_image_tag: "12.1.1-cudnn8-devel-ubuntu22.04"
-          - cuda_base_image_tag: "12.2.2-cudnn8-devel-ubuntu22.04"
     concurrency:
       group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-${{ matrix.cuda_base_image_tag }}
       cancel-in-progress: true


jnwei (approved):
Overall this looks good. We have a [tests Dockerfile]( https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/Dockerfile) that extends the docker image with the packages needed for running tests. Perhaps you could also have a look at that Dockerfile and see if any similar optimizations could be made there.

.github/workflows/ci-test.yml
```diff
@@ -59,10 +59,10 @@ jobs:
         run: docker pull openfoldconsortium/openfold3:stable
       
       - name: Build test layers on Docker image 
-        run: docker build -t openfold3-test-runner -f openfold3/tests/Dockerfile . 
+        run: docker build --target test -t openfold-docker:test -f docker/Dockerfile .
```
jandom: Harmonized to match what's in DOCKER.md

docker/Dockerfile
```diff
@@ -85,7 +88,16 @@ ENV KMP_AFFINITY=none
...
 WORKDIR /opt/openfold3
+
+# Test stage - build on devel layer with test dependencies
+FROM devel AS test
+
+COPY environments/requirements-test.txt /opt/openfold3/requirements-test.txt
+
+WORKDIR /opt/openfold3
+RUN pip install -r requirements-test.txt
+RUN pip install --force-reinstall --no-deps .
```
jandom: I just added 'test' as another target within the same Dockerfile, one Dockerfile less!

docker/DOCKER.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+## Test images
+
+Build the test image
+```
+docker build \
+    -f docker/development/Dockerfile \
+    --target test \
+    -t openfold-docker:test .
```
jandom, line 25: The original Dockerfile now has two targets: devel and target

docker/Dockerfile
```diff
@@ -14,27 +14,32 @@ RUN apt-get update && apt-get install -y \
 
 # Install miniforge
 RUN wget -P /tmp \
-    "https://github.com/conda-forge/miniforge/releases/download/25.3.1-0/Miniforge3-Linux-x86_64.sh" \
+    "https://github.com/conda-forge/miniforge/releases/download/25.11.0-1/Miniforge3-Linux-x86_64.sh" \
```
jandom: Unpinning this may be a problem, @jnwei had to pin to avoid an infinite conda resolution time https://github.com/aqlaboratory/openfold3/pull/411

.github/workflows/ci-test.yml
```diff
@@ -8,80 +8,26 @@ on:
...
-        uses: omsf/stop-aws-gha-runner@v1.0.0
-        with:
-          instance_mapping: ${{ needs.start-aws-runner.outputs.mapping }}
-        env:
-          GH_PAT: ${{ secrets.GH_PAT }}
+      packages: write
+    strategy:
+      matrix:
+        include:
+          - cuda_base_image_tag: "12.1.1-cudnn8-devel-ubuntu22.04"
```
jnwei, line 25: A reminder to remove the extra cuda images that was included here for testing. I believe we plan to add a "nightly" workflow to test a fuller matrix (with multiple cuda images, maybe multiple python versions)

jandom: Ah, true – we've now confirmed that works

jandom: Removed, good catch @jnwei thanks!

PR #75, opened 2025-12-15
build: added environment.lock for 'stable' builds

jandom:
**Summary**

When using environment.yaml, conda may resolve to different package builds at different times. This PR creates a new docker image just to create a hermetic conda environment: taking in environment.yml and producing environment.lock (just for linux for now). 

Big change: new conda environment **openfold3** instead of **base**, which gets auto-activated.

**update** @jnwei and I discussed this and we want to do pinning for the 'stable' image but not for the tests or the development image. I need to re-think a bit how that's put together. This approach will give us the best of both worlds: people who consume the docker image for inference get exactly the correct deps, while developers have a more library-like dependency experience. 

**Changes**

- create a new Dockerfile.update-reqs 
- update the Dockerfile to use the environment.lock instead of environment.yml
- other minor changes

**Related Issues**

- N/A

**Testing**

- https://github.com/aqlaboratory/openfold-3/actions/runs/20234576248/job/58085915000 ✅ 
- https://github.com/aqlaboratory/openfold-3/actions/runs/20302873141 ✅  – with optional production.lock usage
- https://github.com/aqlaboratory/openfold-3/actions/runs/20331635064 ✅  with platform-linux-64.lock

**Other Notes**

None


docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Full performance multi-stage build with complete CUDA toolchain
-ARG CUDA_BASE_IMAGE_TAG=12.2.2-cudnn8-devel-ubuntu22.04
+ARG CUDA_BASE_IMAGE_TAG=12.1.1-cudnn8-devel-ubuntu22.04
```
jandom, line 2: This is unrelated and snuck in via #70, turning back to the old version we used

docker/Dockerfile
```diff
@@ -1,7 +1,10 @@
...
 FROM nvidia/cuda:${CUDA_BASE_IMAGE_TAG} AS builder
 
+# Environment mode: "lock" for reproducible builds, "yaml" for flexible dev builds
+ARG BUILD_MODE=lock
```
jnwei: Is there a way to provide a default option for this argument? There are some users who may prefer to use the Dockerfile directly to build their own image, rather than use the publishsed Dockerfile image, so it would be nice to have a default argument here.

docker/DOCKER.md
```diff
@@ -1,35 +1,75 @@
-## Production images
+## Updating the production.lock file
```
jnwei, line 5: Should we add documentation about where the current `production.lock` is generated?  Specifically, what kind of instance / system, and any other variables that are relevant to environment resolution.

For my understanding: Do we expect the `production.lock` to change if the system is a GPU / CPU? Or should it be the same because we specify the same docker base image with CUDA?

environments/production-linux-64.lock
```diff
@@ -0,0 +1,316 @@
...
+https://conda.anaconda.org/conda-forge/linux-64/readline-8.2-h8c095d6_2.conda#283b96675859b20a825f8fa30f311446
+https://conda.anaconda.org/conda-forge/linux-64/s2n-1.6.1-h23ded05_1.conda#d9f44253e28b48baee6e278d51ece70c
+https://conda.anaconda.org/conda-forge/linux-64/tk-8.6.13-noxft_ha0e22de_103.conda#86bc20552bf46075e3d92b67f089172d
+https://conda.anaconda.org/conda-forge/linux-64/xorg-libsm-1.2.6-he73a12e_0.conda#1c74ff8c35dcadf952a16f752ca5aa49
+https://conda.anaconda.org/conda-forge/linux-64/zlib-1.3.1-hb9d3cd8_2.conda#c9f075ab2f33b3bbee9e62d4ad0a6cd8
+https://conda.anaconda.org/conda-forge/linux-64/zlib-ng-2.3.2-h54a6638_0.conda#0faadd01896315ceea58bcc3479b1d21
+https://conda.anaconda.org/conda-forge/linux-64/zstd-1.5.7-hb78ec9c_6.conda#4a13eeac0b5c8e5b8ab496e6c4ddd829
+https://conda.anaconda.org/conda-forge/linux-64/aws-c-io-0.23.3-had5c4f5_4.conda#a53c9f532e5c4a2b85d4b4c439ea5a5d
+https://conda.anaconda.org/conda-forge/linux-64/brotli-bin-1.2.0-hb03c661_1.conda#af39b9a8711d4a8d437b52c1d78eb6a1
+https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-12.4.127-he02047a_2.conda#a748faa52331983fc3adcc3b116fe0e4
```
jnwei, line 79: just curious - why do we have two versions of cudart in the lockfile? Do they come from different dependencies?

environments/production-linux-64.lock
```diff
@@ -0,0 +1,316 @@
...
+https://conda.anaconda.org/conda-forge/noarch/font-ttf-dejavu-sans-mono-2.37-hab24e00_0.tar.bz2#0c96522c6bdaed4b1566d11387caaf45
+https://conda.anaconda.org/conda-forge/noarch/font-ttf-inconsolata-3.000-h77eed37_0.tar.bz2#34893075a5c9e55cdafac56607368fc6
+https://conda.anaconda.org/conda-forge/noarch/font-ttf-source-code-pro-2.038-h77eed37_0.tar.bz2#4d59c254e01d9cde7957100457e2d5fb
+https://conda.anaconda.org/conda-forge/noarch/font-ttf-ubuntu-0.83-h77eed37_3.conda#49023d73832ef61042f6a237cb2687e7
+https://conda.anaconda.org/conda-forge/linux-64/mpi-1.0-openmpi.tar.bz2#1dcc49e16749ff79ba2194fa5d4ca5e7
+https://conda.anaconda.org/conda-forge/noarch/python_abi-3.12-8_cp312.conda#c3efd25ac4d74b1584d2f7a57195ddf1
+https://conda.anaconda.org/pytorch/noarch/pytorch-mutex-1.0-cuda.tar.bz2#a948316e36fb5b11223b3fcfa93f8358
+https://conda.anaconda.org/conda-forge/noarch/tzdata-2025c-h8577fbf_0.conda#338201218b54cadff2e774ac27733990
+https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.11.12-hbd8a1cb_0.conda#f0991f0f84902f6b6009b4d2350a83aa
+https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-64-12.4.127-h85509e4_2.conda#329163110a96514802e9e64d971edf43
```
jnwei, line 15: the other cudart dependency

jnwei (approved):
LGTM! Thanks for adding the option to pin environments while leaving the CI testing environment flexible.

I just have a few minor comments and questions.

docker/Dockerfile
jandom: Hey, yeah i think this already contains the default, but it probably should be 'yaml' instead of 'lock' 
The non-default version of this is just 
```
ARG BUILD_MODE
```

docker/DOCKER.md
jandom: Yeah, that's a good point – currently the updating and generation is kind of synonymous (it doesn't say that). And it absolutely is platform specific, as you mentioned in your comments later. Yeah, it's specific to the platform (linux64, arm64, etc). The GPU/CPU point is conditional on what's in the environment.yaml – if that pulls a CPU-version of torch, that's what will be installed in the env.

environments/production-linux-64.lock
jandom: Ha, really great question – i don't know! :D But here is what claude had to say 

Screenshot 2025-12-18 at 09 36 37

docker/Dockerfile
jandom: This is fixed now

PR #79, opened 2025-12-19
build: remove setup.py and other build changes

jandom:
**Summary**

This PR provides changes to gradually harmonize some of the build process. Less moving pieces, and closer to best practices. 

**Changes**

- remove setup.py, this basically bypassed the messy deps in pyproject.toml but there are better ways to accomplish the same (`pip install --no-deps .`)
- refactor `requirements-test.txt` into `pyproject.toml`, using pip "dependency group" to bake this into `pyproject.toml` and install only the test deps without the main deps
- ignore `.pre-commit-config.yaml` because it's personal for now 
- add uv so that we can take advantage of of 'dep groups' in pyproject.toml
- separate yaml and lock files for linux-64 and osx-arm64 (experimental, we shouldn't support Mac for now)

**Related Issues**

The 'main' deps declared in `pyproject.toml` will break the production environment. I've confirmed that setup.py route only worked because it skipped all the deps declared in pyproject.toml :) 

**Testing**

Locally all dandy, launching a test job now. 

**Other Notes**


environments/production-linux-64.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - awscli
   - setuptools
   - pip
+  - conda-forge::uv
```
jnwei, line 12: nit: shouldn't uv be installed from conda-forge by default?

jnwei (approved):
Overall this looks great! Thank you for cleaning up the environment handling for the docker image.

IMO, I think if we're not fully supporting osx envrionments, we shouldn't include the experimental builds. Instead, I would suggest we recommend folks to use the [OpenFold3-MLX ](https://github.com/latent-spacecraft/openfold-3-mlx )project for support on Mac hardware.

environments/production-linux-64.yml
jandom: Thank you for the review! 

The OpenFold3-MLX is a slightly different story, since they want to re-write everything in this MLX framework. I'd like to support people running the CPU-native workloads (datapipeline) of OpenFold on Macs. But it's a bit experimental so I'd agree not to **signal** that we can support it (we can't yet, it'd be a lot of work). 

> nit: shouldn't uv be installed from conda-forge by default?

it should, i guess this is marginally more explicit (if another random channel provides this package, ignore it) but to be fair I don't really remember how the package vs multiple-channels thing work.

jandom: Actually, let's maybe keep the osx and say it's 'experimental' – maybe some members of the community will contribute off a nugget like this. People seem to be quite interested in building OF3 for their various platforms (DGX etc). What do you think?

jnwei: I haven't tested running OpenFold on CPUs, but my understanding is that the capacity of structures that could be predicted is quite limited because the triangle attention modules are optimized for GPUs. Did you already have a chance to test the osx builds on your machine?

If we haven't tested the osx build and/or it doesn't work on very many use cases, I would suggest omitting the osx builds for now. I think that organizing our environment files with `production-linux-64.*` provides a reference format for future contributors to add environments for other platforms that they've tested.

jandom: No, no really tested - okay, you've convinced me, I've removed the OSX stuff completely. I think this is ready to merge :D

PR #80, opened 2025-12-28
Fix check_sequence filter

### LINKED ISSUE #72: [BUG] Top templates incorrectly filtered out by training preprocessing script
**Describe the bug**
The training/validation preprocessing script for templates filters out many of the most relevant templates for some chains. The bug is in [check_sequence()](https://github.com/aqlaboratory/openfold-3/blob/afdf78e05a480f8cd2b6f6af652a714a2d94dd97/openfold3/core/data/primitives/sequence/template.py#L145C1-L176C1) which removes a template if its sequence length covers more than 95% of the query, but this should first check if the template sequence is a subsequence of the query. We have found skipped templates to be very common for certain types of chains like TCR Alpha and Beta where the top template alignments may align to >95% of the query positions but with many AA substitutions.

The function `check_sequence()` is called by the pipeline in `openfold3/scripts/data_preprocessing/preprocess_template_alignments_train_val_of3.py`. Inference has an updated pipeline in `openfold3/scripts/data_preprocessing/preprocess_template_alignments_new_of3.py` but the training option in that newer pipeline isn't implemented yet. Is there a different script with an accurate preprocessing pipeline for training/validation templates?

**To Reproduce**
Here is a simple example where a template with 100% coverage but 0% identity is filtered out by check_sequence().
```
from openfold3.core.data.io.sequence.template import parse_hmmsearch_a3m
from openfold3.core.data.primitives.sequence.template import check_sequence

# create short example template alignment file
with open("/tmp/test.a3m", "w") as f:
    f.write(
        ">query_A/1-20"
        "\n"
        "QQQQQQQQQQQQQQQQQQQQ"
        "\n"
        ">template1_A/1-20 mol:protein"
        "\n"
        "YYYYYYYYYYYYYYYYYYYY"
        "\n",
    )

# parse template alignments
with open(
    "/tmp/test.a3m",
    "r",
) as f:
    a3m_string = f.read()
template_hits = parse_hmmsearch_a3m(a3m_string=a3m_string)

# run relevant portions of create_template_cache_for_query()
query = template_hits[0]
hit = template_hits[1]
hit_pdb_id, hit_chain_id = hit.name.split("_")
# 1. Apply sequence filters: AF3 SI Section 2.4
if check_sequence(query_seq=query.hit_sequence.replace("-", ""), hit=hit):
    print(
        f"Template {hit_pdb_id} sequence does not pass sequence filters. Skipping this template."
    )
```
Results: `Template template1 sequence does not pass sequence filters. Skipping this template.`

**Expected behavior**
From AF3 SI Section 2.4: "We also remove templates that contain the exact query sequence with greater than 95% coverage as well as short templates less than 10 residues or covering less than 10% of the query." I expect this example template1_A to be kept because the aligned sequence is not an exact match to the query, is 20 residues long and covers 100% of the query. 

Please add label: OpenFold Consortium Member. Thank you!

gnikolenyi:
Fixes #72

### DIFF
--- a/openfold3/core/data/primitives/sequence/template.py
+++ b/openfold3/core/data/primitives/sequence/template.py
@@ -143,7 +143,7 @@ def parse_representatives(
 
 # Template cache construction
 def check_sequence(
-    query_seq: str,
+    query: TemplateHit,
     hit: TemplateHit,
     max_subseq: float = 0.95,
     min_align: float = 0.1,
@@ -152,8 +152,8 @@ def check_sequence(
     """Applies sequence filters to template hits following AF3 SI Section 2.4.
 
     Args:
-        query_seq (str):
-            The query sequence.
+        query (TemplateHit):
+            The query template_hit.
         hit (TemplateHit):
             Candidate template hit.
         max_subseq (float, optional):
@@ -167,12 +167,31 @@ def check_sequence(
         bool:
             Whether the hit passes the sequence filters.
     """
+    query_seq = query.hit_sequence.replace("-", "")
     hit_seq = hit.hit_sequence.replace("-", "")
-    return (
-        ((len(hit_seq) / len(query_seq)) > max_subseq)
-        | ((hit.aligned_cols / len(query_seq)) < min_align)
-        | (len(hit_seq) < min_len)
+    if len(hit_seq) < min_len:
+        return True, None, None
+    query_aln = np.frombuffer(
+        query.hit_sequence.replace(".", "-").encode("ascii"), dtype="S1"
     )
+    hit_aln = np.frombuffer(
+        hit.hit_sequence.replace(".", "-").encode("ascii"), dtype="S1"
+    )
+
+    query_not_gap = query_aln != b"-"
+    hit_not_gap = hit_aln != b"-"
+
+    columns_to_keep = query_not_gap & hit_not_gap
+    covered = columns_to_keep.sum()
+
+    coverage = covered / (len(query_seq) or 1)
+
+    if coverage < min_align:
+        return True, None, None
+
+    identical = (columns_to_keep & (query_not_gap == hit_not_gap)).sum()
+
+    return coverage >= max_subseq and identical == covered, query_aln, hit_aln
 
 
 def parse_release_date(cif_file: CIFFile) -> datetime:
--- a/openfold3/core/data/pipelines/preprocessing/template.py
+++ b/openfold3/core/data/pipelines/preprocessing/template.py
@@ -454,7 +454,8 @@ def create_template_cache_for_query(
             continue
 
         # 1. Apply sequence filters: AF3 SI Section 2.4
-        if check_sequence(query_seq=query.hit_sequence.replace("-", ""), hit=hit):
+        fails_sequence_filters, _, _ = check_sequence(query=query, hit=hit)
+        if fails_sequence_filters:
             template_process_logger.info(
                 f"Template {hit_pdb_id} sequence does not pass sequence"
                 " filters. Skipping this template."


openfold3/core/data/primitives/sequence/template.py
```diff
@@ -167,13 +167,32 @@ def check_sequence(
         bool:
```
jnwei, line 167: Could we update the return description to reflect the 3 values that are now being returned?

Also, it doesn't appear that we use the other return values of query_aln and hit_aln. Perhaps it would be easier to add these return values later if/when they are needed?

openfold3/core/data/primitives/sequence/template.py
```diff
@@ -143,7 +143,7 @@ def parse_representatives(
 
 # Template cache construction
 def check_sequence(
-    query_seq: str,
+    query: TemplateHit,
```
jnwei, line 146: Not strictly related to this PR, but could we update docstring for TemplateHit? Some of the fields seem out of date: https://github.com/aqlaboratory/openfold-3/blob/0b9df93313e8a4b42ea6d5e5ee900ff93ce26044/openfold3/core/data/io/sequence/template.py#L69-L85

openfold3/core/data/primitives/sequence/template.py
```diff
@@ -152,8 +152,8 @@ def check_sequence(
     """Applies sequence filters to template hits following AF3 SI Section 2.4.
```
jnwei, line 152: Could we add a quick description of the template filters from this section. AFAICT from the code, these filters are:
- Fails if `coverage < min_align` threshold.
- Fails if `coverage >= max_subseq AND covered == identical`.

Digging into the second statement, the anticipated outputs are:
- covered == identical -- this suggests that the template hit is identical to the query hit, because the non-gaps are located in the same places in the query / template hit. This hit would fail the filter
- covered != identical -- this suggests that some of the gap tokens in the matching sequence are not in the same locations, and thus the sequence is not a perfect match. This hit would pass the filter.

If the above understanding is correct, I am not sure if this would resolve the issue raised in the test example given in #72 . In that case, we have a hit which has 100% coverage, but has a different sequence value. In that test case, I believe the function would still fail the checks in this function.

openfold3/core/data/primitives/sequence/template.py
```diff
@@ -143,7 +143,7 @@ def parse_representatives(
 
 # Template cache construction
 def check_sequence(
```
jnwei, line 145: Could we consider making this function name more descriptive. Perhaps something like "check_seqence_similarity_within_range"?

jnwei (changes requested):
Thank you for submitting a fix for this issue @gnikolenyi ! I understand that this fix might be a smaller part of a set of larger fixes planned with the later release. Thank you for adding this fix separately first.

To ensure correctness of the `check_sequence` logic, it would be extremely helpful to have a few test cases. Simple examples like the one provided by @ECalfeeAdaptive in #72 are perfect for testing this function as it is easy to see the contents of the examples, and easy to check what the expected output should be.

Please let me know if I can provide any assistance with adding the the tests / documentation.

openfold3/core/data/primitives/sequence/template.py
ECalfeeAdaptive: Sequence positions that are aligned but not identical (AA substitutions) don't seem to be covered by this PR. I think the template 'duplicate' logic [here](https://github.com/aqlaboratory/openfold/blob/be2ec1841f16c966c65ae0e7599ebbadc725757d/openfold/data/templates.py#L273) from openfold could be re-used for openfold3 and would resolve the issue I raised in #72

jandom:
Agreed to close with @gnikolenyi, it's stale – this is coming later

PR #77, opened 2025-12-16
refactor: download_of3_databases.py and add tests

jandom:
**Summary**

Refactoring around download_of3_databases.py: adding types, tests, and some new functionality. 

Great care has been taken not to disrupt the interface that @vinay-swamy has put together. 

New functionality: show which datasets are available on S3

```
python download_of3_databases.py list 
Filename                        Size      Type     Known
--------------------------------------------------------
bfd.tar.gz                      272.2 GB  Protein  ✓    
cfdb.tar.gz                     269.8 GB  Protein  ✓    
mgnify.fasta.gz                 73.9 GB   Protein  ✓    
nucleotide_collection.fasta.gz  2.2 GB    DNA/RNA  ✓    
pdb_seqres.fasta.gz             52.2 MB   Protein  ✓    
rfam.fasta.gz                   58.6 MB   DNA/RNA  ✓    
rnacentral.fasta.gz             3.9 GB    DNA/RNA  ✓    
uniprot.fasta.gz                57.2 GB   Protein  ✓    
uniref30.tar.gz                 131.7 GB  Protein  ✓    
uniref90.fasta.gz               44.1 GB   Protein  ✓    
```

The download functionality as before, has been moved to a separate subcommand 

```
python download_of3_databases.py download --help
usage: download_of3_databases.py download [-h] [--output-dir OUTPUT_DIR] [--download-bfd] [--download-cfdb] [--download-rna-dbs]
                                          [--jackhmmer-dbs JACKHMMER_DBS [JACKHMMER_DBS ...]] [--hhblits-dbs HHBLITS_DBS [HHBLITS_DBS ...]]

options:
  -h, --help            show this help message and exit
  --output-dir OUTPUT_DIR
  --download-bfd
  --download-cfdb
  --download-rna-dbs
  --jackhmmer-dbs JACKHMMER_DBS [JACKHMMER_DBS ...]
                        Jackhmmer databases to download. Defaults to all: ['uniprot', 'uniref90', 'mgnify', 'pdb_seqres']. Use 'uniref90
                        pdb_seqres' for a minimal test set.
  --hhblits-dbs HHBLITS_DBS [HHBLITS_DBS ...]
                        HHblits databases to download. Defaults to: ['uniref30']. Use --download-bfd and --download-cfdb to add those databases.
```

new functionality here is being able to specify jackhammer and hhbltz dbs, & download only selected ones. 

**Changes**

[...]

PR #81, opened 2026-01-05
fix: preparse_alignments_of3 and add tests

jandom:
**Summary**

Looks like I introduced a bug during a refactor (what else is new!) in this PR #77: added some extra structure with the BaseModel but the underlying code expects a dictionary. It'd be cleaned to make everything consume the BaseModel but in the absence of tests, I just use the BaseModel for user-input validation and then dump to a dict. 

**Changes**

- Fixed the bug
- Added a smoke test to confirm files are being written out (and are not empty!)

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/tests/test_permutation_alignment.py
+++ b/openfold3/tests/test_permutation_alignment.py
@@ -24,7 +24,7 @@ from openfold3.core.data.primitives.permutation.mol_labels import (
 )
 from openfold3.tests.custom_assert_utils import assert_atomarray_equal
 
-TEST_DIR = Path(__file__).parent / "test_data" / " permutation_alignment"
+TEST_DIR = Path(__file__).parent / "test_data" / "permutation_alignment"
 
 
 def test_mol_symmetry_id_assignment():


docs/source/precomputed_msa_how_to.md
```diff
@@ -158,6 +158,16 @@ alignments/
 
 For main MSAs, we provide support for preparsing MSA files from `a3m` and `sto` into `npz` numpy array format. This optional step reduces overall MSA parsing and processing times in the inference pipeline for large batch jobs with redundant sets of sequences as well as the disk space required to store these MSAs (see {ref}`this <5-preparsing-raw-msas-into-npz-format>` section for details). The preparsed, compressed MSA files can be generated from raw MSA data with the directory and file structure specified above using [this preparsing script](https://github.com/aqlaboratory/openfold-3/blob/main/scripts/data_preprocessing/preparse_alignments_of3.py). 
 
+Here is an example use
+```bash
+python ./scripts/data_preprocessing/preparse_alignments_of3.py \
+    --alignments_directory msa-raw/some-input-dir/ \
+    --alignment_array_directory msa-cache/some-output-dir/ \
+    --num_workers 1 \
+    --max_seq_counts '{"uniref90_hits":1024}'
```
vinay-swamy, line 176: For the example here, I think it would be good to include the correct mapping for all 4 DBs here, just so it's consistent with the output of our MSA pipeline. This would be :
```
'{"uniprot_hits": 50000, "uniref90_hits": 10000, "cfdb_hits": 100000000, "mgnify_hits": 5000, "rfam_hits": 10000, "rnacentral_hits": 10000, "nucleotide_collection_hits": 10000}'
```

jandom: That's a good shout actually – I also added a "minimal" example with the above.

openfold3/tests/scripts/data_preprocessing/test_preparse_alignments_of3.py
```diff
@@ -0,0 +1,99 @@
...
+                "--num_workers",
+                "1",
+            ],
+        )
+
+        assert result.exit_code == 0, f"CLI failed with: {result.output}"
+
+        # Check that npz files were created for both chains
+        npz_files = list(tmp_path.glob("*.npz"))
+        assert len(npz_files) == 6, (
```
jnwei: Could we add a little more context for why 6 is the expected number of npz files? Would it make sense to check for the list of filenames instead?

openfold3/tests/scripts/data_preprocessing/test_preparse_alignments_of3.py
```diff
@@ -0,0 +1,99 @@
...
+
+class TestPreparseAlignmentsOf3:
+    """Smoke tests for preparse_alignments_of3.py script."""
+
+    @pytest.fixture
+    def cli_runner(self):
+        return CliRunner()
+
+    def test_preparse_databases(self, cli_runner, tmp_path):
+        """Test preparsing alignments with a single database (uniref90_hits)."""
```
jnwei: Could we add the the expected input file type (e.g. .sto)?

jnwei (commented):
LGTM, some small nits regarding documentation on tets.

openfold3/tests/scripts/data_preprocessing/test_preparse_alignments_of3.py
jandom: Yeah, I re-organized the tests a bit – by default the script processes the whole directory and that's not ideal (we were mixing loose alignment files with 2 PDB, which were the actual inputs)

openfold3/tests/scripts/data_preprocessing/test_preparse_alignments_of3.py
jandom: It actually could be both a3m/sto – I'm hesitant to write documentation for the script in the test :P What's the intent here?

jnwei: When reviewing, it isn't obvious to me what are the inputs to this script since as you mention later, we have a combination of inputs and outputs in the `testdata/alignments` directory.

It would be helpful to leave a signpost, either in the test or in the test_dir, of the expected inputs to help future readers / maintainers. 

Another option could be to rearrange the test directory to have the inputs / outputs separated.

jandom: Ah, makes sense now – so this is how this directory is organized 

image

The two directories make sense. The loose file directly under alignments/ I have no idea what these are - they're not outputs. In fact, they don't seem to be called by anything. I'm going to do an exploratory 'rm'... 

Added the docstring explaining what the script does to these inputs.

PR #83, opened 2026-01-06
fix: typo in Dockerfile

jandom:
**Summary**

Looks like I broke main :( 

https://github.com/aqlaboratory/openfold-3/actions/runs/20757496085/job/59603591517

**Changes**

Remove a typo

**Related Issues**

**Testing**

- the 'safe-to-test' label will use the 'main' version of the docker image, workflow dispatch allows us to run this against the ref of choice. Manually triggered the run here

https://github.com/aqlaboratory/openfold-3/actions/runs/20777084831 ✅ 

**Other Notes**

### DIFF
--- a/.github/workflows/ci-test-reusable.yml
+++ b/.github/workflows/ci-test-reusable.yml
@@ -52,6 +52,8 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@v6
+        with: 
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
 
       - name: Log in to GHCR
         uses: docker/login-action@v3


.github/workflows/ci-test-reusable.yml
```diff
@@ -52,6 +52,8 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@v6
+        with: 
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
jandom, line 56: without this, the workflow will by default checkout the main branch – not ideal

environments/production-linux-64.yml
```diff
@@ -23,7 +23,7 @@ dependencies:
   - modelcif
   - ml-collections
   - mkl
-  - rdkit
+  - rdkit=2025.09.3
```
jandom, line 26: needed temporarily because the latest rdkit `2025.09.4` is only on conda but not pip at the moment – this confuses the package manager.

jnwei: Maybe add a small to do to remove this pin once the latest rdkit pin is on pip? RDKit is a core dependency for many projects so it would be nice to remove the pin as soon as the new pip package is available.

jnwei (approved):
LGTM. Confirmed that it is fine to use the pull request head directly as we only intend to run the CI pipeline for PRs that have been reviewed by a maintainer.

PR #87, opened 2026-01-13
refactor: tests for the cleanup.py

jandom:
**Summary**

Starting to lay down some tests for the data pipeline, starting with the cleanup.py module and going to try and keep the PRs small. 

**Changes**

Added a tests for the remove_waters method

**Related Issues**

**Testing**

**Other Notes**


openfold3/tests/conftest.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+import pytest
+from biotite.structure import AtomArray
+
+
+@pytest.fixture
+def dummy_atom_array():
```
jandom, line 7: moved from test_writer.py for easier re-use across tests

jnwei: Nice! I wonder if it would be better for the chain_id attribute to be defined by the test in test_writer.py, similar to how you defined res_name separately in the test for test_cleanup_waters.py

jandom: Ah, sorry I pressed too soon - but that's a good comment, I'll be adding more tests here, so will refactor that – just to clarify, pul out the chain_id definition from the fixture into the test in the test_writer?

jnwei: Yes, since the chain_id is only tested in the test_writer tests and might not be applicable (or lead to incorrect behavior) if used as a fixture for other tests.

PR #89, opened 2026-01-14
refactor: tests for cleanup.py

jandom:
**Summary**

Carpet bombing the cleanup.py module with tests

**Changes**

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/tests/core/data/primitives/structure/test_cleanup.py
+++ b/openfold3/tests/core/data/primitives/structure/test_cleanup.py
@@ -22,6 +22,7 @@ from openfold3.core.data.primitives.structure.cleanup import (
     remove_crystallization_aids,
     return_on_empty_atom_array,
 )
+from openfold3.tests.custom_assert_utils import assert_atomarray_equal
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


openfold3/tests/conftest.py
```diff
@@ -20,3 +20,33 @@ def dummy_atom_array():
...
+    atom_array.element[:8] = ["N", "C", "C", "O", "C", "C", "SE", "C"]
+    atom_array.hetero[:8] = True
+
+    # ALA residue (res_id=2)
+    atom_array.chain_id[8:] = "A"
+    atom_array.res_id[8:] = 2
+    atom_array.res_name[8:] = "ALA"
+    atom_array.atom_name[8:] = ["N", "CA", "C", "O", "CB"]
+    atom_array.element[8:] = ["N", "C", "C", "O", "C"]
+    atom_array.hetero[8:] = False
```
jandom, line 50: @jnwei I'm a bit worried about this – Claude will just pollute with fixtures like this, when maybe this could be constructed more programmatically?

openfold3/tests/core/data/primitives/structure/test_cleanup.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+class TestReturnOnEmptyAtomArrayDecorator:
+    """Tests for the return_on_empty_atom_array decorator."""
+
+    def test_empty_atom_array_returns_immediately(self):
+        """When given an empty AtomArray, the decorator returns it without calling the wrapped function."""
+        function_was_called = False
+
+        @return_on_empty_atom_array
```
jnwei: nit: If possible, I would suggest pulling out the dummy function here and in the next test: `test_non_empty_atom_array_calls_function` since it appears to be the same function

The main behavior we are testing is that the function is skipped when given an empty atom array. Calling the exact same function will help make this more explicit.

jandom: This should be addressed now – fully agree, I was hand-cooking way too many asserts. 

The nit-to-nit would be for this assert to return True/False, so that we can assert both ways for when changes are and are not made :P

PR #90, opened 2026-01-15
feat: adding pytest-cov and pytest-xdist

jandom:
**Summary**

This will help my test writing efforts tremendously: adding an ability to generate coverage reports, and, running tests in parallel

**Changes**

Added pytest-cov and pytest-xdist as test-only deps

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/.github/workflows/ci-test-reusable.yml
+++ b/.github/workflows/ci-test-reusable.yml
@@ -85,7 +85,7 @@ jobs:
           docker run \
             -v ${{ github.workspace }}:/opt/openfold3 \
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:test-${{ inputs.cuda_base_image_tag }}-${{ github.sha }} \
-            pytest openfold3/tests -vvv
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test-reusable.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           docker run \
             -v ${{ github.workspace }}:/opt/openfold3 \
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:test-${{ inputs.cuda_base_image_tag }}-${{ github.sha }} \
-            pytest openfold3/tests -vvv
+            pytest openfold3/tests -vvv -n auto --cov=openfold3 --cov-report=xml
```
jandom, line 88: This could be baked into `pyproject.toml` but I like it more opt-in and explicit this way (at least for now)

jnwei (commented):
LGTM! could you also report the current code coverage so we know where we're starting from?

PR #95, opened 2026-01-21
fix: CCD/SMILES inconsistency

### LINKED ISSUE #94: [BUG] Inference fails with some CCD codes but works with SMILES
**Describe the bug**

This bug is not really a surprise, reported by Pat Walters. It relates to how we inject custom CCD .bcif into biotite and how that doesn't happen at inference time, and cannot be configured (today). 

**To Reproduce**

Here is the config that breaks 

```json
{
  "queries": {
    "cyp3a4_9bv5": {
      "chains": [
        {
          "sequence": "MALIPDLAMETWLLLAVSLVLLYLYGTHSHGLFKKLGIPGPTPLPFLGNILSYHKGFCMFDMECHKKYGKVWGFYDGQQPVLAITDPDMIKTVLVKECYSVFTNRRPFGPVGFMKSAISIAEDEEWKRLRSLLSPTFTSGKLKEMVPIIAQYGDVLVRNLRREAETGKPVTLKDVFGAYSMDVITSTSFGVNIDSLNNPQDPFVENTKKLLRFDFLDPFFLSITVFPFLIPILEVLNICVFPREVTNFLRKSVKRMKESRLEDTQKHRVDFLQLMIDSQNSKETESHKALSDLELVAQSIIFIFAGYETTSSVLSFIMYELATHPDVQQKLQEEIDAVLPNKAPPTYDTVLQMEYLDMVVNETLRLFPIAMRLERVCKKDVEINGMFIPKGVVVMIPSYALHRDPKYWTEPEKFLPERFSKKNKDNIDPYIYTPFGSGPRNCIGMRFALMNMKLALIRVLQNFSFKPCKETQIPLKLSLGGLLQPEKPVVLKVESRDGTVSGA",
          "molecule_type": "protein",
          "chain_ids": [
            "A"
          ]
        },
        {
          "smiles": "C=CC1=C(C)/C2=C/c3c(C)c(CCC(=O)O)c4[n]3[Fe@SP2]35<-[N]2=C1/C=c1/c(C)c(C=C)/c([n]13)=C/C1=[N]->5/C(=C\\4)C(CCC(=O)O)=C1C",
          "molecule_type": "ligand",
          "chain_ids": [
            "B"
          ]
        },
        {
          "ccd_codes": "A1ASV",
          "molecule_type": "ligand",
          "chain_ids": [
            "C"
          ]
        }
      ]
    }
  }
}

```

Here is the config that works

```json
{
  "queries": {
    "cyp3a4_9bv5": {
      "chains": [
        {
          "sequence": "MALIPDLAMETWLLLAVSLVLLYLYGTHSHGLFKKLGIPGPTPLPFLGNILSYHKGFCMFDMECHKKYGKVWGFYDGQQPVLAITDPDMIKTVLVKECYSVFTNRRPFGPVGFMKSAISIAEDEEWKRLRSLLSPTFTSGKLKEMVPIIAQYGDVLVRNLRREAETGKPVTLKDVFGAYSMDVITSTSFGVNIDSLNNPQDPFVENTKKLLRFDFLDPFFLSITVFPFLIPILEVLNICVFPREVTNFLRKSVKRMKESRLEDTQKHRVDFLQLMIDSQNSKETESHKALSDLELVAQSIIFIFAGYETTSSVLSFIMYELATHPDVQQKLQEEIDAVLPNKAPPTYDTVLQMEYLDMVVNETLRLFPIAMRLERVCKKDVEINGMFIPKGVVVMIPSYALHRDPKYWTEPEKFLPERFSKKNKDNIDPYIYTPFGSGPRNCIGMRFALMNMKLALIRVLQNFSFKPCKETQIPLKLSLGGLLQPEKPVVLKVESRDGTVSGA",
          "molecule_type": "protein",
          "chain_ids": [
            "A"
          ]
        },
        {
          "smiles": "C=CC1=C(C)/C2=C/c3c(C)c(CCC(=O)O)c4[n]3[Fe@SP2]35<-[N]2=C1/C=c1/c(C)c(C=C)/c([n]13)=C/C1=[N]->5/C(=C\\4)C(CCC(=O)O)=C1C",
          "molecule_type": "ligand",
          "chain_ids": [
            "B"
          ]
        },
        {
          "smiles": "CC(C)(C)C(=O)Nc1cc(ccc1n2ccnc2)C(F)(F)F",
          "molecule_type": "ligand",
          "chain_ids": [
            "C"
          ]
        }
      ]
    }
  }
}
```

**Expected behavior**

Ideally, we'd want things to be consistent between CCD and SMILES (Boltz has tens of issues around various inconsistencies). 

- DGX

**Additional context**

I saw this snippet in inference.py and thought we can provide this at inference time 

```
# Parse CCD
        if dataset_config.ccd_file_path is not None:
            logger.debug("Parsing CCD file.")
            self.ccd = pdbx.CIFFile.read(dataset_config.ccd_file_path)
        else:
            self.ccd = BiotiteCCDWrapper()
```

But while we can, this is not actually what ends up being used when we call `structure_with_ref_mol_from_ccd_code`. 

Instead the hack that worked was 

```
        # Parse CCD
        if dataset_config.ccd_file_path is not None:
            logger.debug("Parsing CCD file.")
            self.ccd = pdbx.CIFFile.read(dataset_config.ccd_file_path)
        else:
            self.ccd = BiotiteCCDWrapper()

        import biotite.structure as struc
        struc.info.set_ccd_path("/home/jandom/workspace/openfold-3/components.bcif")
```

Now this is not ideal because it involves downloading and preparing the bcif file, like so 

```
wget https://files.wwpdb.org/pub/pdb/data/monomers/components.cif.gz
gunzip components.cif.gz
python scripts/data_preprocessing/preprocess_ccd_biotite.py components.cif components.bcif
```

**Stack trace**

<details>
<summary>Trace</summary>

```
run_openfold predict     --query_json=research/2024-q1/pat-walters/data/inputs/ccd_codes/cyp3a4_9bv5.json     --output_dir=./output/pat/cyp3a4_9bv5/
/home/jandom/micromamba/envs/openfold3-env/lib/python3.13/site-packages/torch/cuda/__init__.py:283: UserWarning: 
    Found GPU0 NVIDIA GB10 which is of cuda capability 12.1.
    Minimum and Maximum cuda capability supported by this version of PyTorch is
    (8.0) - (12.0)
    
  warnings.warn(
💡 Tip: For seamless cloud uploads and versioning, try installing [litmodels](https://pypi.org/project/litmodels/) to enable LitModelCheckpoint, which syncs automatically with the Lightning model registry.
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
WARNING:openfold3.core.data.tools.colabfold_msa_server:Using output directory: /tmp/of3_colabfold_msas for ColabFold MSAs.
Submitting 1 sequences to the Colabfold MSA server for main MSAs...
COMPLETE: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 150/150 [elapsed: 00:02 remaining: 00:00]
/home/jandom/workspace/openfold-3/openfold3/core/data/tools/colabfold_msa_server.py:331: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
  tar_gz.extractall(path)
No complexes found for paired MSA generation. Skipping...
Preprocessing templates...
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
/home/jandom/micromamba/envs/openfold3-env/lib/python3.13/multiprocessing/popen_fork.py:67: DeprecationWarning: This process (pid=267696) is multi-threaded, use of fork() may lead to deadlocks in the child.
  self.pid = os.fork()
WARNING:openfold3.core.data.framework.single_datasets.inference:----------------------------------------
Failed to process cyp3a4_9bv5 with preferredException type: KeyError
Traceback: Traceback (most recent call last):
  File "/home/jandom/micromamba/envs/openfold3-env/lib/python3.13/site-packages/biotite/structure/info/atoms.py", line 82, in residue
    component = get_component(
        get_ccd(),
        res_name=res_name,
        allow_missing_coord=allow_missing_coord,
    )
  File "/home/jandom/micromamba/envs/openfold3-env/lib/python3.13/site-packages/biotite/structure/io/pdbx/convert.py", line 1330, in get_component
    raise KeyError(
    ...<2 lines>...
    )
KeyError: "No rows with residue name 'A1ASV' found in 'chem_comp_atom' category"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/framework/single_datasets/inference.py", line 338, in __getitem__
    features = self.create_all_features(query)
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/framework/single_datasets/inference.py", line 295, in create_all_features
    structure_objs = self.get_structure_with_ref_mols(
        query=query,
    )
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/framework/single_datasets/inference.py", line 191, in get_structure_with_ref_mols
    atom_array, processed_reference_molecules = structure_with_ref_mols_from_query(
                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        query=query,
        ^^^^^^^^^^^^
    )
    ^
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/primitives/structure/query.py", line 603, in structure_with_ref_mols_from_query
    structure_with_ref_mol_from_ccd_code(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        ccd_code=chain.ccd_codes[0],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        chain_id=chain_id,
        ^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/primitives/structure/query.py", line 465, in structure_with_ref_mol_from_ccd_code
    atom_array = atom_array_from_ccd_code(
        ccd_code,
    ...<2 lines>...
        molecule_type=MoleculeType.LIGAND,
    )
  File "/home/jandom/workspace/openfold-3/openfold3/core/data/primitives/structure/query.py", line 121, in atom_array_from_ccd_code
    res_array = get_residue_cached(ccd_code)
  File "/home/jandom/micromamba/envs/openfold3-env/lib/python3.13/site-packages/biotite/structure/info/atoms.py", line 88, in residue
    raise KeyError(f"No atom information found for residue '{res_name}' in CCD")
KeyError: "No atom information found for residue 'A1ASV' in CCD"

```
</details>

jandom:
**Summary**

Fixes #94 

**Changes**

- Added tests that show the problem

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/scripts/update_ccd.py
+++ b/openfold3/scripts/update_ccd.py
@@ -28,9 +28,7 @@ import biotite.setup_ccd
 import boto3
 
 from openfold3.core.utils.s3 import compute_local_crc64nvme_base64
-
-S3_BUCKET = "openfold3-data"
-S3_KEY = "components.bcif"
+from openfold3.setup_openfold import S3_BUCKET, S3_KEY
 
 logging.basicConfig(level=logging.INFO, format="%(message)s")
 logger = logging.getLogger(__name__)


openfold3/setup_openfold.py
```diff
@@ -172,6 +174,18 @@ def download_parameters(param_dir) -> None:
     logger.info("Download completed successfully.")
 
 
+def setup_biotite_ccd() -> None:
```
jandom: I'm also half-tempted to call this when the inference pipeline is called, tbh – what do you think?

ljarosch: It takes pretty long to run this (like several minutes for me), so I'd probably not make it the default to run every time. Maybe it could be a flag one can enable. If we want to be fancy we could probably query the last download date of the Biotite CCD on inference and throw a warning with a suggestion to run `python -m biotite.setup_ccd` once it is outdated beyond a certain threshold.

openfold3/setup_openfold.py
```diff
@@ -172,6 +174,18 @@ def download_parameters(param_dir) -> None:
...
+    logger.info("Starting Biotite CCD setup...")
+    if not biotite.setup_ccd.OUTPUT_CCD.exists():
```
ljarosch: Would we actually expect this condition to ever trigger? Biotite ships a CCD during its own setup, so I'd always expect one to exist there. Imo the problem in Pat's case isn't that there is no Biotite CCD downloaded, it is that the ligands he is working with are so recent that they won't be found in outdated CCDs, which is why one needs to trigger a manual update.

jandom: No, actually – empirically I found that I need to run this and then tests start to pass. The installed biotite somehow comes with a different CCD file?

jandom: Ah, no I think you're partially right actually – there is some default CCD file that installs with biotite but it's out-of-date and running the setup command updates it

ljarosch: Yeah this is caused by our version constraint of biotite 1.2.0, which likely ships an outdated CCD. I can confirm that a non-constrained install of biotite doesn't run into this problem.

openfold3/setup_openfold.py
jandom: I think i'm leaninng towards the side of being smart (not my usual choice). Because i don't want the users to know/care what biotite is. This complexity should be abstracted away. 

Sorry about my confusion: i thought that biotite ships without a CCD file but it in fact does – it's just stale and needs to be updated.

openfold3/setup_openfold.py
ljarosch: It's still a general problem though, of course we can ameliorate it for some time by installing a fresh CCD on setup, but the PDB will keep adding new ligands weekly so you can quickly run into the same problem again. It might be reasonable enough to just include a helpful error message when a ligand could not be found in the CCD that suggests running the setup_ccd command.

ljarosch: (and we'd need to get rid of the guard because `biotite.setup_ccd.OUTPUT_CCD.exists()` should always return true after biotite was installed)

jnwei: Overall this is great!

Is there a way to fetch the date / version of the CCD file? Then `setup_ccd` could prompt the user if they want to download a newer version when it is available.

docs/source/Installation.md
```diff
@@ -72,6 +72,7 @@ This script will:
 - Setup a directory for OpenFold3 model parameters [default: `~/.openfold3`]
     - Writes the path to `$OPENFOLD_CACHE/ckpt_path` 
 - Download the model parameters, if the parameter file does not already exist 
+- Download and setup the CCD with Biotite
```
jnwei: Suggestion: Provide full name of CCD, e.g.

`Download and setup the [Chemical Component Dictionary (CCD) with Biotite](https://www.biotite-python.org/latest/apidoc/biotite.structure.info.get_ccd.html)`

jandom: Updated

openfold3/setup_openfold.py
jandom: Negative, we can get an ETag/last modified of the remote CCD file – but it's not the same file as biotite saves, because it does some modification prior. So we can't do a simple ETag1 == ETag2. 

Instead I opted for a "stale ETag list" and if the file on disk is stale, it'll re-download. 

I excluded all solutions with a sidecar file, this will only make it more messy. 

This is overall not great because we hard-code a md5 sum in code but until we move beyond biotite==1.2.0, I don't see a better way.

docs/source/Installation.md
```diff
@@ -72,6 +72,7 @@ This script will:
 - Setup a directory for OpenFold3 model parameters [default: `~/.openfold3`]
     - Writes the path to `$OPENFOLD_CACHE/ckpt_path` 
 - Download the model parameters, if the parameter file does not already exist 
+- Download and setup the [Chemical Component Dictionary (CCD) with Biotite](https://www.biotite-python.org/latest/apidoc/biotite.structure.info.get_ccd.html)
```
ljarosch: Not sure if the pointer of this link is super helpful, this is just a function to work with Biotite's CCD representation in code. Why not just add `python -m biotite.setup_ccd` if the intent is to show users how to download and set it up?

jandom: We could surface this as work-around but that's really our abstraction leaking. One day we might remove biotite, and the users should be none the wiser about it.

jnwei: My intention was to provide users who might be unfamiliar with the CCD some context about what it is since it is now included in our top level Installation documentation and how biotite is used in relation to the CCD.

We can certainly replace this link with a more general pointer to the CCD (perhaps this one: https://www.wwpdb.org/data/ccd) if that would be more helpful. This might make more sense if we now plan to provide the CCD file ourselves

openfold3/tests/core/data/primitives/structure/test_query.py
```diff
@@ -0,0 +1,40 @@
...
+            "c1ccc(cc1)C(c2ccccc2)([C@@H]3CCN(C3)CCc4ccc5c(c4)CCO5)C(=O)N",
+            "A1CIW",
+        ),  # cyp3a4_9plk
+    ],
+)
+def test_consistent_structure_from_smiles_and_ccd_code(smiles, ccd_code):
+    struct_from_smiles = structure_with_ref_mol_from_smiles(smiles, chain_id="X")
+    struct_from_ccd = structure_with_ref_mol_from_ccd_code(ccd_code, chain_id="X")
+
+    assert len(struct_from_smiles.atom_array) == len(struct_from_ccd.atom_array)
```
jnwei, line 46: Some quick suggestions to make this test more comprehensive: 

- Would it make sense to use the `custom_assert_utils.assert_atom_array_equal` utility here? Or would we expect some differences in atom ordering / bond ordering so that this assert would fail.

- If `assert_atom_array_equal` is not a good fit here, perhaps we could add a test to check the number of carbons or some other aggregate property.

jandom: Ah, it'd be really nice indeed if we could use that – but it currently doesn't work, just tried empirically (atm atom names different between smiles and CCD, eg C01 vs C1) – maybe we can just match on element count?

jnwei: Yeah figures. I think the atom counter is good enough here for now, thanks for adding it.

jnwei (approved):
LGTM

PR #102, opened 2026-01-28
feat: docker image for SandboxAQ/aqaffinity

jandom:
**Summary**

Here is the experimental build I did on an AWS g5.xlarge with the deep learning AMI as base image. 

Requires to be logged into hugging face cli locally, to clone the repo at image build. Confirm login with 

```
hf auth whoami
user:  my-username
orgs:  my-orgs
```

Build with 

```
docker build \
    -f docker/Dockerfile \
    --secret id=hf_token,src=$HOME/.cache/huggingface/token \
    --target affinity \
    -t openfold-docker:affinity .
```

Run with 

```
docker run --rm \
    --gpus all \
    --ipc=host \
	-v "${HOME}/.openfold3/:/root/.openfold3" \
	-v "/home/ubuntu/workspace/openfold-3:/opt/openfold3" \
    -v "${PWD}/output:/output" \
    -it openfold-docker:affinity \
    aqaffinity predict \
        --query_json ./examples/example_inference_inputs/query_single_protein_single_ligand.json \
        --runner_yaml ./examples/example_runner_yamls/affinity.yaml \
        --num_diffusion_samples=1 \
        --use_templates=false \
        --output_dir /output \
        --binding_affinity_ckpt_path /opt/aqaffinity/model_weights/model_weights_only.pt
```

And then the output will be in 

```
cat output/mcl1/seed_42/mcl1_seed_42_sample_1_binding_head.txt 
-2.8919382095336914
```

**Changes**

- new target 'affinity' in the Dockerfile, very easy to incorporate the SandboxAQ package

**Related Issues**

**Testing**

**Other Notes**


examples/example_inference_inputs/query_single_protein_ligand_affinity.json
jnwei, line 1: Can we reuse any of the existing examples?

The current file name `query_single_protein_ligand_affinity.json` suggests to me that there is a difference between the inputs required for affinity prediction vs the inputs for openfold3 inference. From a glance, it doesn't look like there are any additional input fields required for the affinity modle.

If we want a simpler protein/ ligand filename, perhaps we could name the example file `query_single_protein_single_ligand.json`?

jnwei (approved):
LGTM, but maybe this would be a PR better suited for the Sandbox AQ affinity repo itself?

Two small comments:
- Could you mention the system that you tested this docker image on?
- I have a suggestion regarding the example file naming, see in comment in line.

examples/example_inference_inputs/query_single_protein_ligand_affinity.json
jandom: Correct, and i was surprised by that too – it only works on single-chain, single-ligand complexes. None of the examples, as far as I could see, had that. 

We should probably document that behavior somewhere. 

+1 to simpler filename, yeah

jandom: Updated the filename now

Issue #101, opened 2026-01-28
[BUG] Issue in template alignments for short template sequences


jnwei:
**Describe the bug**
When template alignment matches return templates with short sequences, an error occurs with the alignment from `kalign2`.

**To Reproduce**
<details>
<summary>Example query input</summary>

````
{  
"queries": {
      "query1": {
         "chains": [
            {
               "molecule_type": "protein",
               "chain_ids": [
                  "A"
               ],
               "sequence": "AQVINTFDGVADYLQTYHKLPDNYITKSEAQALGWVASKGNLADVAPGKSIGGDIFSNREGKLPGKSGRTWREADINYTSGFRNSDRILYSSDWLIYKTTDHYQTFTKIR"
            },
            {
               "molecule_type": "protein",
               "chain_ids": [
                  "B"
               ],
               "sequence": "KKAVINGEQIRSISDLHQTLKKELALPEYYGENLDALWDALTGWVEYPLVLEWRQFEQSKQLTENGAESVLQVFREAKAEGADITIILS"
            },
}
````
</details>

**Expected behavior**
We should be able to handle sequence alignments against short sequences

**Stack trace**

```
IndexError                                Traceback (most recent call last)
Cell In[26], [line 9](vscode-notebook-cell:?execution_count=26&line=9)
      6 template_sequence = "GC"
      8 parser = A3mParser(max_sequences=None)
----> [9](vscode-notebook-cell:?execution_count=26&line=9) parser(
     10     f">query_X/1-{len(query_seq_str)}\n{query_seq_str}\n>{template_entry_id}/{1}-{len(template_sequence)}\n{template_sequence}\n",
     11     query_seq_str,
     12     realign=True,
     13 )

File ~/github/openfold3/openfold3/core/data/io/sequence/template.py:739, in A3mParser.__call__(self, alignment_source, query_seq_str, realign)
    734 realigned_str = run_kalign(all_sequences)
    735 realigned_alignments, _ = parse_fasta(realigned_str)
    737 return self._process_alignment_hits(
    738     query_seq_str=query_seq_str,
--> [739](https://file+.vscode-resource.vscode-cdn.net/Users/jennifer/github/openfold3/~/github/openfold3/openfold3/core/data/io/sequence/template.py:739)     query_aln_str=realigned_alignments[0],
    740     template_alignments=realigned_alignments[1:],
    741     headers=headers,  # Use original headers for metadata
    742 )

IndexError: list index out of range
```

[...]

PR #114, opened 2026-02-05
docs: debugging instructions for docker

jandom:
**Summary**

While working on https://github.com/aqlaboratory/openfold-3/issues/101 I found it very useful to run with an interactive debugger. Decided to write some docs explaining how to set this up

**Changes**

See comments on the PR

**Related Issues**

**Testing**

**Other Notes**


.dockerignore
```diff
@@ -13,4 +13,5 @@
 !environments/**
 !examples/**
 !openfold3/**
-!scripts/**
\ No newline at end of file
+!scripts/**
+!examples/**
```
jandom, line 17: Unrelated but helpful: we previously were excluding the JSON examples from the docker image, they're actually quite nice to have baked in

pyproject.toml
```diff
@@ -70,6 +70,7 @@ test = [
   "pytest-xdist",
   "pytest-cov",
   "pytest-benchmark",
+  "debugpy",
```
jandom, line 73: This allows for the interactive debugger, added as a test dep

jnwei (changes requested):
Overall looks good! Thank you for adding this.

Under which tab do you think this should go in the documentation? Currently we follow the diataxis break down and you can see the organization of docs here: https://github.com/aqlaboratory/openfold-3/blob/main/docs/source/index.md?plain=1#L53-L86

Maybe this could go under deep-dives? It could potentially go under How-To, but it might be nice to separate this as a special "how to" for development.

jandom:
This is definitely a "how to" in my mind – let me update the contents.

PR #117, opened 2026-02-07
fix: chain template alignments auth labelling (inference)

jandom:
**Summary**

Hopefully helps to solve https://github.com/aqlaboratory/openfold-3/issues/101

**Changes**

So far wrote a test that reproduced the failure, and then added a "fix"

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/tests/core/data/tools/test_colabfold_msa_server.py
+++ b/openfold3/tests/core/data/tools/test_colabfold_msa_server.py
@@ -68,7 +68,10 @@ def _mock_fetch_label_to_author(pdb_ids):
 
 
 def _make_m8_dataframe(template_ids: list[str], m_index: int = 101) -> pd.DataFrame:
-    """Build a minimal m8-format DataFrame for testing."""
+    """Build a minimal m8-format DataFrame for testing.
+
+    See docs/source/template_how_to.md § 1.1.3 for the m8 column spec.
+    """
     n = len(template_ids)
     return pd.DataFrame(
         {
--- a/openfold3/tests/core/data/tools/test_rscb.py
+++ b/openfold3/tests/core/data/tools/test_rscb.py
@@ -1,3 +1,21 @@
+"""Tests for the RCSB GraphQL API helpers in ``openfold3.core.data.tools.rscb``.
+
+Tests marked ``@pytest.mark.vcr`` use *pytest-recording* (vcrpy) to replay
+HTTP responses from YAML cassettes stored alongside this file in
+``cassettes/``.
+
+Generating cassettes for the first time::
+
+    pytest openfold3/tests/core/data/tools/test_rscb.py --vcr-record=all
+
+Re-recording after RCSB schema changes or new test methods::
+
+    pytest openfold3/tests/core/data/tools/test_rscb.py --vcr-record=new_episodes
+
+In CI the cassettes are replayed without network access (the default
+``--vcr-record=none`` mode).
+"""
+
 import pytest
 
 from openfold3.core.data.tools.rscb import (


openfold3/tests/test_data/template_alignments/colabfold_template.m8
```diff
@@ -0,0 +1,62 @@
+101	1b27_B	1.0	110	0	0	1	110	1	110	1.022e-45	160	110M
```
jandom: This needs to be generated at test-time, I don't want to be adding files into the repo

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -0,0 +1,83 @@
...
+        template_structure_file = tmp_path / f"{template.entry_id}.cif"
+
+        cif_file = _load_ciffile(template_structure_file)
+
+        chain_id_seq_map = get_asym_id_to_canonical_seq_dict(cif_file)
+
+        # template.chain_id is an author chain ID; map it to label asym_id
+        label_to_author = get_label_to_author_chain_id_dict(cif_file)
+        author_to_label = {v: k for k, v in label_to_author.items()}
+        label_chain_id = author_to_label[template.chain_id]
```
jandom: This is the re-mapping from "auth" chains IDs to "label" chain IDs... very wishful in terms of inputs not being pathological

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -0,0 +1,83 @@
...
+            query_seq_str=query_seq_str,
+            max_sequences=200
+
+        )
+        
+        # find the offending "1rnb_A"
+        template = templates[16]
+        assert template.chain_id == "A" and template.entry_id == "1rnb"
+
+        fetch(
```
jnwei: Can we mock this call instead of explicitly calling the RCSB database using `fetch`?

As this is a unit test, it would be good to remove dependencies on web servers so that we don't have latency issues / failures due to the availability of the service.

openfold3/tests/test_data/template_alignments/colabfold_template.m8
jnwei: I think we could be helpful to include a explicit colabfold_template.m8 sample for a few reasons.

1) It allows the user to directly see the expected inputs to our template parser function
2) Generating this file at test time would require a call to the colabfold msa server, since we don't use the colabfold binary. It is good practice to avoid dependencies on web services during tests to avoid latency and test breakages due to service unavailability.

If the main concern is adding an additional file, I would suggest adding the example content as a text block in the module using `textwrap` ([example](https://github.com/aqlaboratory/openfold-3/blob/25daf12a195e6291beea08f4450e7477b17af703/openfold3/tests/test_dataset_config.py#L43-L66)).  This particular example could be made much shorter, e.g. we could include just 5 alignment lines, including the offending 1rnb_A line.

gnikolenyi: I agree with Jennifer, we already have some smaller files in the repo for this purpose under openfold3/tests/test_data. I think it should be fine to have this there.

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
gnikolenyi: So if there are multiple label IDs mapping to the same author ID, this will just always map an author ID to the l ast label ID. @ljarosch can confirm, but I think this only happens with homomeric chains, so it should be fine. We should just document this behavior.

A way to make this more robust would be to explicitly sort the label_to_author dict when iterating over it, so maybe add that here so we are not relying on the dict ordering for this mapping.

jandom:
@jnwei @gnikolenyi many thanks for the reviews – I'm not sure why this wasn't a draft, this is clearly not ready for prime time. 

Agreed that we need this to use some fixture files (in an integration test context), but I would also like to have an end-to-end test that does everything. Ideally, we would just have in-memory generated fixtures but we don't have the tooling setup atm.

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
jandom: Switched to just a cif file as fixture

gnikolenyi:
@jnwei @jandom Added the template structure download and chain ID remapping logic to the colabfold pipeline and removed it from the template pipeline. See logs below for an example with the remapping printed explicitly:

```
Submitting 5 sequences to the Colabfold MSA server for main MSAs...
COMPLETE: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 750/750 [elapsed: 00:02 remaining: 00:00]
Downloading template CIFs: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 38/38 [00:04<00:00,  8.53it/s]
Remapped 4uyk: author A -> label A
Remapped 1914: author A -> label A
Remapped 1e8o: author A -> label A
Remapped 3jaj: author S1 -> label JC
Remapped 5aox: author D -> label D
Remapped 4ue5: author E -> label E
Remapped 1e8s: author A -> label A
Remapped 4uyk: author B -> label B
Remapped 7nfx: author t -> label WA
Remapped 5aox: author B -> label B
Remapped 1914: author A -> label A
Remapped 4uyj: author B -> label B
Remapped 5aox: author E -> label E
Remapped 1e8o: author D -> label D
Remapped 7obr: author t -> label WA
Remapped 1e8o: author B -> label B
Remapped 4ue5: author B -> label B
Remapped 2w9j: author B -> label B
Remapped 2w9j: author A -> label A
Remapped 4gnx: author A -> label A
Remapped 3kdf: author A -> label A
Remapped 7uy6: author F -> label E
Remapped 7lmb: author F -> label G
Remapped 6d6v: author F -> label C
Remapped 4gnx: author B -> label B
Remapped 2pqa: author C -> label C
Remapped 2pqa: author A -> label A
Remapped 1l1o: author E -> label E
Remapped 2z6k: author A -> label A
Remapped 1quq: author A -> label A
Remapped 2pi2: author C -> label C
Remapped 6i52: author B -> label B
Remapped 2z6k: author B -> label B
Remapped 1quq: author C -> label C
Remapped 3kdf: author D -> label B
Remapped 4joi: author B -> label B
Remapped 4joi: author A -> label A
Remapped 7u5c: author F -> label F
Remapped 6w6w: author C -> label D
Remapped 8d0k: author B -> label B
Remapped 8c5y: author B -> label B
Remapped 4gnx: author C -> label C
Remapped 1jmc: author A -> label B
Remapped 1fgu: author B -> label B
Remapped 6i52: author C -> label C
Remapped 1l1o: author C -> label C
Remapped 1l1o: author F -> label F
Remapped 1ynx: author A -> label A
Remapped 8aaj: author A -> label A
Remapped 8c5z: author A -> label A
Remapped 8oej: author A -> label A
Remapped 8oej: author D -> label D
Remapped 6d6v: author D -> label A
Remapped 8c5y: author J -> label J
Remapped 3u50: author C -> label A
Remapped 1o7i: author B -> label B
Remapped 7wcg: author A -> label A
Remapped 3dm3: author B -> label B
Remapped 2k5v: author A -> label A
Submitting 2 paired MSA queries to the Colabfold MSA server...
COMPLETE: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 300/300 [elapsed: 00:01 remaining: 00:00]
COMPLETE: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 450/450 [elapsed: 00:02 remaining: 00:00]
Computing paired MSAs: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:03<00:00,  1.92s/it]
```
For example, checking 3jaj author S1 -> label JC: https://www.rcsb.org/sequence/3JAJ#JC

Note the following:
 - only up to the first 25 templates are parsed and remapped from auth to label asym ids to reduce the amount of cif files that need to be downloaded - we can expose this as an argument to the runner in a later PR
 - the cif files are deduplicated within an inference run and only the non-redundant set of cifs are downloaded, but across multiple runs, the cif files are re-downloaded
 - the cif files are not reused for the actual template processing, which would require a more significant refactor - both this and the re-download across inference runs are areas which we can optimize later

openfold3/core/data/primitives/caches/filtering.py
```diff
@@ -855,83 +855,6 @@ def set_nan_fallback_conformer_flag(
     return None
 
 
-# TODO: Do this in preprocessing instead to avoid it going out-of-sync with the data?
-def get_model_ranking_fit(pdb_id):
```
jandom, line 859: Moved to a new rscb.py module

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -642,6 +646,68 @@ def save_colabfold_mappings(
...
+    per_rep: dict[str, pd.DataFrame] = {}
+    unique_pdb_ids: set[str] = set()
+    for rep_id in rep_ids:
+        m_i = rep_id_to_m[rep_id]
+        if m_i not in m_with_templates:
+            continue
+        chain_alns = template_alignments[template_alignments[0] == m_i]
+        top_n = chain_alns.copy()
+        per_rep[rep_id] = top_n
+        unique_pdb_ids.update(top_n[1].str.split("_").str[0])
```
jandom, line 681: I don't love this code – but it's just refactored into a function...

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -725,47 +808,33 @@ def query_format_main(self):
             # Create empty DataFrame with expected column structure (at least column 0)
             # to match the structure when file is read with header=None.
             logger.warning(
```
jandom, line 810: Maybe we should error here

openfold3/core/data/tools/rscb.py
```diff
@@ -0,0 +1,163 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+import requests
```
jandom, line 17: This file is all brand new

ljarosch:
Hi @jandom and @gnikolenyi, I'll give this a better review later this week. One question already - if we download the CIF anyway (to get the actual template structure), is it actually a good idea to rely on another endpoint (the graphQL query interface)?

One more general issue I could see with our current code:
- It's not entirely clear what template distribution the CF-server provides (it's local to the server and consistent with their PDB70 alignment file) - the PDB occasionally retrospectively updates structure details on already released structures which can change the expected chain correspondence (we've been bit by this multiple times)
- We officially support custom CF-servers https://openfold-3.readthedocs.io/en/latest/inference.html#use-a-privately-hosted-colabfold-msa-server, where we have no control over the template structures the CF-server returns (they might be custom, older PDB versions, ...)

So a bit of a broader point, but I wonder if instead of relying on all these RCSB endpoints we should just have a self-contained logic that just takes the CIF that the CF-server returns and figures out the mapping, template coordinates, ... from that?

openfold3/tests/core/data/tools/test_rscb.py
jnwei, line 1: I think it's fine to have some tests that hit the real RCSB API, as you said, it is important to have some integration tests. 

However, given that our group alone runs our CI tests ~10 times and possibly more in a day, and that other developers may also run our CI battery of tests, I recommend we add a label to these tests so that we can filter these tests and reduce the frequency so that it is not part of the battery of unit tests. Maybe something like the `pytest.mark` for the inference integration tests, here: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_inference_full.py#L35

In the long run, we can set up a cronjob to run this and other integration tests to run once a day by default, and upon manual trigger by the developer.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
jnwei, line 1: Not reviewed. Will review this file in #145

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -0,0 +1,75 @@
...
+from openfold3.core.data.io.structure.cif import _load_ciffile
+from openfold3.core.data.primitives.structure.metadata import (
+    get_asym_id_to_canonical_seq_dict,
+    get_author_to_label_chain_ids,
+)
+
+_TEST_DATA_DIR = Path(openfold3.__file__).parent / "tests" / "test_data"
+
+
+class TestTemplatePreprocessor:
```
jnwei: Would it be a big lift to also have a test case to ensure a chain that has a consistent template alignment / author chain id is left unadulterated?

Unfortunately, 1RNB seems to be a small monomer protein, so it looks like we cannot simply use a different chain of this structure. Perhaps we can revisit this and add another test case later if we add more test cases with template alignments later.

openfold3/tests/core/data/primitives/structure/test_metadata.py
jnwei, line 1: Nice simple sanity check unit test, I like it.

openfold3/core/data/tools/colabfold_msa_server.py
jnwei: We could raise an error here instead. 

My concern is if a user is running a large batch of predictions, they may prefer to be notified later about the issue with missing templates, rather than have the workflow interrupted for a few broken examples. We could think of a better way to record this issue and bring attention to the missing template alginments?

openfold3/core/data/tools/rscb.py
jnwei: nice, I like this separation.

openfold3/core/data/tools/colabfold_msa_server.py
jandom: Hard to tell – either way it's out of scope in a way, because it's not related to the bug-fix per se. Should we handle this in another PR? This PR is already 20 files, we're ballooning

openfold3/tests/core/data/primitives/structure/test_metadata.py
jandom: We could also VCR  for this case – otherwise we're hitting colabfold. But here it's harder – colabfold doesn't produce a single JSON response (like RSCB API) but instead launches a job and the user needs to then download and unpack various files.

openfold3/tests/core/data/tools/test_rscb.py
jandom: Agreed – added pytest-recordings to save the response as "cassettes" (YAML files) containing the response

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
jandom: I'm a little confused here – what test case are we looking for? Something where the author id = label id?

jandom: Looking at this test case, it's kind of ugly – it's basically a copy of the inference pipeline (where the fix was originally applied). 

The two cases we're concerned about are 
- author-id not the same as label-id
- author-id same as label-id

But these are now covered by these tests 
- test_remap_author_to_label
- TestFetchLabelToAuthorChainIds

I'm going to remove this test, i think it adds nothing

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -44,6 +45,49 @@
...
+}
+
+
+def _mock_fetch_label_to_author(pdb_ids):
+    """Return mock label->author mappings for known test PDB IDs."""
+    return {pid: _MOCK_LABEL_TO_AUTHOR.get(pid, {}) for pid in pdb_ids}
+
+
+def _make_m8_dataframe(template_ids: list[str], m_index: int = 101) -> pd.DataFrame:
+    """Build a minimal m8-format DataFrame for testing."""
```
jnwei: nit: add a small reference to the m8 format description? We have one in our docs https://openfold-3.readthedocs.io/en/latest/template_how_to.html#m8

jnwei (approved):
Overall this is great. The pytest-recroding is perfect for our use case.

Could you also add a small README / comment to test_rscb.py that describes how to handle the recording, in case it needs to be regenerated?

From [their docs](https://pypi.org/project/pytest-recording/) looks like it should just be

```
pytest --record-mode=once test_rcsb.py
```

Or maybe `rewrite` should be used instead?

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
jandom: On it

PR #121, opened 2026-02-09
Update kalign usage to use python API

### LINKED ISSUE #115: [BUG] Support kalign3
**Describe the bug**
[kalign3 has a python API](https://github.com/TimoLassmann/kalign/releases/tag/v3.4.6-rc1) and there is [development towards making kalign available on pypi](https://github.com/TimoLassmann/kalign/pull/53). We should change our kalign usage to support the python API

Doing so will also allow us to retrieve all of our dependencies from pypi alone.

jnwei:
**Summary**

Updates usage of kalign to use python API and updates kalign installation to use pypi. 

After this change, openfold3 will be available to install entirely from pypi. Thank you @etowahadams and @TimoLassmann for the work to enable these changes in https://github.com/TimoLassmann/kalign/pull/53

BREAKING: After this set of changes is merged in, OpenFold3 will no longer be compatible with kalign2. Kalign must be installed via pypi

**Changes**

- Updates usage of kalign to use python API. This means we no longer create fastas / parse fastas for use with kalign.
- Changes setup_openfold to make conda environment setup optional
- Changes installation documentation to remove mamba installation of kalign2
- Changes production.yml recipe. TOOD: Generate new lockfile

**Related Issues**
#115 

**Testing**
- [x] Test that `openfold3/tests/tests_template_parsers.py` path
- [ ] Test setup and inference with `setup_openfold` 

**Other Notes**
Attention @gnikolenyi  for updates to kalign usage and @sdvillal for updates in dependencies.

### DIFF
--- a/openfold3/setup_openfold.py
+++ b/openfold3/setup_openfold.py
@@ -40,17 +40,18 @@ logger = logging.getLogger(__name__)
 def check_conda_use() -> bool:
     """Check the current environment and return environment info."""
     logger.info("Checking environment...")
-    
+
     use_conda = bool(os.environ.get("CONDA_PREFIX"))
     if use_conda:
         logger.info(
-        f"Running in conda environment: {os.path.basename(os.environ['CONDA_PREFIX'])}"
-    )
+            "Running in conda environment: "
+            f"{os.path.basename(os.environ['CONDA_PREFIX'])}"
+        )
     else:
         logger.info("Not running in a conda environment.")
         logger.info("The script will run without conda-specific features.")
-    
-    return use_conda 
+
+    return use_conda
 
 
 def setup_openfold_cache(use_conda) -> tuple[Path, Path]:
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
   # For cutlass >=4, import library is cutlass_cppgen, but deepspeed is
   # currently using cutlass.
   "nvidia-cutlass <4",
+  "kalign",
```
sdvillal, line 36: in an ideal world this could be named "kalign-python", to avoid a clash with a possible system package "kalign" that would not necessarily include the python bindings

etowahadams: The version installed with python is invoked using `kalign-py` so I think this issue would be avoided?

sdvillal: Sorry for the confusion, here we are only talking about the package name, not the executable name.

Because there are package managers that can package both non-python and python dependencies (conda, homebrew, linux package manages...), it is not best practice to name the pypi package the same as a potential system package. I tried to explain the issue here: https://github.com/TimoLassmann/kalign/issues/56

TimoLassmann: I changed the package name to kalign-python - hope this helps! /T

sdvillal: Awesome, thanks!

openfold3/setup_openfold.py
```diff
@@ -73,29 +75,38 @@ def setup_openfold_cache() -> tuple[Path, Path]:
...
+        try:
+            subprocess.run(
+                [
+                    "conda",
+                    "env",
+                    "config",
+                    "vars",
+                    "set",
+                    f"OPENFOLD_CACHE={openfold_cache}",
+                ],
```
jandom, line 88: This is not strictly related to this PR, so more of a nitpick. Micromamba/mamba/conda all expose $CONDA_PREFIX – but the 'conda' binary might not be available. 

I would rather we don't do this at all, and ask users to define their OPENFOLD_CACHE in environment.yml

PR #125, opened 2026-02-12
ci: require the safe-to-test label only on the public repo

jandom:
**Summary**

Only require the label on the public repo

**Changes**

**Related Issues**

**Testing**

**Other Notes**


jandom:
Thanks @jnwei

PR #126, opened 2026-02-12
build: harmonize docker image pushes

jandom:
**Summary**

Pushes to GHCR (all) and DockerHub (only public)

**Changes**

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -41,7 +41,7 @@ jobs:
           GH_PAT: ${{ secrets.GH_PAT }}
 
   push-to-registry:
-    name: Push Docker image to Docker Hub
+    name: Push Docker image to registries
     runs-on: ${{ fromJSON(needs.start-aws-runner.outputs.instances) }}
     needs:
       - start-aws-runner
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -64,7 +75,9 @@ jobs:
         id: meta
         uses: docker/metadata-action@a60f0f62b5c12316066c8db464bfb2faafd7907b
         with:
-          images: openfoldconsortium/openfold3
+          images: |
+            name=ghcr.io/${{ github.repository }}/openfold3-docker
+            name=openfoldconsortium/openfold3,enable=${{ github.repository == 'aqlaboratory/openfold-3' }}
```
jandom, line 80: This was curious to me

.github/workflows/docker.yml
```diff
@@ -75,12 +88,14 @@ jobs:
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83
         with:
           context: .
-          file: ./Dockerfile
```
jandom, line 78: This file no longer exists

.github/workflows/docker.yml
jnwei: which part is curious? If this is about the missing `-` in the name of the docker image, it's because dockerhub doesn't allow docker names with hyphens.

jnwei (approved):
LGTM, thank you for updating the docker publishing workflow

.github/workflows/docker.yml
jandom: Oh, just the multiple name tags and how they can be mixed with the enable tag. Never seen something like this before.

PR #127, opened 2026-02-17
fix: fix_arginine_naming skipping structures

jandom:
**Summary**

@ljarosch tells me that this is an issue around numpy compatibility (I don't think it is but I'm not sure), this was not an issue in training but it does happen for me when preparing new datasets. 

**Changes**

Skip the `fix_arginine_naming` if there are missing atoms

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/core/data/primitives/structure/cleanup.py
+++ b/openfold3/core/data/primitives/structure/cleanup.py
@@ -138,7 +138,7 @@ def fix_arginine_naming(atom_array: AtomArray) -> AtomArray:
 
         # Skip if any of the required atoms are missing
         if nh1_coord.size == 0 or nh2_coord.size == 0 or cd_coord.size == 0:
-            logger.warning(
+            logger.debug(
                 f"Skipping arginine naming fix for residue "
                 f"missing atoms - NH1={np.sum(nh1_mask)}, "
                 f"NH2={np.sum(nh2_mask)}, CD={np.sum(cd_mask)}"


openfold3/tests/core/data/primitives/structure/test_cleanup.py
```diff
@@ -233,6 +233,34 @@ def test_cleans_up_temporary_annotation(self, bad_arginine_atom_array):
...
+        """When an arginine is missing required atoms, it is skipped without error.
+
+        This represents real-world cases where an arginine residue is only partially
+        resolved in the crystal structure. See PDB entries like 4YVD and 8DMQ.
+        """
+        mask = ~np.isin(bad_arginine_atom_array.atom_name, atoms_to_remove)
+        incomplete_arg = bad_arginine_atom_array[mask]
+
+        # Should not raise an error
+        result = fix_arginine_naming(incomplete_arg)
```
jnwei, line 258: nit: If desired, you can use the built-in fixture [pytest.logging.caplog](https://docs.pytest.org/en/stable/reference/reference.html#pytest.logging.caplog) to check that the logging warning was triggered. Something like this pattern

```
def test_skips_arginine_with_missing_atoms(self, bad_arginine_atom_array, atoms_to_remove, caplog):
    with caplog.at_level(logging.WARNING):
        fix_arginine_naming(incomplete_arg)

    assert "Skipping arginine naming fix" in caplog.text

```

jnwei (approved):
This looks great, thank you for adding the fix and the test @jandom 

For the record, could you describe the previous erroneous behavior that occurred without the current fix that skips incomplete arginine structures?

openfold3/tests/core/data/primitives/structure/test_cleanup.py
jandom: Ah, this is very good indeed, because I don't want any warnings in the logs broadly – whenever they appear (and they're not explicitly asked for) it's typically a bad sign.

openfold3/core/data/primitives/structure/cleanup.py
```diff
@@ -136,6 +136,15 @@ def fix_arginine_naming(atom_array: AtomArray) -> AtomArray:
         nh2_coord = arginine_view.coord[nh2_mask]
         cd_coord = arginine_view.coord[cd_mask]
 
+        # Skip if any of the required atoms are missing
+        if nh1_coord.size == 0 or nh2_coord.size == 0 or cd_coord.size == 0:
+            logger.warning(
```
ljarosch: NIT: I think this could be downgraded to debug log-level. It's not unexpected that some arginine atoms are missing and the way the code is handling it now is safe and shouldn't "worry" the user with a lot of warning messages.

jandom: Resolved now

PR #128, opened 2026-02-17
test: triangular attention snapshots

jandom:
Introduces **numerical snapshot regression testing** for the triangular attention and triangular multiplicative update layers using [`pytest-regressions`](https://github.com/ESSS/pytest-regressions). This catches silent numerical changes to layer outputs across code changes, hardware, or dependency upgrades.

### What changed

- **New test fixtures** (`conftest.py`):
  - `device` — parametrizes each test over CPU and CUDA (auto-skips CUDA when unavailable)
  - `seeded_rng` — isolates torch/numpy/python RNG state via `torch.random.fork_rng()` so tests don't leak randomness to each other
  - `_cuda_deterministic` (autouse) — enables deterministic CUDA ops for reproducible results
  - `original_datadir` — redirects `pytest-regressions` snapshot storage to `test_data/snapshots/<test_module>/`
  - `SnapshotEnv` — records environment metadata (torch version, CUDA version, GPU name) alongside snapshots; emits a warning when running against snapshots generated on a different environment

- **Migrated tests from `unittest` to pytest**:
  - `test_triangular_attention.py` — parametrized over `starting={True, False}`, runs on both CPU/CUDA, snapshots output via `ndarrays_regression`
  - `test_triangular_multiplicative_update.py` — same pattern; extracted `_make_module` helper for the fused/non-fused variant selection

- **Snapshot data**: `.npz` reference files for each (test, device, parameter) combination, generated on NVIDIA GB10 / CUDA 13.0 / PyTorch 2.10

- **New dependency**: `pytest-regressions` added to `[test]` extras in `pyproject.toml`

### How to regenerate snapshots

```bash
pytest openfold3/tests/test_triangular_attention.py openfold3/tests/test_triangular_multiplicative_update.py --force-regen
```

This overwrites the `.npz` files and writes a fresh `_snapshot_env.json` to each snapshot directory.

### Test plan
- [x] Tests pass on CPU
- [x] Tests pass on CUDA (NVIDIA GB10)
- [x] Snapshots verified cross-platform (x86 CPU VM + aarch64 DGX)
- [ ] Verify `--force-regen` writes correct `_snapshot_env.json`


openfold3/tests/conftest.py
```diff
@@ -76,3 +78,9 @@ def ensure_biotite_ccd(request):
 def biotite_ccd_wrapper():
     """Cache CCD wrapper fixture for tests that need it."""
     return BiotiteCCDWrapper()
+
+
+@pytest.fixture(scope="module")
+def original_datadir(request: pytest.FixtureRequest) -> Path:
+    """Redirect pytest-regressions snapshot storage to test_data/snapshots/."""
+    return Path(__file__).parent / "test_data" / "snapshots" / Path(request.path).stem
```
jandom: This is what allows us to place the snapshots along other test_data, sibling to the 'cassettes'

jnwei: Nice! Does it make sense to update the snapshot paths for the templates test you added in `https://github.com/aqlaboratory/openfold-3/tree/main/openfold3/tests/test_data/cassettes/test_rscb`

Can be a different PR if it is cumbersome to update here.

openfold3/tests/test_triangular_attention.py
```diff
@@ -12,38 +12,64 @@
...
-        starting = True
+# starting=True -> "starting node" variant: rows attend to rows,
+# biased by z[i, k]. False would transpose internally for the
+# "ending node" variant (columns attend to columns).
+@pytest.mark.parametrize("starting", [True, False])
+def test_shape(starting, ndarrays_regression):
+    # NOTE: seeding may need further work — torch.manual_seed controls both
+    # the random input and the module's weight init. If init changes upstream,
+    # regenerate snapshots with: pytest --force-regen
+    torch.manual_seed(123)
```
jnwei: I think you can use a context manager to handle the rng state, something like this (I have not tested this myself)

```
#conftest.py

import random
import numpy as np
import torch
import pytest
from torch.random import fork_rng

@pytest.fixture()
def seeded_rng():
    """Isolate all RNG state (torch, numpy, python) for the duration of the test."""
    py_state = random.getstate()
    np_state = np.random.get_state()
    with fork_rng():          # saves/restores torch (+ CUDA) state
        torch.manual_seed(123)
        # set other random states if neeeded
        random.seed(123)
        np.random.seed(123)
        yield
    # torch state restored by freeze_rng_state on exit
    random.setstate(py_state)         # restore python state manually
    np.random.set_state(np_state)     # restore numpy state manually
```

[Docs for fork_rng](https://docs.pytorch.org/docs/stable/random.html#torch.random.fork_rng)

openfold3/tests/conftest.py
jandom: These are different snapshots – we have two types 
- the VCR snapshots for web requests 
- ndarrays_regression snapshots for numerics 

I'd be weary of mixing those up – they mean a different thing and solve a different problem

openfold3/tests/test_triangular_attention.py
jandom: Oh, this is great – in general I think something like this will be a good fit for these tests, because they're very similar to each other, and we'll duplicate a lot of code between different module tests.

openfold3/tests/test_triangular_attention.py
```diff
@@ -12,38 +12,61 @@
...
+    tan = TriangleAttention(
+        c_z,
+        c,
+        no_heads,
+        starting=starting,
+    ).to(device)
+    # AlphaFold initializes the output projection to zero (so residual blocks
+    # start as identity). Reinitialize all params so the test exercises the
+    # actual computation and produces non-trivial output.
+    for p in tan.parameters():
```
jnwei, line 42: nice

jnwei (approved):
Awesome work on this @jandom ! The fixtures will make it much easier to add regression test going forward.

A few overall suggestions:
- Could you update the overall summary of the PR, to reflect the new fixtures for device / rng? Eventually we should make docs in the dev section to describe some of these fixtures, but I don't think we need to do that here.
- Regarding the Snapshot environment - are we limited to only one device snapshot / test set? It's fine if so, but perhaps we should instead have the snapshots on the A10 as we use those to run our daily GPU based tests.

PR #132, opened 2026-03-10
Fix organometallic conformer gen and custom CCD support

### LINKED ISSUE #109: [BUG] Custom CCD is not properly read
**Describe the bug**
When running inference with a custom CCD file, it is not properly respected by the query builder. This means that inference workflows with custom compounds will not work properly, whereas AlphaFold3 supports this workflow.

**To Reproduce**
Set `ccd_file_path` to a custom CCD that contains different chemical component IDs, then run an inference query.

**Expected behavior**
The inference pipeline should featurize the custom compounds as detailed in their CCD information.

**Additional context**
[This function](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/core/data/primitives/structure/query.py#L99) uses `get_residue_cached`, which obtains its information from `biotite.structure.info`. Biotite uses its internal CCD for these operations, not the user-supplied one.

### LINKED ISSUE #110: [BUG] Conformer generation struggles with organometallics like HEM
**Describe the bug**
Inference queries that use organometallics such as HEM groups will run into conformer timeout problems. This is because RDKit by default cannot seem to properly handle conformer generation for such compounds.

**To Reproduce**
```python
from openfold3.core.data.primitives.structure.query import structure_with_ref_mol_from_ccd_code

structure_with_ref_mol_from_ccd_code("HEM", chain_id="A")

> ConformerGenerationError: Failed to generate 3D coordinates
```

**Expected behavior**
A conformer should be generated without failure.

**Additional context**
Our training pipeline makes use of input sanitization by `pdbeccdutils`, which solves these conformer generation issues. Because of that, this issue is inference-specific.

ljarosch:
# Summary
Refactors inference-time conformer logic to be `pdbeccdutils`-based (fixes organometallic conformer generation issues), and adds support for custom CCD file.

# Changes
## 1) Custom CCD support
Custom CCDs are now read in using the new function [`update_biotite_ccd_from_file`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/primitives/structure/biotite_ccd.py#L105) and can be supplied in both `.cif` and `.bcif` format. This function will update Biotite's own internal CCD to the user-supplied one, resulting in automatic compatibility with any Biotite-based utilities (such as residue information, bond writing, etc.). In parallel we introduce careful piping to make Biotite's internal CCD work with `pdbeccdutils`, see [note](#note-about-ccd-behavior-going-forward).

> [!NOTE]
> `.cif` input requires a few minutes to be converted to `.bcif` format under the hood. We have a script for preconverting in [`preprocess_ccd_biotite.py`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/scripts/data_preprocessing/preprocess_ccd_biotite.py) that we point to in the documentation, but could consider making it a native openfold3 command.

> [!WARNING]
> Biotite's internal CCD represents a module-level state, which is currently only compatible with the `fork` multiprocessing method. We likely want to future-proof this as newer Python versions switch the default to `forkserver`. This PR introduces a [`_worker_init_with_ccd`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/framework/data_module.py#L554) for this purpose which will reapply the biotite CCD change to each worker, but we have upstream `DataModule` issues blocking actual compatibility with forkserver/spawn that seemed out of scope for this PR.

### Note about CCD behavior going forward
The idea going forward is that the canonical CCD is always handled through Biotite's internally stored one: If a CIFFile-like object is needed as in the template code, we use [`BiotiteCCDWrapper`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/primitives/structure/component.py#L687). If interfacing with `pdbeccdutils` is required, we can use [`mol_from_biotite_ccd`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/primitives/structure/component.py#L780), which will handle the piping from the Biotite CCD bCIF format into the raw CIF format that `gemmi` and `pdbeccdutils` expect.

While Biotite's internal CCD format and these conversions are a little awkward, it seemed worse to implement lots of parallel code flows depending on which input format is supplied, so this consolidation was the best solution.

## 2) Basing conformer processing on `pdbeccdutils`
`pdbeccdutils` contains useful input sanitization that will rescue the conformer generation of organometallics by setting proper coordination bond types. We also rely on this library for training-time conformer processing. This switches inference conformer processing from pure Biotite logic back to `pdbeccdutils` logic, using [`processed_reference_molecule_from_ccd_code`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/primitives/structure/query.py#L282) and [`mol_from_biotite_ccd`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/primitives/structure/component.py#L780). As `pdbeccdutils` adds some slight overhead due to running more functionalities than we technically require, this PR adds caching to these functions to speed them up.

> [!NOTE]
> We might decide on getting rid of `pdbeccdutils` altogether in the future, but its molecule sanitization proved a bit tricky to track down and maintain ourselves, so for now it seemed better to rely on a PDBe-developed library.

For additional robustness, this PR also adds support for fallback to CCD-derived ideal coordinates in case conformer generation fails.

## 3) Proactive dative bond conversion in CIF output
`pdbeccdutils` introduces `COORDINATION` bond types that Biotite cannot serialize in the `_chem_comp_bond` table. Previously, [`_create_cif_file`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/core/data/io/structure/cif.py#L256) caught the resulting `KeyError` and retried after converting bonds. This replaces that try/except with a proactive conversion of intra-residue dative bonds to `SINGLE` before writing, avoiding throwing a warning and making the behavior explicit.

# Related Issues
Closes #109
Closes #110 

# Testing
Added tests to [`test_structure_from_query.py`](https://github.com/aqlaboratory/openfold-3/blob/ljarosch/2026-02-12/fix/issue-109-110-conformer-custom-ccd/openfold3/tests/test_structure_from_query.py):
- HEM ligand conformer generation (regression test for organometallics)
- Custom CCD with modified atom order, verified through both the query and direct `ccd_code` input paths
- Conformer fallback to CCD Ideal coordinates with partial NaN handling

Tested full inference workflow with:

## Standard inference and HEM conformer
**query.json**

```json
{
	"queries": {
		"protein_with_hem": {
			"chains": [
				{
					"molecule_type": "protein",
					"chain_ids": "A",
					"sequence": "MVLSPADKTNVKAAWGKVGAHAGEYGAEALERMFLSFPTTKTYFPHFDLSH"
				},
				{
					"molecule_type": "ligand",
					"chain_ids": "B",
					"ccd_codes": "HEM"
				}
			]
		},
		"protein_with_atp": {
			"chains": [
				{
					"molecule_type": "protein",
					"chain_ids": "A",
					"sequence": "MTEYKLVVVGAGGVGKSALTIQLIQNHFVDEYDPT"
				},
				{
					"molecule_type": "ligand",
					"chain_ids": "B",
					"ccd_codes": "ATP"
				}
			]
		}
	}
}
```

**runner.yml**

```yml
data_module_args:
  num_workers: 2

model_update:
  presets:
    - predict
    - low_mem
    - pae_enabled
```

## Custom CCD
CCD file was modified to contain a ligand named TEST1.

**query.json**

```json
{
	"queries": {
		"protein_with_test1": {
			"chains": [
				{
					"molecule_type": "protein",
					"chain_ids": "A",
					"sequence": "MTEYKLVVVGAGGVGKSALTIQLIQNHFVDEYDPT"
				},
				{
					"molecule_type": "ligand",
					"chain_ids": "B",
					"ccd_codes": "TEST1"
				}
			]
		}
	}
}
```

**cif-based runner.yml**

```yml
# Test runner for custom CCD via .cif path
# Tests that the .cif -> .bcif auto-conversion works end-to-end

data_module_args:
  num_workers: 0

dataset_config_kwargs:
  ccd_file_path: ./custom_components.cif

model_update:
  presets:
    - predict
    - low_mem
    - pae_enabled
```

**bcif-based runner.yml**

```yml
# Test runner for custom CCD via .bcif path (pre-converted, fast)
# Tests the direct .bcif code path

data_module_args:
  num_workers: 0

dataset_config_kwargs:
  ccd_file_path: ./custom_components.bcif

model_update:
  presets:
    - predict
    - low_mem
    - pae_enabled
```


openfold3/core/data/framework/single_datasets/inference.py
```diff
@@ -112,12 +114,15 @@ def __init__(
...
-            logger.debug("Parsing CCD file.")
-            self.ccd = pdbx.CIFFile.read(dataset_config.ccd_file_path)
-        else:
-            self.ccd = BiotiteCCDWrapper()
+        # If a custom CCD file is provided, overwrite Biotite's global CCD.
+        # The resolved path is stored for re-applying in DataLoader workers
+        # started with spawn/forkserver.
+        self._biotite_ccd_path = update_biotite_ccd_from_file(
+            dataset_config.ccd_file_path
+        )
```
jnwei, line 122: Will this run into an issue if the user provided CCD only contains customized structures, but not standard structures like the residues?

openfold3/core/data/primitives/structure/biotite_ccd.py
jnwei, line 1: If this file contains content copied from the Biotite library, can we add a reference about which file(s) were copied from biotite, and what our changes are on top of their functions?

openfold3/tests/test_structure_from_query.py
```diff
@@ -134,3 +228,149 @@ def test_smiles_with_explicit_hydrogen():
...
+    """Assert both outputs use the expected custom atom-name order."""
+    atom_names = structure_with_ref_mols.atom_array.atom_name.tolist()
+    reference_mol = structure_with_ref_mols.processed_reference_mols[0].mol
+    reference_atom_names = [
+        atom.GetProp("annot_atom_name") for atom in reference_mol.GetAtoms()
+    ]
+
+    assert atom_names == CUSTOM_CCD_EXPECTED_ATOM_NAMES
+    assert reference_atom_names == CUSTOM_CCD_EXPECTED_ATOM_NAMES
+    assert atom_names == reference_atom_names
```
jnwei, line 287: nit: This 3rd assertion seems redundant with the two assertions before it

openfold3/tests/test_structure_from_query.py
```diff
@@ -134,3 +228,149 @@ def test_smiles_with_explicit_hydrogen():
...
+    [
+        "query",
+        "ccd_code",
+    ],
+    ids=[
+        "query_uses_global_biotite_ccd",
+        "structure_with_ref_mol_from_ccd_code_aligns_atom_order",
+    ],
+)
+def test_ligand_ccd_paths_respect_custom_ccd_and_atom_order(input_type):
```
jnwei, line 301: nit: Since the verification is just one line anyway using `assert_atom_names_align_with_reference_mol`, does it make sense to write this as two different tests, rather than one test with parameterization?

openfold3/tests/test_structure_from_query.py
```diff
@@ -134,3 +228,149 @@ def test_smiles_with_explicit_hydrogen():
...
+        else:
+            structure_with_ref_mols = structure_with_ref_mol_from_ccd_code(
+                ccd_code=CUSTOM_CCD_CODE,
+                chain_id="A",
+            )
+
+    _assert_atom_names_align_with_reference_mol(structure_with_ref_mols)
+
+
+def test_conformer_fallback_to_ideal_with_partial_nan(monkeypatch):
```
jnwei, line 329: nit: monkeypatch seems unused here in favor of unittest.patch, can we remove it?

openfold3/tests/test_structure_from_query.py
```diff
@@ -36,6 +52,84 @@
 
 reference_data_path = Path(__file__).parent / "test_data" / "structure_from_query"
 
+# Custom CCD with alanine renamed to "GLU", with atoms in deliberately non-standard row
```
jnwei, line 55: This is a good test case. I like that the full cif is written in the test file for easy viewing within this test.

To expand my question regarding the usage of custom CCDs: Perhaps we could include one more test case that includes both a custom structure (provided by the this custom CCD cif file) and a standard residue? So for example, Valine + custom GLU?

openfold3/core/data/primitives/structure/component.py
```diff
@@ -717,3 +717,129 @@ def __getitem__(self, category: str):
...
+        An AnnotatedMol with atom names and (potentially) ideal/model conformers.
+        Despite caching, the returned Mol will be a new object that can be safely
+        modified without affecting future calls with the same CCD code.
+    """
+    mol = _mol_from_biotite_ccd_cached(ccd_code)
+
+    # Copy to avoid issues with mutable objects
+    mol_cp = Chem.Mol(mol)
+
+    return mol_cp
```
jnwei, line 820: While the use of the functional `lru_cache` around another function in line is technically correct, it looks a bit strange with the hanging docstring. It is much more common to see `lru_cache` as a decorator around an internal function.

Could we rewrite these two functions slightly to follow this pattern instead?

```
@lru_cache(maxsize=500)
def _mol_from_biotite_ccd_cached(ccd_code: str) -> AnnotatedMol:
    # internal, do not call directly — returns shared cached object
    ...

def mol_from_biotite_ccd(ccd_code: str) -> AnnotatedMol:
    """...Returns a safe copy; cached internally."""
    return Chem.Mol(_mol_from_biotite_ccd_cached(ccd_code))
```

openfold3/core/data/primitives/structure/component.py
```diff
@@ -717,3 +717,129 @@ def __getitem__(self, category: str):
...
+
+
+_get_residue_cached = lru_cache(maxsize=500)(struc.info.residue)
```
jnwei, line 823: Same thing with `_get_residue_cached` and the `lru_cache`

openfold3/tests/test_structure_from_query.py
```diff
@@ -134,3 +228,149 @@ def test_smiles_with_explicit_hydrogen():
...
+    # NaN positions should be zeroed, not NaN
+    assert np.isfinite(positions).all()
+    assert np.allclose(positions[0], [0, 0, 0])
+
+    # used_atom_mask should be False for the NaN atom
+    masks = [atom.GetBoolProp("annot_used_atom_mask") for atom in ref_mol.GetAtoms()]
+    assert masks[0] is False
+    assert all(masks[1:])
+
+    # Featurization should succeed without NaN
```
jnwei, line 371: nice that this test also tests featurization runs

jnwei (changes requested):
Overall really nice work  @ljarosch ! The CCD issue / organometallic conformers has been a nasty issue to debug and iron out. I think your overall strategy of preferring the biotite CCD and then converting custom CCDs into biotite format makes sense, thank you for providing the explanation.

One main question regarding how to use the custom CCD: If a user has a structure for one molecule (for example, a custom ligand) but wishes to use standard structures from the CCD for everything else (e.g. for standard amino acids), will the custom CCD be "appended" to the standard CCD? Or will the user need to provide their own CCD with the standard aas included?

It could be helpful to provide an explicit example of using a custom CCD in our huggingface directory, so that we give users a model of how to use this route in addition to documentation. Perhaps this example could be the example you used in inference testing.

Otherwise, I have just a few small fixes in the caching definitions and tests.

docs/source/input_format.md
```diff
@@ -266,8 +266,7 @@ Below is a complete example of an input JSON file specifying a single bioassembl
                 }
             ],
         }
-    },
-    "ccd_file_path": "/path/to/CCD/file.cif"
```
jandom, line 270: Wait so this will no longer be supported?

examples/reference_full_config/full_config.yml
```diff
@@ -158,4 +158,4 @@ template_preprocessor_settings:
   structure_array_directory: null
   cache_directory: <tmp-dir>/of3_template_data/template_cache
   log_directory: null
-  ccd_file_path: null
```
jandom, line 161: Ah, still there but moved up

openfold3/core/data/framework/data_module.py
```diff
@@ -545,6 +548,20 @@ def setup(self, stage=None):
...
+        """Wrap the base worker init to re-apply the custom Biotite CCD path.
+
+        The custom CCD setting in Biotite is process-local global state, so we reapply
+        it in each worker for future-proofing with spawn/forkserver.
+        """
+        self._base_worker_init(worker_id, rank)
+        dataset = torch.utils.data.get_worker_info().dataset
+        ccd_path = getattr(dataset, "_biotite_ccd_path", None)
+        if ccd_path is not None:
+            update_biotite_ccd(ccd_path)
```
jandom, line 564: I see Claude doing this pattern a lot – and I don't love it – a better pattern would be to declare _biotite_ccd_path as a nullable attribute that's defaulting to None.

openfold3/core/data/primitives/structure/biotite_ccd.py
jandom: Ditto – let's vendor as little code as humanly possible, do we need to vendor all of it? Which parts are we changing?

openfold3/core/data/primitives/structure/cleanup.py
```diff
@@ -1091,9 +1091,12 @@ def convert_intra_residue_dative_to_single(atom_array: AtomArray):
     Returns:
         Copy of the AtomArray with intra-residue dative bonds converted to single bonds.
     """
-    # Copy AtomArray
-    atom_array = atom_array.copy()
+    # Fast path: skip if no coordination bonds exist
+    bondlist_arr = atom_array.bonds.as_array()
+    if not np.any(bondlist_arr[:, 2] == BondType.COORDINATION):
+        return atom_array
```
jandom, line 1097: This is good

openfold3/core/data/primitives/structure/component.py
```diff
@@ -717,3 +717,129 @@ def __getitem__(self, category: str):
                 )
 
             return result
+
+
+def _bcif_to_cif_category(bcif_category) -> CIFCategory:
```
jandom, line 722: Is this vendored code again?

openfold3/core/data/primitives/structure/query.py
```diff
@@ -276,6 +279,66 @@ def processed_reference_molecule_from_mol(
...
+
+    # Compute conformer, falling back to Ideal coordinates on failure
+    try:
+        mol, conf_id, _ = multistrategy_compute_conformer(
+            mol, remove_hs=True, timeout_standard=120, timeout_rand_init=120
+        )
+        assert conf_id == 0
+    except ConformerGenerationError:
+        if ideal_conf is None:
+            raise
```
jandom, line 323: This is good

openfold3/core/data/primitives/structure/query.py
```diff
@@ -276,6 +279,66 @@ def processed_reference_molecule_from_mol(
...
+        logger.warning(
+            "Conformer generation failed for %s; "
+            "falling back to CCD Ideal coordinates.",
+            ccd_code,
+        )
+        mol = set_single_conformer(mol, ideal_conf)
+
+    # Shouldn't be necessary for ideal coordinates but better to be safe
+    mol = add_conformer_atom_mask(mol)
+    replace_nan_coords_with_zeros(mol)
```
jandom, line 333: What's this guarding against? After conformer generation, we should either have an ideal conformer/a valid generated conformer. The shouldn't have nans, surely?

openfold3/tests/test_structure_from_query.py
jandom: we should either be using mock.patch or monkeypatch, not both, for consistency

scripts/data_preprocessing/preprocess_ccd_biotite.py
```diff
@@ -1,202 +1,17 @@
...
-    mask_bool = mask == MaskValue.PRESENT
-    values = string_array[mask_bool]
-    try:
-        values = values.astype(int)
-    except ValueError:
-        with contextlib.suppress(ValueError):
-            values = values.astype(float)
-    array = np.zeros(string_array.shape, dtype=values.dtype)
-    array[mask_bool] = values
-    return array
```
jandom, line 197: Bye-bye code always good

jandom (changes requested):
Overall, i'm really happy this is here but some work still remains. A lot of code was added here and we need to make sure this really works.

openfold3/core/data/primitives/structure/biotite_ccd.py
ljarosch: Yeah I can be more specific about it. Re code vendoring, I think we could remove some of these functions. All functions below `concatenate_ccd` are taken from biotite directly, and `concatenate_ccd` itself was only slightly modified to work with a local path instead of pulling from the web.

Re code vendoring I was thinking that adding these util functions makes it self-contained, as they are all "private" functions in biotite, so maybe not API-stable. But happy to change this back to importing them, it does introduce some awkwardness and I'm unsure about the best practice.

openfold3/core/data/primitives/structure/component.py
ljarosch: No this is our own code

openfold3/core/data/primitives/structure/query.py
ljarosch: If our own conformer generation fails we fall back to the CCD-derived conformer, which can sometimes have missing coordinates. So far I've only ever seen this for "model" coordinates, not "ideal" coordinates, but technically the [PDB spec](https://mmcif.wwpdb.org/dictionaries/mmcif_ma.dic/Items/_chem_comp.pdbx_ideal_coordinates_missing_flag.html) indicates that it may be possible for coordinates to be "missing or incomplete" also for ideal conformers, so I thought better safe than sorry if PDB doesn't guarantee this.

I could pull this into the ideal conformer clause to be more precise, as it shouldn't apply to RDKit-generated conformers.

openfold3/tests/test_structure_from_query.py
ljarosch: right sorry, that's just a leftover I forgot to clean up

openfold3/core/data/primitives/structure/query.py
jandom: > I could pull this into the ideal conformer clause to be more precise, as it shouldn't apply to RDKit-generated conformers.

I'd be in favor of that, because i like sane logic – your explainer about would also make for a terrific comment to anybody studying this code in the future

openfold3/core/data/framework/data_module.py
```diff
@@ -545,6 +548,20 @@ def setup(self, stage=None):
             dist.broadcast_object_list(placeholder, src=0)
             self.inference_config.query_set = placeholder[0]
         super().setup()
+        self._base_worker_init = self.worker_init_function_with_data_seed
+        self.worker_init_function_with_data_seed = self._worker_init_with_ccd
```
bcm49:
```suggestion
        self.worker_init_function_with_data_seed = self._worker_init_with_ccd
        self._base_worker_init = self.worker_init_function_with_data_seed
```
This needs to be switched so self.worker_init_function_with_data_seed is defined before it is used, right?

jandom:
@ljarosch wants to revisit this PR in a week or two, dropping to draft as it's still in progress

jandom:
Closing due to lack of activity – we have some tweak there via https://github.com/aqlaboratory/openfold-3/pull/197

PR #138, opened 2026-03-17
docs: Add documentation around environment variables

etowahadams:
Adds documentation around environment variables. Setting `CUDA_HOME`, `CUTLASS_PATH`, `LD_LIBRARY_PATH`, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ pip install openfold3
 setup_openfold
 ```
 
-3. Run your first prediction using the ColabFold MSA server with the `run_openfold` binary. It may also be necessary to configure your environment variables depending on your system.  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ pip install openfold3
 setup_openfold
 ```
 
-3. Run your first prediction using the ColabFold MSA server with the `run_openfold` binary
+3. Run your first prediction using the ColabFold MSA server with the `run_openfold` binary. It may also be necessary to configure your environment variables depending on your system.
```
jandom: Would add a ref here to the document just above – because a reader will see this and go "well how do I do that?"

etowahadams: added!

jnwei (approved):
This is really nice, thank you @etowahadams !

PR #145, opened 2026-03-23
tests: query_colabfold_msa_server

jandom:
**Summary**

Quick cleanup of this pretty crucial method and a test

**Changes**

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -74,7 +74,7 @@ jobs:
 
       - name: Extract metadata (tags, labels) for Docker
         id: meta
-        uses: docker/metadata-action@dde548f1603e003959483c86c72ec03454216f9c
+        uses: docker/metadata-action@54262e8fce11c7b5496f28398368170da45de7e7
         with:
           images: |
             name=ghcr.io/${{ github.repository }}/openfold3-docker
--- a/docs/source/configuration_reference.md
+++ b/docs/source/configuration_reference.md
@@ -131,6 +131,10 @@ data_module_args:
 ```
 
 ---
+### 3.X Checkpoint Confiugration (`checkpoint_config`)
+
+Configures Checkpoint writing settings, which are passed to [pl.ModelCheckpoint callback](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.ModelCheckpoint.html). 
+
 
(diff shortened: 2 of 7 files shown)


openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -521,3 +522,56 @@ def test_cli_output_dir_conflict_raises(self, tmp_path):
...
+_BARNASE_SEQ = (
+    "AQVINTFDGVADYLQTYHKLPDNYITKSEAQALGWVASKGNLADVAPGKSIGGDIFSNREGKLPGK"
+    "SGRTWREADINYTSGFRNSDRILYSSDWLIYKTTDHYQTFTKIR"
+)
+
+
+class TestQueryColabfoldMsaServer:
+    """Functional test — hits real ColabFold API (~30-60s)."""
+
+    def test_barnase_with_templates(self, tmp_path):
```
jnwei, line 540: I bet we could reuse the ColabFold mmseqs mock itself : 
Mock definition: https://github.com/sokrypton/ColabFold/blob/main/tests/mock.py#L109
Usage: https://github.com/sokrypton/ColabFold/blob/9712f2ff262d3977d571919317e06cc96c29cd95/tests/test_msa.py#L7

That would probably be the cleanest way to handle it.

jandom: We would have to vendor it though :/ Not ideal – ColabFold is a very bulk dep to add, i'd rather avoid it

jandom: We've also modified their code a fair bit, i think

jandom: Hey, there @jnwei let me re-vive this PR – I think we can do the pytest-VCR approach to these tests

Ah, script that – we need the full colabfold job to complete. They appear to be using some sort of caching, subsequent calls are instant basically

jandom: Just reviewed the mock code you provided, yeah let's vendor that – it'll be the easiest.

jandom: Looking at this, the vendoring looks like a slog – I'd honestly just close this PR out rather than it in here.

PR #143, opened 2026-03-23
feature: pixi-beta

jandom:
Completing the work started in https://github.com/aqlaboratory/openfold-3/pull/34

Remaining tasks 
- [x] fix random ruff problems
- [x] update the docker build to use both pixi and conda
- [x] confirm that inference can be done
- [ ] confirm that training can be done (decision: deferred)

PR #148, opened 2026-03-26
Fixes for forkserver/spawn serialization and fix for LMDB upgrade issues

christinaflo:
**Summary**
LMDB refactor to allow for forkserver/spawn serialization and resolve issues that required pinning to LMDB==1.6.2 for training.

**Changes**
1. Exposes more dataloader options like persistent workers, prefetch factor, multiprocessing context
2. Lazy load CCD obj to prevent issues with serialization
3. Modify LMDB handling to allow for serialization. Close parent db connection prior to forking.

**Related Issues**
PR #143 defaults to forkserver. This PR adds additional fixes needed for training with forkserver.


jandom (commented):
I'm not sure we need this any more with the changes in #143 – the two small reads from LMDB will be immediately released now (via context manager) and the persistent lmdb env is staying for the entire duration of the data loader (hopefully!)

but I'm 100% confident – training tends to reveal more problems than running unit tests

Do you have any repro/test that I could run to confirm that my claim is accurate?

christinaflo:
These changes address a different issue than the one addressed in #143, the original version is not serializable because of the persistent lmdb env if you use forkserver/spawn (the __getstate__ and __setstate__ are needed). Also forking the persistent env across workers causes weird behavior/failures with versions of lmdb > 1.6, it hasnt worked for me on multiple systems which is why we pinned it. They added some documentation about this recently: https://lmdb.readthedocs.io/en/latest/#forking-multiprocessing

christinaflo:
> Minor comment: the DB preloading logic might be better handled by using [vmtouch](https://hoytech.com/vmtouch/), which is available in conda‑forge and often present on HPC systems, falling back to reading the whole DB file only when vmtouch is not available. Conceptually, this would only need to be done once per node, since the page cache is shared across processes.
> 
> If using vmtouch, one could optionally consider daemon mode and page locking to reduce the risk of eager eviction under memory pressure, although this may or may not be desirable depending on system limits and sharing policies (I assume we typically reserve full nodes?). Overall, I suspect the usefulness of this depends strongly on DB size vs memory size, access locality, and competition for memory.
> 
> I just found [this](https://github.com/lokeshdevnani/pypagecache) which could be vendored. I never used it.

I've never used it before, it isnt currently available on the cluster im using but happy to try it out via conda forge. The biggest db we have is around 10 GB. On one system, i did notice that I had to "rewarm" the cache between epochs so this could be useful to avoid that.

jnwei (commented):
LGTM! could you also leave a brief descriptions of the systems that you ran the tests on?

jnwei:
It looks like we can't run the multiprocessing tests on LMDB -- can you pytest.mark them as slow tests?

christinaflo:
Ran tests and tested model training on GH200 with cuda 12.6, ran tests locally on M4

PR #152, opened 2026-03-28
fix: namespace temp directories by username to avoid multi-user conflicts

### LINKED ISSUE #150: [BUG] Multi-user machine /tmp/of3-directories cause errors
**Describe the bug**

Several default temporary directories use a shared, fixed path under `/tmp/` (via `tempfile.gettempdir()`), which causes `PermissionError` on multi-user machines. When user A creates e.g. `/tmp/of3_colabfold_msas/` (owned by A with default `755` permissions), user B cannot write to it.

This affects at least:

| Default path | Location |
|---|---|
| `/tmp/of3_colabfold_msas/` | `openfold3/core/data/tools/colabfold_msa_server.py:1000` |
| `/tmp/of3_template_data/` | `openfold3/core/data/pipelines/preprocessing/template.py:1605` |
| `/tmp/ramdisk/` | `openfold3/core/data/tools/jackhmmer.py:206` |

The problem surfaces easily when two users run the test suite on the same machine, since tests exercise these code paths and create these directories.

**To Reproduce**

This requires two user accounts on the same machine (e.g. `alice` and `bob`):

```bash
# As alice:
pytest openfold3/tests/ -x     # creates /tmp/of3_colabfold_msas/ owned by alice

# As bob:
pytest openfold3/tests/ -x     # PermissionError writing to /tmp/of3_colabfold_msas/
```

**Expected behavior**

Each user should get their own isolated temporary directory, so concurrent or sequential usage by different users does not conflict.

**Suggested fix**

`pytest` itself solves this exact problem by namespacing its temp directories per user:

```
/tmp/pytest-of-alice/
/tmp/pytest-of-bob/
```

OpenFold3 could adopt the same pattern, e.g.:

```python
import getpass
Path(tempfile.gettempdir()) / f"of3-{getpass.getuser()}" / "colabfold_msas"
```

Alternatively, use `tempfile.mkdtemp()` to create unique directories per run, which is already done safely in other parts of the codebase (e.g. `openfold3/core/data/tools/utils.py:29`).

 - OS: any multi-user Linux system
 - Python version: any
 - Installation path: any

**Additional context**

The hardcoded `/tmp/ramdisk/` path in `jackhmmer.py:206` has the same class of issue but is additionally problematic because the path doesn't include any openfold-specific prefix, increasing the chance of collisions with other software.

Alm0stSurely:
## Problem

As reported in #150, OpenFold3 uses hardcoded paths under `/tmp/` for temporary directories. On multi-user machines, this causes `PermissionError` when user A creates a directory (e.g., `/tmp/of3_colabfold_msas/`) and user B tries to write to it.

### Affected paths:
| Default path | Location |
|---|---|
| `/tmp/of3_colabfold_msas/` | `openfold3/core/data/tools/colabfold_msa_server.py:1000` |
| `/tmp/of3_template_data/` | `openfold3/core/data/pipelines/preprocessing/template.py:1605` |
| `/tmp/ramdisk/` | `openfold3/core/data/tools/jackhmmer.py:206` |

## Analysis

The issue occurs because:
1. Directories are created with default permissions (755)
2. The paths are identical for all users on the same machine
3. When user A creates the directory first, user B gets a permission denied error

## Solution

Namespace the temporary directories by username, following the pattern used by pytest (`/tmp/pytest-of-{user}/`):

```
/tmp/of3-{username}/colabfold_msas/
/tmp/of3-{username}/template_data/
/tmp/of3-{username}/ramdisk/
```

This ensures each user gets their own isolated directory hierarchy, preventing permission conflicts while maintaining backward compatibility (users can still override via configuration).

## Changes

- `colabfold_msa_server.py`: Added `getpass` import, namespaced the default MSA output directory
- `template.py`: Added `getpass` import, namespaced the default template data directory  
- `jackhmmer.py`: Added `getpass` and `tempfile` imports, namespaced the ramdisk directory with automatic directory creation

## Testing

- Verified syntax with `python3 -m py_compile` on all modified files
- Tested path generation logic confirms proper user isolation
- Confirmed no hardcoded `/tmp` paths remain in the modified files

## Notes

- This fix is minimal and non-breaking — users who explicitly set output directories are unaffected
- The pattern mirrors pytest's proven approach for temp directory isolation
- Fixes #150

### DIFF
--- a/openfold3/core/data/pipelines/preprocessing/template.py
+++ b/openfold3/core/data/pipelines/preprocessing/template.py
@@ -14,6 +14,7 @@
 
 """Preprocessing pipelines for template data ran before training/evaluation."""
 
+import getpass
 import logging
 import multiprocessing as mp
 import os
@@ -1602,7 +1603,7 @@ class TemplatePreprocessorSettings(BaseModel):
             )
 
         self.output_directory = (
-            self.output_directory or Path(tempfile.gettempdir()) / "of3_template_data"
+            self.output_directory or Path(tempfile.gettempdir()) / f"of3-{getpass.getuser()}" / "template_data"
         )
         base = self.output_directory
 
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import getpass
 import json
 import logging
 import os
@@ -997,7 +998,7 @@ class MsaComputationSettings(BaseModel):
     server_user_agent: str = "openfold"
     server_url: Url = Url("https://api.colabfold.com")
     save_mappings: bool = True
-    msa_output_directory: Path = Path(tempfile.gettempdir()) / "of3_colabfold_msas"
+    msa_output_directory: Path = Path(tempfile.gettempdir()) / f"of3-{getpass.getuser()}" / "colabfold_msas"
     cleanup_msa_dir: bool = True
 
     @model_validator(mode="after")
--- a/openfold3/core/data/tools/jackhmmer.py
+++ b/openfold3/core/data/tools/jackhmmer.py
@@ -15,11 +15,14 @@
 
 """Library to run Jackhmmer from Python."""
 
+import getpass
 import glob
 import logging
 import os
 import subprocess
+import tempfile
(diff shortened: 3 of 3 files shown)


jandom:
Thank you for this contribution, we've combined it with #151 and both are superseded by https://github.com/aqlaboratory/openfold-3/pull/158

PR #165, opened 2026-04-02
dtype options float32/16 for full confidence npz output

rizsp:
**Summary**

Allows explicitly setting the dtype for the writeout of the full confidence for npz files to either float16 or float32 (default is float32), by setting this inside a runner.yaml like this:

```yaml
output_writer_settings:
  full_confidence_output_format: npz
  full_confidence_output_dtype: float16
  write_full_confidence_scores: true
```

**Changes**

* Added an option to cast to float16 or float32 for npz writeout inside `writer.py`
* Documentation + typo fix

**Related Issues**

**Testing**

Added parameter combinations to `test_writer.py` and changed `assert_array_equal` to `assert_array_almost_equal`. 

(While using `almost_equal` is clearly required for the float16 option, the float32 option also requires this since the fixture matrices are initialized as float64 due to `np.random.uniform`. In practice, `pae`, `pde` and `plddt` are predicted in float32 however, so the default float32 option would lead to exactly equal values)

**Other Notes**

The PAE and PDE matrices scale with N^2, so their output files can be very large, particularly if written out as json. Writing them out as npz with float16 cuts down file sizes somewhat. 

A next possible step might be to quantize them to uint8, similar to how the official AFDB/EMBL [does it for the PAE](https://www.alphafold.ebi.ac.uk/faq) (they output a JSON with the quants and a range and expect the user to postprocess/dequantize the values). uint8 would have the added benefit of being easy to compress.

### DIFF
--- a/docs/source/configuration_reference.md
+++ b/docs/source/configuration_reference.md
@@ -208,7 +208,7 @@ Configures the format of output files.
 **All Options**:
 - `structure_format` *(Literal["pdb", "cif", "cif.gz"])*: Output format (default: `cif`)
 - `full_confidence_output_format` *(Literal["json", "npz"])*: Confidence output format (default: `json`)
-- `full_confidence_output_dtype` *(Literal["float32", "float16"])*: Data type for confidence scores when using npz format (default: `float32`)
+- `full_confidence_output_dtype` *(Literal["float32", "float16"])*: Data type for confidence scores when using npz format (default: `float16`)
 - `write_features` *(bool)*: Write intermediate features (default: `false`)
 - `write_latent_outputs` *(bool)*: Write model intermediate outputs (default: `false`)
 - `write_full_confidence_scores` *(bool)*: Write full confidence scores, e.g. PAE, PDE, PLDDT (default: `true`)
--- a/docs/source/configuration_reference.md
+++ b/docs/source/configuration_reference.md
@@ -208,7 +208,7 @@ Configures the format of output files.
 **All Options**:
 - `structure_format` *(Literal["pdb", "cif", "cif.gz"])*: Output format (default: `cif`)
 - `full_confidence_output_format` *(Literal["json", "npz"])*: Confidence output format (default: `json`)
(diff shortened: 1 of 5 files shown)


jnwei (approved):
Thanks again for the amazing contribution @rizsp complete with docs and tests. I can see that the defaults settings to output the full confidence scores at float32 in json format might cause issues for folks who have large structure queries.

What do you think of changing the default output type to float16 instead? Or do you have suggestions for other default settings? I'm not familiar with the uint8 optimization you mentioned that AFDB uses, could you explain a bit further on how that would work for PAE values that are not integers?

rizsp:
I set the default to float32 since this is the current behavior. Changing the default to float16 makes sense. float16 would be around 3 decimal places which is probably sufficient for most users. In this case, the default output format could also be set to npz.

I didn't touch JSON output related code. In that case, it would make more sense to talk about rounding to decimals instead of the datatype since it's a string output. Perhaps an implicit setting? Such as, float16 corresponds internally to rounding to 3 decimals, and float32 to rounding to 6 decimals?

I reread the PAE section in the AFDB FAQ and I must have misunderstood. It seems they really just do a round to nearest integer, while I assumed they would have a proper quantization scheme (why else would they also return the max value).

They way I imagine it is to do a simple affine quantization. uint8 would mean 256 bins, we'd then linearly map the floats to the bins while taking into account a min and max value. To dequantize the binned/quantized value, the user would need to know the number of bins (256) and the range (min and max value) and postprocess the npz themselves. As an example, binning the range 0 to 32 would lead to an effective bin precision of 0.125.

jnwei:
Hi @rizsp per your suggestions I've ammended your PR slightly to make the default float16 and to change the output rounding to 3 decimals and 6 decimals for the json print out.

Thank you for your explanation of the uint8 quantization. Given that our [pae / plddt heads already predict logits / bin](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/core/metrics/aggregate_confidence_ranking.py#L61-L65) this quantization might not be a big lift to add. We can consider this in a future PR if there's interest

rizsp:
@jnwei, thank you for implementing the suggestions! If I find the time, I can also try writing a PR for a quantized output since I myself would be using it

PR #168, opened 2026-04-07
fix: don't write invalid JSON (no NaNs for resolution)

### LINKED ISSUE #167: [BUG] Dataset cache JSON files contain invalid `NaN` literals
## Summary

The training and validation dataset cache JSON files on S3 contain bare `NaN` literals (e.g. `"resolution": NaN`), which are not valid JSON. This causes strict JSON parsers like `ijson` to fail when streaming these files. 'ijson' is handy because these JSON files are a few GBs and often hard to load into memory. 

## Affected files

Probably all the JSON metadata files, stumbled on these two building the mini-PDB dataset. 

```
s3://openfold3-data/pdb_training_set/dataset_caches/training_cache_with_templates.json
s3://openfold3-data/pdb_training_set/dataset_caches/validation_cache_with_templates.json
```

## Reproduce

```bash
aws s3 cp s3://openfold3-data/pdb_training_set/dataset_caches/training_cache_with_templates.json . --no-sign-request
```

```python
import ijson

pdb_ids = []
with open("training_cache_with_templates.json", "rb") as f:
    parser = ijson.parse(f)
    for prefix, event, value in parser:
        if prefix == "structure_data" and event == "map_key":
            pdb_ids.append(value)
```

```
IncompleteJSONError: lexical error: invalid char in json text.
          ",             "resolution": NaN,             "chains": {
                     (right here) ------^
```

## Root cause

`get_resolution()` in `openfold3/core/data/primitives/structure/metadata.py` returns `float("nan")` for structures where no resolution data is found in the mmCIF (e.g. NMR structures). Python's `json.dump()` defaults to `allow_nan=True`, which writes the Python `NaN` as a bare `NaN` literal — valid Python/JavaScript, but invalid JSON.

## Fix

PR changes `get_resolution()` to return `None` instead of `float("nan")`, updates all downstream type annotations (`float` -> `float | None`), adds `allow_nan=False` to `json.dump()` to fail fast if any NaN slips through, and adds a NaN safety net in the JSON encoder.

## Workaround

For existing JSON files on disk:

```bash
sed -i.bak 's/\bNaN\b/null/g' training_cache_with_templates.json
sed -i.bak 's/\bNaN\b/null/g' validation_cache_with_templates.json
```

jandom:
**Summary**

- fixes #167
- this is a minor annoyance when preparing the mini-datasets, our JSON metadata is not actually valid JSON 

**Changes**

**Related Issues**

**Testing**

**Other Notes**


openfold3/tests/core/data/pipelines/featurization/test_loss_weights.py
jnwei, line 1: I'm not following how the tests in test_loss_weights are related to the json encoding changes for the DatasetCache. Maybe these changes were meant to be part of a different branch?

jandom: Just being paranoid that this code correctly handles cases where the resolution is None, these belong here – not strictly related to the JSON change, but more related to the internal signature changing from resolution being a float everywhere to being float | none

jnwei: I see, fair enough

PR #174, opened 2026-04-09
build: use scm tags for version numbers

jandom:
**Summary**

Small tweak to use the git/scm version tags – prevents us from having to manually update the version number when a new version tag is published. 

**Changes**

**Related Issues**

**Testing**

**Other Notes**


jnwei (approved):
very helpful, thanks for adding!

PR #178, opened 2026-04-13
Replace scipy's truncnorm.rvs with nn.init.trunc_normal_ to make model load faster

rizsp:
**Summary**

Via line-by-line timing I noticed that the `load_state_dict` inside `_warn_on_missing_version_tensor_in_load_statedict()` takes up around 27 seconds on my machine on initial model load. Further clicking around in the profile pointed to `trunc_normal_init_` which initializes truncated random variables according to kaiming or lecun using `scipy.stats.rvs`.

I adjusted the code so that it:
* Caches the `truncnorm.std` call which always uses the same arguments (an alternative would be a Magic Number... this would also eliminate the scipy import... see code comment)
* Replaces `scipy.stats.truncnorm.rvs` with `nn.init.trunc_normal_` which should be much faster since it initializes directly on the GPU instead of having to generate stuff on CPU (expensive CDF calls) and then move the random number array from CPU to GPU

On my machine this reduces the model load from 27 seconds to around 3 seconds (the file loading itself only takes a second).

**Changes**

* Changes `trunc_normal_init()` to use `nn.init.trunc_normal_` instead
* Removes unused `_prod`

**Related Issues**

**Testing**

I didn't spot any related test covering this function?

**Other Notes**

### DIFF
--- a/openfold3/core/model/primitives/initialization.py
+++ b/openfold3/core/model/primitives/initialization.py
@@ -16,18 +16,11 @@
 """Initialization functions for network parameters."""
 
 import math
+from functools import lru_cache
 
-import numpy as np
 import torch
 from scipy.stats import truncnorm
 from torch import nn
-
-
-def _prod(nums):
-    out = 1
-    for n in nums:
-        out = out * n
-    return out
 
 
 def _calculate_fan(linear_weight_shape, fan="fan_in"):
@@ -45,18 +38,26 @@ def _calculate_fan(linear_weight_shape, fan="fan_in"):
     return f
 
 
+@lru_cache
+def _cached_truncnorm_std(a, b, loc, scale):
+    return truncnorm.std(a, b, loc, scale)
+
+
 def trunc_normal_init_(weights, scale=1.0, fan="fan_in"):
-    shape = weights.shape
-    f = _calculate_fan(shape, fan)
+    f = _calculate_fan(weights.shape, fan)
+
     scale = scale / max(1, f)
-    a = -2
-    b = 2
-    std = math.sqrt(scale) / truncnorm.std(a=a, b=b, loc=0, scale=1)
-    size = _prod(shape)
-    samples = truncnorm.rvs(a=a, b=b, loc=0, scale=std, size=size)
-    samples = np.reshape(samples, shape)
+    # truncnorm.std is always 0.8796256610342398
+    std = math.sqrt(scale) / _cached_truncnorm_std(a=-2, b=2, loc=0, scale=1)
+
     with torch.no_grad():
-        weights.copy_(torch.tensor(samples, device=weights.device))
+        nn.init.trunc_normal_(
+            weights,
+            mean=0.0,
+            std=std,
+            a=-2.0 * std,
+            b=2.0 * std,
+        )
 
 
 def lecun_normal_init_(weights):
--- a/openfold3/core/model/primitives/initialization.py
+++ b/openfold3/core/model/primitives/initialization.py
@@ -48,7 +48,7 @@ def trunc_normal_init_(weights, scale=1.0, fan="fan_in"):
 
     scale = scale / max(1, f)
     # truncnorm.std is always 0.8796256610342398
-    std = math.sqrt(scale) / _cached_truncnorm_std(a=-2, b=2, loc=0, scale=1)
+    std = float(math.sqrt(scale) / _cached_truncnorm_std(a=-2, b=2, loc=0, scale=1))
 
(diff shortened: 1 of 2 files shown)


rizsp:
Sorry, I noticed some tests are failing which cover `lecun_normal_init`, I'll have to take a look

jandom:
Excellent contribution @rizsp – cherry on the cake, if you want, you could codify this with pytest-benchmark, ie test the time it takes to load a model. it'll catch any future regressions.

rizsp:
Okay, the tests were failing (failed to open libnvrtc-builtins.so.13.0) because the previous way of initializing random numbers on CPU with scipy didn't trigger the JIT. Once that part was changed to GPU, the JIT sprang in and my incomplete CUDA setup failed. Once I exported the CUTLASS and LD_LIBRARY_PATH to the right CUDA version, the tests passed again.

I can take a look into implementing a pytest benchmark.

rizsp:
@jandom I added a test to check the timing for the state dict loading and set it to fail for timings greater than 10 seconds (excluding the time to load from disk). It fails on main (takes 21s) and succeeds on this PR (takes 2s).

jandom:
This is excellent, thank you for resolving this so quickly (lightning speed!). My last worry would be "are we doing something to the weights by applying this procedure that we weren't doing before" – now obviously we can't do a snapshot of the entire checkpoint (2.1gb) with method A vs method B – but maybe there is something reasonable that we can do here to confirm that we're not breaking something, that the weights are indeed loaded the same just faster.

rizsp:
Since this is really just a tiny change of the code (changing the random initialization function for kaming/lecun) it should be easy to reason about.

This patch replaces scipy's `truncnorm.rvs` with `nn.init.trunc_normal_`. You actually already use methods from `nn.init` inside `initialization.py`, notably `nn.init.kaiming_normal_` and `nn.init.xavier_uniform_`. I assume you implemented `lecun_normal_init` and `he_normal_init` from scratch using scipy because `nn.init` does not have truncated versions of those.

As long as one can show that `nn.init.trunc_normal_` creates the same random distribution as `truncnorm.rvs`, the behavior should be the same? Maybe generating samples for both methods, and checking with a 2 sample Kolmogorov-Smirnov test is enough?

Here's just a sketch:

```py
torch_tensor = torch.empty(1024, 512)
std =  math.sqrt(1.0 / 512) / 0.8796

# new nn.init method
nn.init.trunc_normal_(torch_tensor, mean=0.0, std=std, a=-2.0*std, b=2.0*std)
# old scipy method
scipy_tensor = truncnorm.rvs(a=-2, b=2, loc=0, scale=std, size=torch_tensor.shape)

statistic, pvalue = ks_2samp(scipy_samples, torch_samples)
```

The pvalue in this snippet is 0.98 which speaks for the null hypothesis that the distributions are identical.
Or visually, as a plot:

image

jandom:
> As long as one can show that nn.init.trunc_normal_ creates the same random distribution as truncnorm.rvs, the behavior should be the same? Maybe generating samples for both methods, and checking with a 2 sample Kolmogorov-Smirnov test is enough?

This looks sane enough to me, thanks – I've tagged Christina for review, since she's more familiar with this

jandom:
@rizsp this is currently failing in tests – can you reproduce locally?

rizsp:
Sorry, you're right. If I see correctly in the logs, this is because TriangleAttention is instantiated which contains a Linear layer which is initialized with lecun. Since the lecun init is now done using nn.init instead of rvs, the random numbers will be different now. The ndarrays_regression.check checks against a preinitialized array created with the previous rvs init, probably stored in `test_data`.

I'm not sure what the right procedure is, but if you're sure that the init replacement is fine, someone on your side would probably just need to update the array in `test_data` using `pytest --force-regen`.

jnwei:
@rizsp do you think we should still add this PR now that we merged #220? It seems like the change to skip weight initialization you introduced in #220 already saved a lot of loading time for infernce. Will switching to nn.init.truc_normal_ have much more additional benefit?

rizsp:
@jnwei since random init during inference is practically instant/skipped now, the only benefit of this PR would be for the training phase where a 30 second one-time speed up might be considered negligible. As I see it, the benefit now would mainly be uniformity, so that all initialization functions would use torch.nn.init instead of a single one using a scipy based init.

However, this PR requires someone with your hardware to perform a `pytest --force-regen` for the specific test that fails, since the RNG will be different (GPU vs CPU). I don't have your hardware and these regression tests fail for me on the main branch anyway (on Ada 6000 and 5090).

PR #181, opened 2026-04-15
cuequivariance support in pixi (including aarch64)

jandom:
## Summary

Adds cuequivariance support for `linux-aarch64` (e.g. DGX Spark / GB10) and upgrades the version pin from `<0.8` to `>=0.8` across all platforms. This required adapting `_cueq_triangle_attn` for a breaking API change in cuequivariance 0.8+.

### `pixi.toml` / `pixi.lock`
- Added `linux-aarch64` pypi-dependency targets for both `cuequivariance-cuda12` and `cuequivariance-cuda13` features
- Unified all cuequivariance pins to `>=0.8` (previously `<0.8` on linux-64)
  - `<0.8` was originally pinned because of a triangle_attention bias shape API change — now fixed in the attention code
  - `0.7.0` crashes on import on GB10/aarch64 due to an upstream pynvml bug (fixed in 0.8+)
- All platforms now resolve to cuequivariance 0.9.1

### `openfold3/core/model/primitives/attention.py`
- Updated `_cueq_triangle_attn` for cuequivariance 0.8+ API: `triangle_attention` now requires `bias` shape `(B, 1, H, Q, K)` with exact batch dim match (no implicit broadcasting)
- **4D chunked inputs**: `chunk_layer` flattens batch dims, producing 4D tensors. These are promoted to 5D with `N=1` so each chunk entry is an independent batch item. Bias batch dim is expanded when `chunk_layer` kept it at `B=1`.
- **6D template inputs**: unchanged (already worked — the `view` preserves the `1` in bias)

### `openfold3/core/kernels/cueq_utils.py`
- Added `is_cuequivariance_installed()` — package-only check (no CUDA requirement)
- `is_cuequivariance_available()` refactored to call it internally (no behavior change)

### `openfold3/tests/compare_utils.py`
- `skip_unless_cueq_installed()` now produces distinct skip messages:
  - "Requires cuequivariance to be installed" (package missing)
  - "Requires CUDA (cuequivariance is installed but no GPU available)" (no GPU)
- Fixes typo "Equivaraince" → "cuequivariance"

## Testing

- All 12 cueq kernel tests pass on aarch64 / GB10 with cuequivariance 0.9.1
- Tested on `openfold3-cuda13-pypi` environment
- linux-64 CI should be unaffected (same packages, just a version bump from 0.7.0 to 0.9.1)

- JD: https://github.com/aqlaboratory/openfold-3/actions/runs/24459465405 🟢 manually triggered

## Known issue

N/A


openfold3/core/model/primitives/attention.py
```diff
@@ -746,40 +746,69 @@ def _cueq_triangle_attn(q, k, v, biases, scale):
...
+        # triangle_bias: (chunk, H, S, S) → (chunk, 1, H, S, S)
+        #   or: (1, H, S, S) → (1, 1, H, S, S) when chunk_layer kept B=1
+        triangle_bias = triangle_bias.unsqueeze(1)
+        # chunk_layer skips expanding bias when all its batch dims are 1,
+        # so bias may have B=1 while q has B=chunk. Expand to match.
+        if triangle_bias.shape[0] != q.shape[0]:
+            # (1, 1, H, S, S) → (chunk, 1, H, S, S)
+            triangle_bias = triangle_bias.expand(q.shape[0], *triangle_bias.shape[1:])
+        if mask_bias.shape[0] != q.shape[0]:
+            mask_bias = mask_bias.expand(q.shape[0], *mask_bias.shape[1:])
```
jandom, line 796: @christinaflo this migrates how we call a recent cueq - does it look sane? (it obviously doesn't :D)

jnwei: I think this is OK. IMO we're not going to get much cleaner without having separate functions depending on the number of dimensions in the input.

jnwei (approved):
LGTM! Tested the following on A100:

- openfold3-cuda12-pypi : 
    - all kernel tests
    - all unit tests. Everything passes except for 3 attention regression tests, which are locked to the GB10 instance
    - sample ubiquitin prediction
- openfold3-cuda13-pypi : all kernel tests pass

I added a few edits to the documentation for cuequivariance usage, namely use  `pixi run -e openfold3-cuda12-pypi` to run with cuequivariance, feel free to edit

jandom:
> LGTM! Tested the following on A100:

This is incredible, thank you so much!

openfold3/core/model/primitives/attention.py
borisfom: Well actually it's the right way to fix chunking when BS>1. But we do want to avoid expand() on bias for BS=1 (it becomes huge and we do allocate real memory for it on contiguous() call), so I suggest adding a few conditionals to process BS=1 case just like before : 
```
diff --git a/openfold3/core/model/primitives/attention.py b/openfold3/core/model/primitives/attention.py
index 46c924aa..dbe74e87 100644
--- a/openfold3/core/model/primitives/attention.py
+++ b/openfold3/core/model/primitives/attention.py
@@ -773,10 +773,12 @@ def _cueq_triangle_attn(q, k, v, biases, scale):
         triangle_bias = triangle_bias.view(batch * n_tmpl, *triangle_bias.shape[2:])

     # 4D → 5D: chunk_layer flattens batch dims and slices into chunks.
+    # chunk_layer skips expanding bias when all its batch dims are 1,
+    # so bias may have B=1 while q has B=chunk. In this case, we're good - otherwise:
     # Promote to 5D with N=1 so each chunk entry is an independent batch item.
     # cuequivariance >=0.8 requires bias shape (B, 1, H, Q, K) with exact
     # batch match — no implicit broadcasting.
-    is_chunked_input = len(q.shape) == 4
+    is_chunked_input = len(q.shape) == 4 and triangle_bias.shape[0] > 1
     if is_chunked_input:
         # q: (chunk, H, S, D) → (chunk, 1, H, S, D)
         q = q.unsqueeze(1)
@@ -787,8 +789,7 @@ def _cueq_triangle_attn(q, k, v, biases, scale):
         # triangle_bias: (chunk, H, S, S) → (chunk, 1, H, S, S)
         #   or: (1, H, S, S) → (1, 1, H, S, S) when chunk_layer kept B=1
         triangle_bias = triangle_bias.unsqueeze(1)
-        # chunk_layer skips expanding bias when all its batch dims are 1,
-        # so bias may have B=1 while q has B=chunk. Expand to match.
+        # This should not happen. Just in case. Expand to match.
         if triangle_bias.shape[0] != q.shape[0]:
             # (1, 1, H, S, S) → (chunk, 1, H, S, S)
             triangle_bias = triangle_bias.expand(q.shape[0], *triangle_bias.shape[1:])
@@ -803,11 +804,14 @@ def _cueq_triangle_attn(q, k, v, biases, scale):
     o = triangle_attention(q, k, v, bias=triangle_bias, mask=mask_bias, scale=scale)

     # Undo the promotions in reverse order.
-    if is_chunked_input:
+    if len(q.shape) == 4:
+        ##VS: There's a bug in cueq where if the input is missing the batch dim
+        ## the outputs adds it in and so we need to remove it here
+        o = o.squeeze(0)
+    elif is_chunked_input:
         # (chunk, 1, H, S, D) → (chunk, H, S, D)
         o = o.squeeze(1)
-
-    if is_batched_input:
+    elif is_batched_input:
         # (batch*n_tmpl, N, H, S, D) → (batch, n_tmpl, N, H, S, D)
         o = o.view(batch, n_tmpl, *o.shape[1:])
```

borisfom: @christinaflo : I have tested that code with BS=1 and BS=2 using test_kernels, and it worked for me: 
```
--- a/openfold3/tests/test_kernels.py
+++ b/openfold3/tests/test_kernels.py
@@ -437,8 +437,6 @@ class TestKernels(unittest.TestCase):
         batch_size = consts.batch_size
         if chunk_size is not None and (
             use_deepspeed_evo_attention
-            or use_cueq_triangle_kernels
-            or use_triton_triangle_kernels
         ):
```

(actually, triton_kernels also worked with BS>1, with minor accuracy error).

PR #184, opened 2026-04-16
tests: geometry (tier-1)

jandom:
**Summary**

- Improving test coverage of the geometry package, and the mypy annotations (since it's a leaf package, isolated)

**Changes**

- minimal code changes (typing annotations), and only tests added

**Related Issues**

**Testing**

**Other Notes**


openfold3/tests/utils/geometry/test_rigid_matrix_vector.py
```diff
@@ -0,0 +1,403 @@
...
+import torch
+
+from openfold3.core.utils.geometry.rigid_matrix_vector import Rigid3Array
+from openfold3.core.utils.geometry.rotation_matrix import Rot3Array
+from openfold3.core.utils.geometry.vector import Vec3Array
+
+from .helpers import rigid as _rigid
+from .helpers import rot_x as _rot_x
+from .helpers import rot_z as _rot_z
+from .helpers import v as _v
```
jnwei: Prefer to use the explicit import path, rather than the relative path, e.g. openfold3.tests.utils.geometry.helpers

openfold3/core/utils/geometry/vector.py
jnwei, line 1: nit: It would be slightly more clear if we annotated the return types that are scalars (formerly Float) to with something like

`def dot(self, other: Vec3Array) -> torch.Tensor:  # return: scalar`

openfold3/tests/utils/geometry/test_rigid_matrix_vector.py
```diff
@@ -0,0 +1,403 @@
...
+        assert rig.shape == (4,)
+
+    def test_from_array_round_trip(self):
+        mat = torch.eye(4)
+        mat[:3, 3] = torch.tensor([1.0, 2.0, 3.0])
+        rig = Rigid3Array.from_array(mat)
+        recovered = rig.to_tensor()
+        assert torch.allclose(recovered, mat)
+
+    def test_from_tensor_4x4_round_trip(self):
```
jnwei: since `.from_tensor_4x4` and `.to_tensor_4x4` is an alias of `.from_tensor` and `.to_tensor`, I think this test is redundant and could be removed

openfold3/tests/utils/geometry/helpers.py
```diff
@@ -0,0 +1,73 @@
...
+            [
+                [c, -s, 0.0],
+                [s, c, 0.0],
+                [0.0, 0.0, 1.0],
+            ]
+        )
+    )
+
+
+def rigid(rot: Rot3Array, tx: float, ty: float, tz: float) -> Rigid3Array:
```
jnwei: nit - It would be a lot more readable if tx, ty, and tz were passed in as a tuple, rather than as 3 separate floats. 

In the helper function, these values get converted into a Vector3D anyway.

openfold3/tests/utils/geometry/test_rotation_matrix.py
```diff
@@ -0,0 +1,338 @@
...
+        assert torch.allclose((eye @ r).to_tensor(), r.to_tensor(), atol=1e-6)
+        assert torch.allclose((r @ eye).to_tensor(), r.to_tensor(), atol=1e-6)
+
+
+# ===================================================================
+# Scalar multiply
+# ===================================================================
+
+
+class TestScalarMultiply:
```
jnwei, line 255: nit: Should probably include a test case for multiply by a scalar !=  1 or 0

openfold3/tests/utils/geometry/test_vector.py
```diff
@@ -0,0 +1,373 @@
...
+    dot,
+    euclidean_distance,
+    norm,
+    normalized,
+    square_euclidean_distance,
+)
+
+from .helpers import v as _v
+from .helpers import vb as _vb
+
```
jnwei: Prefer direct imports rather than relative imports

jnwei (approved):
LGTM with some fixes requested for imports and a readability nit

openfold3/tests/utils/geometry/test_vector.py
jandom: Agreed

openfold3/tests/utils/geometry/test_rotation_matrix.py
jandom: Agreed

openfold3/tests/utils/geometry/helpers.py
jandom: True true, done

openfold3/tests/utils/geometry/test_rigid_matrix_vector.py
jandom: Yup, good catch, thx

openfold3/core/utils/geometry/vector.py
jandom: ah, this is something I still need to add

openfold3/tests/utils/geometry/test_rigid_matrix_vector.py
jandom: switched, i also added a ruff rule to enforce this repo-wide, so a few other files were touched

PR #185, opened 2026-04-16
Fix template chain remap crash on unmappable chains

giulioisac:
## Summary

- When `remap_colabfold_template_chain_ids` encounters a template whose author chain ID is not in the RCSB mapping, it now logs a warning and drops that template row instead of raising a `RuntimeError` that kills the entire prediction.
- Updated existing test from expecting `RuntimeError` to verifying the template is skipped.
- Added test for mixed valid + invalid templates (only bad rows dropped, valid ones kept).

## Why #171 didn't fix this

PR #171 addressed the case where the RCSB API returns **no chain mapping at all** (e.g. obsolete or removed PDB entries, where `author_to_label` is empty). That fix falls back to using the author chain ID as the label chain ID.

However, it explicitly kept the `raise RuntimeError` for the second failure mode: when the RCSB mapping **exists but doesn't contain the specific chain** that ColabFold assigned. This happens when ColabFold's `pdb70` hits reference a chain ID that isn't present in the RCSB author-to-label mapping. For example:

- `Author chain L not found in 2a9n. Available author chains: ['A', 'B']`
- `Author chain E not found in 3nn8. Available author chains: ['A', 'B', 'C', 'D']`
- `Author chain E not found in 4nkd. Available author chains: ['A', 'B', 'C', 'D']`

In these cases falling back to the author chain ID is not safe because the mapping is available but simply doesn't include that chain — the template hit is invalid and should be dropped.

This PR handles that second mode by dropping the unmappable template row and continuing with the remaining valid templates, rather than crashing the entire prediction.

## Test plan

- [x] Unit tests pass (valid remap, unknown chain skipped, mixed valid+invalid)

### DIFF
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -686,10 +686,11 @@ def remap_colabfold_template_chain_ids(
         for entry_id, l2a in label_to_author_maps.items()
     }
 
-    # Remap chain IDs
+    # Remap chain IDs, dropping rows whose author chain ID cannot be resolved
     for top_n in per_rep.values():
         remapped_ids = []
-        for template_id in top_n[1]:
+        rows_to_drop: list[int] = []
+        for idx, template_id in zip(top_n.index, top_n[1]):
             entry_id, author_chain_id = template_id.split("_")
 
             author_to_label = author_to_label_maps.get(entry_id, {})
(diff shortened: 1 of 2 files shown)


jandom (approved):
@giulioisac excellent first contribution, I like it! Especially because it's not a just a fix but also progressively tighter/nuanced tests.

PR #191, opened 2026-04-22
Change inference integration test to use unittest instead of pytest

### LINKED ISSUE #187: [REQUEST] `pytest` should be part of `pip install openfold3`
Dear OpenFold Team,

To perform integration testing when going through `setup_openfold` the package `pytest` is needed.
It will be helpful if this is included during `pip install openfold3`.

jnwei:
**Summary**
Change the main inference integration test called by `setup_openfold` to use unittest instead of pytest. This way, users do not need to install pytest (a large dependency, 2.1G) to run `setup_openfold`

**Changes**
- Change test_full_inference.py to use unittest
- Change call of inference test in `setup_openfold` to use unittest.

**Related Issues**
Closes #187 

**Testing**
- [x] `test_full_inference.py` passes

**Other Notes**
In general, OpenFold3 has been making increasing use of pytest, making use of many fixtures and other features of pytest. Changing the inference integration test to use unittest instead goes against this flow, but I think it is a better alternative compared to adding `pytest` to the core dependencies. We can revisit this decision later if needed.

### DIFF
--- a/openfold3/setup_openfold.py
+++ b/openfold3/setup_openfold.py
@@ -18,7 +18,6 @@ Setup script for OpenFold3 parameters.
 Downloads model parameters and runs verification tests.
 """
 
-import importlib.util
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


openfold3/setup_openfold.py
```diff
@@ -207,30 +206,23 @@ def run_integration_tests() -> None:
...
+    try:
+        root_logger.setLevel(logging.WARNING)
+        program = unittest.main(
+            module="openfold3.tests.test_inference_full",
+            exit=False,
+            verbosity=2,
+        )
+    finally:
+        root_logger.setLevel(original_level)
+    exit_code = 0 if program.result.wasSuccessful() else 1
```
jandom, line 225: Oh dear

jandom (approved):
Let's just bite the bullet on this

openfold3/setup_openfold.py
jnwei: Let me know if you have other suggestions for handling the logging. I wanted to sequester logging info from the main inference run (MSA, some rdkit conformer generation messages) that were showing up upon running `setup_openfold`. But I agree this is messy

jandom: No, i've got nothing better here, unfortunately

PR #196, opened 2026-04-27
Make template preprocessing timeout configurable via CLI

### LINKED ISSUE #164: Add an option to change the timeout length of Function _preprocess_templates_for_query ?
**Is your feature request related to a problem? Please describe.**
The localcolab fold step sometimes takes longer than 60seconds ; resulting in a timeout - 
func_timeout.exceptions.FunctionTimedOut: Function _preprocess_templates_for_query ... (kwargs={}) timed out after 60.000000 seconds."

**Describe the solution you'd like**
Either add an option to change the timeout or change the default to 300 seconds?

**Describe alternatives you've considered**
Currently we run it in two times, keeping the intermediate files

CesarPuentes:
**Resolves #164**

### Summary
Replaces the hardcoded 60-second limit in template preprocessing with a configurable `--preprocess-timeout` CLI parameter. This is necessary to prevent crashes when processing exceptionally large protein sequences.

### Changes Made

**1. `openfold3/core/data/pipelines/preprocessing/template.py`**
- Added `preprocess_timeout` (default: 60) to the `TemplatePreprocessorSettings` model.
- Updated `func_timeout` to evaluate this dynamic variable instead of the static `60`.
- Added an `except FunctionTimedOut:` block catch timeouts, skip the offending template chain, and print a helpful warning suggesting the user increase `--preprocess-timeout`.

**2. `openfold3/run_openfold.py`**
- Exposed `--preprocess-timeout` as a Click option in the `predict` CLI function.
- Piped the argument directly into `expt_config.template_preprocessor_settings.preprocess_timeout`.

### Testing Performed
- **Ubiquitin (`query_ubiquitin.json`)**: Verified that sequences for short proteins complete with the standard and custom timeout, and that a forced failure can be achieved with a `--preprocess-timeout` of 1.

```bash
pixi run -e openfold3-cuda13 python openfold3/run_openfold.py predict \
    --query_json examples/example_inference_inputs/query_ubiquitin.json \
    --use_templates True
```
### Titin (Q8WZ42.fasta)
Attempted to replicate the issue where exceptionally long sequences time out during template processing. Verified that extending the limit with --preprocess-timeout 300 correctly intercepts the limit and prevents the crash.

```bash
pixi run -e openfold3-cuda13 python openfold3/run_openfold.py predict \
    --query-json ../query_Q8WZ42.json \
    --use_templates True \
    --preprocess-timeout 300
```

### DIFF
--- a/docs/source/configuration_reference.md
+++ b/docs/source/configuration_reference.md
@@ -268,6 +268,7 @@ Configures template structure preprocessing and filtering.
 - `create_logs` *(bool)*: Create preprocessing logs (default: `false`)
 - `n_processes` *(int)*: Number of preprocessing processes (default: `1`)
 - `chunksize` *(int)*: Tasks per worker in multiprocessing (default: `1`)
+- `preprocess_timeout` *(int)*: Maximum preprocessing time in seconds (default: `60`)
 - `structure_directory` *(Path | None)*: Directory for template structures (default: `null`)
 - `structure_file_format` *(str)*: File format of structures - `cif` or `pdb` (default: `cif`)
 - `output_directory` *(Path | None)*: Output directory for templates (default: `null`)
--- a/examples/reference_full_config/full_config.yml
+++ b/examples/reference_full_config/full_config.yml
@@ -154,6 +154,7 @@ template_preprocessor_settings:
   create_logs: false
   n_processes: 1
   chunksize: 1
+  preprocess_timeout: 60
   structure_directory: <tmp-dir>/of3-of-<user>/template_data/template_structures
   structure_file_format: cif
   output_directory: <tmp-dir>/of3-of-<user>/template_data
--- a/openfold3/core/data/pipelines/preprocessing/template.py
+++ b/openfold3/core/data/pipelines/preprocessing/template.py
@@ -33,6 +33,7 @@ from biotite.structure import AtomArray
 from biotite.structure.io import pdbx
 from biotite.structure.io.pdbx import CIFFile
 from func_timeout import func_timeout
+from func_timeout.exceptions import FunctionTimedOut
 from pydantic import (
     BaseModel,
     BeforeValidator,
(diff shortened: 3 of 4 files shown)


jnwei (changes requested):
Thank you very much for this contribution, @CesarPuentes ! Thank you also for providing information with the explicit test cases you used, that is very helpful.

A few small requests, mostly around organization and documentation
 
- Could you remove the command line option? In general, we prefer to limit the number of arguments available at the command line, restricting the options to only those arguments that are likely to be changed frequently (such as number of diffusion samples). Unless we expect this parameter to be changed frequently, I think it is better to leave the parameter accessible through just the runner.yaml

- Would you mind updating the documentation for the new configuration? The relevant files can be found [here](https://github.com/aqlaboratory/openfold-3/blob/main/docs/source/configuration_reference.md?plain=1#L249), for read the docs page, and [here](https://github.com/aqlaboratory/openfold-3/blob/main/examples/reference_full_config/full_config.yml#L140), for the full configuration reference.

- As a bonus, it would be great to have your test examples as a unit test. Unfortunately, I see we have yet to create tests for the TemplatePreprocessor, so perhaps this is best left for a different PR.

CesarPuentes:
Hi @jnwei,

Glad to hear you find it useful! I've removed the option from the CLI and made additions to the documentation, as requested.

For future testing reference, I validated the timeout logic by running a prediction on the large Titin protein (Q8WZ42, ~34k residues) with both 60-second and 300-second timeouts using the following command:

bash
```
pixi run -e openfold3-cuda13 python openfold3/run_openfold.py predict --query-json ./Q8WZ42.json --runner-yaml test_runner.yml --output-dir ./results
```

Where Q8WZ42.json was structured as follows:

`
json
{
  "queries": {
    "Q8WZ42": {
      "chains": [
        {
          "molecule_type": "PROTEIN",
          "chain_ids": [
            "A"
          ],
          "sequence": "Insert Sequence here"
        }
      ]
    }
  }
}
`

- At 60s (default): It correctly triggers the "Template preprocessing TIMED OUT" exception, gracefully skips the templates for that chain, and continues the prediction pipeline without crashing.
- At 300s (custom YAML): It successfully respects the configured override, waiting the specified duration before timing out.

I also tested it with the much smaller ubiquitin protein on the default times, and it successfully completed the template preprocessing without triggering a timeout, as expected.

Let me know if you need anything else!

Kind regards,

jnwei (approved):
LGTM! Thank you @CesarPuentes

jandom:
Wonderful contribution – thank you @CesarPuentes

PR #197, opened 2026-04-28
refactor: improved conformer generation

jandom:
**Summary**

Refactor conformer generation into a multistrategy chain, add `small_ring_torsions` as a strategy that handles macrocyclic ligands (e.g. heme), and back it with a real test suite.

**Changes**

- `primitives/structure/conformer.py`: introduce `ConformerStrategy` + ordered `CONFORMER_STRATEGIES` registry 
-  the strategies are: `default` → `small_ring_torsions` → `random_init`
- `pipelines/sample_processing/conformer.py`: collapse the per-strategy if/elif in `get_processed_reference_conformer` into one `multistrategy_compute_conformer(start_from=...)` call.
- `primitives/structure/query.py`: migrate caller to `ConformerResult`.

- New tests for chain dispatch, `start_from` slicing, snapshot regression on `_compute_conformer`, flag propagation, `get_processed_reference_conformer`, and a parametrized real-RDKit run over PDB CCDs HEM / CLA / CHL / BCL with per-ligand `default_succeeds`.

**Related Issues**
N/A

**Testing**
`pytest openfold3/tests/core/data/primitives/structure/test_conformer.py` — 36 passed in 15.83s.

**Other Notes**
- `conformer_gen_strategy` strings are persisted in the preprocessing cache; a test pins the literals.
- `multistrategy_compute_conformer` is keyword-only past `mol`; `timeout_standard` / `timeout_rand_init` are replaced by `timeouts={}`.
- doesn't feature a CCD treadmill, where we could evaluate all the CCD molecules and how they do under conf-gen in-between rdkit versions etc


openfold3/core/data/pipelines/sample_processing/conformer.py
jnwei, line 87: It irks me that the list of confgen strategies here does not match CONFORMER_STRATEGIES. But I understand that `use_fallback` is a special case and shouldn't be part of CONFORMER stragies... also we are using `use_fallback` in our training dataset preprocessing pipeline, so we cannot change this value to be flag instead, correct?

openfold3/core/data/primitives/structure/conformer.py
```diff
@@ -37,9 +38,44 @@ class ConformerGenerationError(ValueError):
...
+    The `name` is persisted as `conformer_gen_strategy` in the preprocessing cache
+    (see `PreprocessingReferenceMoleculeData`); do not rename existing entries.
+    """
+
+    name: str
+    kwargs: dict = field(default_factory=dict)
+    default_timeout: float | None = None
+
+
+CONFORMER_STRATEGIES: tuple[ConformerStrategy, ...] = (
```
jnwei, line 54: nice tuple of frozen dataclasses

openfold3/core/data/primitives/structure/conformer.py
```diff
@@ -115,73 +155,62 @@ def compute_conformer(
...
-            timeout=timeout_standard,
-        )
-    except (ConformerGenerationError, FunctionTimedOut) as e:
-        logger.warning(
-            f"Exception when trying standard conformer generation for {smiles}: {e}, "
-            + "trying random initialization"
-        )
+    # Validate `start_from` and locate it in the chain.
+    _get_strategy(start_from)
+    chain_start = next(
```
jnwei: small nit: Can we please call this variable "iter_start" or something else? chain_start to me implies that it is related to Chain data and is a bit confusing

openfold3/tests/core/data/pipelines/sample_processing/test_processed_reference_conformer.py
```diff
@@ -0,0 +1,313 @@
...
+    used_mask = [a.GetBoolProp("annot_used_atom_mask") for a in out.mol.GetAtoms()]
+    assert used_mask == [False] * acetic_acid_mol.GetNumAtoms()
+
+
+def test_unit_atom_reorder_when_mol_order_differs_from_atom_array(
+    acetic_acid_atom_array,
+):
+    # Build a mol whose annot_atom_name order is the REVERSE of the atom_array's
+    # order. The function must reorder mol atoms via `get_name_match_argsort` so
+    # that the output mol's atoms align with the atom_array.
```
jnwei, line 248: I want to check my understanding of the logic for this test and the intended behavior of these functions:

Assumptions:
- We have a mol whose atoms are in a different order relative to the fallback conformer reference (the given example here is with ethanol)

Aims:
- We want to use the fallback conformer `use_fallback`
- But - in order to use the fallback, we have to reassign the atoms to the atoms in the copy of the fallback structure that gets returned as the output of `get_preprocessor_conformer`

If this is all correct:
- What happens during training to this molecule? Do we ever need the original order of input molecule with the reversed atom names?
- Maybe a more direct question - when we predict the atom positions, will it still have the atoms assigned in the reversed order?

openfold3/tests/core/data/pipelines/sample_processing/test_processed_reference_conformer.py
```diff
@@ -0,0 +1,313 @@
...
+
+    out_coords = out.mol.GetConformer(0).GetPositions()
+    assert np.isfinite(out_coords).all()
+
+    if coords_should_match_fallback:
+        np.testing.assert_allclose(out_coords, fallback_coords, atol=1e-6)
+    else:
+        assert not np.allclose(out_coords, fallback_coords)
+
+    # annot_atom_name preserved through reorder; permutations are real.
```
jnwei, line 309: did we expect the atom names to be reordered in this test? Neither the inputs nor any of the conformer generation functions should perform atom reordering. Maybe this is just a safeguard and I'm still thinking of the previous reordering test.

openfold3/tests/core/data/primitives/structure/test_conformer.py
```diff
@@ -0,0 +1,437 @@
...
+# molecule with conf_id=0". Replaced with a real (mol, 0) tuple inside the test, since
+# parametrize evaluates at collection time before the `ethanol` fixture exists.
+_SUCCESS = object()
+
+
+@pytest.mark.parametrize(
+    ("outcomes", "expected_strategy"),
+    [
+        pytest.param(
+            [_SUCCESS],
```
jnwei: The correctness of these outcomes list depends on the order in CONFORMER_STRATEGIES in core.data.primitives.structures.conformer. In other words, if the order of CONFORMER_STRATEGIES changes, or we add a new strategy, then these may break.

Could we add a reference to CONFORMER_STRATEGIES here so that the future developer understands the connection? Ideally we'd incorporate it directly into the test, but we can also settle for a comment for now.

openfold3/tests/core/data/primitives/structure/test_conformer.py
```diff
@@ -0,0 +1,437 @@
...
+    result.mol.GetConformer(result.conf_id)  # would raise if id is invalid
+
+
+# ----------------------------------------------------------------------------
+# compute_conformer tests
+# ----------------------------------------------------------------------------
+
+# A fixed seed for the random-coord initializer so snapshots are reproducible
+# regardless of where this test sits in the seeded_rng-derived global RNG state.
+_FIXED_RDKIT_SEED = 424242
```
jnwei, line 140: I wonder if this seed should be defined somewhere else more globally, like in one of the conftest.py files... if we introduce future snapshot tests involving rdkit, it could be confusing to have this global seed defined in a single test module.

openfold3/tests/core/data/primitives/structure/test_conformer.py
```diff
@@ -0,0 +1,437 @@
...
+    else:
+        with pytest.raises(_DEFAULT_FAILURE_EXCEPTIONS):
+            _compute_conformer(mol, use_small_ring_torsions=False, timeout=5.0)
+
+
+@pytest.mark.parametrize(("ccd_code", "smiles", "default_succeeds"), _PDB_TETRAPYRROLES)
+def test_pdb_tetrapyrrole_small_ring_torsions_succeeds(
+    ccd_code, smiles, default_succeeds
+):
+    del default_succeeds  # claim under test is universal across the parametrize set
```
jnwei: ~It might be cleaner to exclude `default_succeeds` from this test argument if we're not using it at all?~

OK, I see this is coming from _PDB_TETRAPYRROLES. Then instead, a small nit, you might be able to skip the default_succeeds input with

```
@pytest.mark.parametrize(("ccd_code", "smiles", "default_succeeds"), _PDB_TETRAPYRROLES)
def test_pdb_tetrapyrrole_small_ring_torsions_succeeds(
    ccd_code, smiles, _
):
   mol = Chem.MolFromSmiles(smiles)
   ...
```

openfold3/tests/core/data/primitives/structure/test_conformer.py
```diff
@@ -0,0 +1,437 @@
...
+# ----------------------------------------------------------------------------
+
+_DEFAULT_FAILURE_EXCEPTIONS = (ConformerGenerationError, FunctionTimedOut)
+
+_PDB_TETRAPYRROLES = [
+    pytest.param(
+        "HEM",
+        "Cc1c2n3c(c1CCC(=O)O)C=C4C(=C(C5=[N]4[Fe]36[N]7=C(C=C8N6C(=C5)C(=C8C)"
+        "C=C)C(=C(C7=C2)C)C=C)C)CCC(=O)O",
+        False,
```
jnwei: We should provide a definition for the contents of the param dicts. Something like "CCD_code, smiles, whether the default succeeds"

openfold3/tests/conftest.py
jnwei, line 1: these changes should have been rebased from the pixi-rocm PR, right?

jnwei (changes requested):
Overall, this is really great. It is awesome to have such extensive tests for conformer generation. The addition of the small_ring_torsion conformer generation should hopefully unblock HEM and many other conformer stumbling blocks.

I have a few questions regarding reordering the atoms in some of the tests, but otherwise, LGTM

openfold3/core/data/pipelines/sample_processing/conformer.py
jandom: Yeah, it's not ideal in semantics – one of the strategies... is not actually a strategy :D 
At inference we fallback to nan coordinates, in training we fallback but ideal CCD coords.

openfold3/core/data/primitives/structure/conformer.py
jandom: you know my love for lightweight and simple data structures is strong!

openfold3/core/data/primitives/structure/conformer.py
jandom: that's a very sensible suggestion, i agree this is confusing and also bugged me

openfold3/tests/conftest.py
jandom: correct: the pixi-rocm PR introduced platform-dependent snapshots (that are needed for the kernels) and made that the default. But for the data pipeline we're unlikely to need per-platform snapshots, so this is separating these platform-dependent (minor) and platform-independent (major) snapshots.

openfold3/tests/core/data/primitives/structure/test_conformer.py
jandom: You mean as a comment on top, or as a kwargs to? I'm not sure pytest.param provides this. We certainly do label the individual test cases with these names, eg 


```
@pytest.mark.parametrize(("ccd_code", "smiles", "default_succeeds"), _PDB_TETRAPYRROLES)
```

openfold3/tests/core/data/primitives/structure/test_conformer.py
jandom: Agreed

jandom: Actually, this is not permitted by pytest, it requires a named argument match!

openfold3/tests/core/data/primitives/structure/test_conformer.py
jandom: My sense is that yes we should do it once we have more than one consumer for this. Currently it's just this test: would it go into a root conftest, or some intermediate level conftest (which?)

openfold3/tests/core/data/pipelines/sample_processing/test_processed_reference_conformer.py
jandom: Arguably this test covers changes unrelated to this PR, let me think a bit about this.

openfold3/tests/core/data/primitives/structure/test_conformer.py
jandom: I tried a few things here, your concern is completely correct: it's a brittle test, so I tried to parametrize based on the strategies – so that the cases change a strategies are added/removed/modified. 

But I basically concluded that it's is a useless test (testing implementation not behavior). So removed it after some tweaking :P

openfold3/tests/core/data/primitives/structure/test_conformer.py
jnwei: darn, ok

openfold3/tests/core/data/primitives/structure/test_conformer.py
jnwei: Perhaps just as an inline command for each of these test case blocks? Mostly to make it obvious what False is referring to

openfold3/tests/core/data/pipelines/sample_processing/test_processed_reference_conformer.py
jandom: This is probably the biggest comment in the review, so it'll take a while to unpack. I don't think my code changes the existing behavior here, it just that this test prompted some "what's going on here exactly". 

This is my "understanding" of what's happening
- Yes, the input mol can have atoms in a different order than the GT atom_array (the reversed-names case in the test).
- Yes, Chem.RenumberAtoms(mol, reorder_index) at (openfold3/core/data/pipelines/sample_processing/conformer.py#L143) realigns the mol to the atom_array's atom_name_unique order. This happens before the use_fallback / regenerate branch, so both paths produce a mol whose atom indexing matches the GT.

Q: Do we ever need the original order of the input mol with the reversed atom names?

I don't think so, the code seems internally consistent to re-align and go with that forward (aligned/reordered to GT). 

Q: When we predict atom positions, will they still be assigned in the reversed order?

No. Predicted coordinates are written to the input atom_array, not to the reference mol.

Now, it could be worthwhile to add tests that **demonstrate** that this is the case :D

jnwei: Thank you for the explanation. Yes, my interest was piqued by the example and was wondering about the implications, but it isn't specifically about this test case.

It would indeed be helpful to have additional tests to test the fate of atoms when they are permuted / renumbered throughout the pipeline, but that is a job for another PR.

openfold3/tests/core/data/pipelines/sample_processing/test_processed_reference_conformer.py
jnwei: It's fine to leave it as a safeguard if you think it is helpful. Basically, this checks that the the permutation outputs maintain the same atom order, in case they should ever change.

jnwei (approved):

openfold3/tests/core/data/primitives/structure/test_conformer.py
jandom: Oh I see – sorry I didn't get it earlier. Well, now each case is just a named tuple (dataclass is awkward to unpack)

```
        *_TetrapyrroleCase(
            ccd_code="HEM",
            smiles=(
                "Cc1c2n3c(c1CCC(=O)O)C=C4C(=C(C5=[N]4[Fe]36[N]7=C(C=C8N6C(=C5)"
                "C(=C8C)C=C)C(=C(C7=C2)C)C=C)C)CCC(=O)O"
            ),
            default_succeeds=False,
        ),
```

So it's clearer what true and false are referring to, I think

jandom:
Thank you for the review @jnwei this was helpful – i'm not too happy about these permutation tests, it feels like a drive-by test that A) doesn't properly exercise the code that was added here B) doesn't properly exercise the permutation code either :D

PR #204, opened 2026-04-30
Revert CuEq bias expansion when chunking with BS=1

christinaflo:
Adding in @borisfom's patch from #181. The other PR fixed input shapes for cueq with chunking and BS>1. This isn't needed for BS=1, reverting to the old behavior to avoid expanding the bias tensor as this will blow up mem usage.


jnwei (approved):
LGTM thank you @christinaflo and @borisfom

PR #205, opened 2026-04-30
docs: update install instructions

etowahadams:
I noticed that we don't explicitly say to clone the repo before running the pixi commands. So I updated the docs to reflect this.


jnwei (approved):
LGTM, thanks Etowah!

PR #207, opened 2026-05-01
Apply power-of-two chunking consistently

### LINKED ISSUE #203: [BUG/QUESTION] Chunk size incremented by 4 off largest power of 2

### LINKED ISSUE #206: Restoring chunking for batch_size > 1 with optimized kernels
The AuxiliaryHeadsAllAtom module uses a pairformer embedding with batch equal to diffusion_samples. This hits a special case in PairformerEmbedding where chunking is turned off if any of the optimized triangle attention kernels are in use and batch size is > 1:

https://github.com/aqlaboratory/openfold-3/blob/77d57c95fe60a47d8b0b8d36bcc216a69ea53472/openfold3/core/model/heads/prediction_heads.py#L212-L222

To the extent that chunking is necessary to avoid OOMs, this seems not great. It results in inputs to attention with shape `(diffusion_samples, seq_len, seq_len, heads, dim)`. If we're not chunking that call, then I think it doesn't make a whole lot of sense to chunk any of the others either, which all have smaller inputs AFAICT. I experimented with setting `torch.cuda.set_per_process_memory_fraction` to simulate a lower memory GPU and confirmed that this call causes an OOM.

The comment about disabling documents some unspecified issues in the deepspeed and cuequivariance kernels. I experimented with turning chunking back on and I think I've identified the issue. The batch dimensions all get combined into a single dimension and expanded to match. So the pair bias gets expanded from `(diffusion_samples, 1, seq_len, heads, dim)` to `(diffusion_samples, seq_len, seq_len, heads, dim)` and then `(diffusion_samples*seq_len, seq_len, heads, dim)` (and later unsqueezed for an initial batch dim).

https://github.com/aqlaboratory/openfold-3/blob/6ac5a725e057d5a7077d3142b0ec38aa2573631d/openfold3/core/utils/chunk_utils.py#L255-L262

But the optimized kernels expect pair bias to be 1 in the second dimension and implicitly broadcast it.

- Cuequivariance: https://github.com/NVIDIA/cuEquivariance/blob/5196d219b3975428f707105e7fae1827706e0ebe/cuequivariance_torch/cuequivariance_torch/primitives/triangle.py#L58-L61
- Deepspeed: https://github.com/deepspeedai/DeepSpeed/blob/8a77f381a4d8f3bf65747ac571b94324eaa4a29a/deepspeed/ops/deepspeed4science/evoformer_attn.py#L98
- Triton (implicit in the fact that it doesn't even look at pair_bias.stride(1). We should add an assert): https://github.com/aqlaboratory/openfold-3/blob/6ac5a725e057d5a7077d3142b0ec38aa2573631d/openfold3/core/kernels/triton/evoformer.py#L962C40-L965

I think the easiest fix is to turn on apply_per_sample when using the optimized kernels. This is easy to add here:

https://github.com/aqlaboratory/openfold-3/blob/6ac5a725e057d5a7077d3142b0ec38aa2573631d/openfold3/core/model/heads/head_modules.py#L187-L192

It's a bit annoying to not be able to chunk freely over multiple batch dimensions and potentially get better chunking, but to the extent that chunking is needed at all, I don't think it makes sense to leave it off for this one module.

GMNGeoffrey:
**Summary**
Currently chunking of the `AuxiliaryHeadsAllAtom` pairformer stack is disabled if using optimized attention kernels to avoid a bug with multi-sample chunking. This is the largest-memory attention call, so it doesn't make a whole lot of sense to disable it here. Instead, we split by sample when using optimized kernels. Additionally makes chunk sizes consistently a power of two rather than 4 being added to the largest chunk size (it was suggested to combine these changes. Happy to split them up if desired).

**Changes**
- Avoid weird addition of 4 to power-of-two chunk sizes. This was added in https://github.com/aqlaboratory/openfold-3/commit/a9a12890d without explanation. We can hypothesize that it was related to adding 4 to an input dimension in trace_utils.py (trying to get a test case to fit in one chunk?), but that file was long ago deleted. This just looks like a bug and makes us hit unhappy paths all over the place. Fixes https://github.com/aqlaboratory/openfold-3/issues/203
- Enable chunking for `AuxiliaryHeadsAllAtom` pairformer embedding when using optimized kernels. Without chunking, this is the first call to cause OOMs because its `diffusion_samples*sequence_length` batches. Chunking gets turned off in prediction_heads.py due to batch size > 1 and use of optimized kernels because cross-sample chunking requires expanding out pair bias and they all require it to have size 1 in the second dimension with implicit broadcasting. So we turn on `apply_per_sample` when optimized kernels are in use. This splits the > 1 batch dimension, which avoids this problematic path and then we can do normal chunking for the rest if it's still too large. We could do something more elaborate (see suggestions in linked issue), but this is an improvement for now. Fixes https://github.com/aqlaboratory/openfold-3/issues/206
- Assert in Triton triangle attention kernel that the second dim of pair bias has size 1. This was implicitly assumed with `pair_bias.stride(1)` not even passed into the function.
- Update tests for all optimized kernels in test_kernels.py to avoid batch_size > 1 + chunking
- Document specific issue with multi-sample chunking and optimized kernels

**Related Issues**
- https://github.com/aqlaboratory/openfold-3/issues/203
- https://github.com/aqlaboratory/openfold-3/issues/206

**Testing**
- Confirmed that when limiting torch GPU memory usage to 12GB, the multimer example OOMed prior to this change and runs to completion after.
- Confirmed that attention kernel calling pattern moves from batches of 516 to 512.
- I didn't find a good way to unit test this that was actually a useful test (e.g. I can assert that chunk_size is now 512 instead of 516, but I don't think that test carries its weight).

**Other Notes**
I'm happy to make some of the more elaborate changes discussed in https://github.com/aqlaboratory/openfold-3/issues/206 either in this PR or a separate one, but wanted to share this relatively simple fix.

### DIFF
--- a/openfold3/core/kernels/triton/evoformer.py
+++ b/openfold3/core/kernels/triton/evoformer.py
@@ -905,8 +905,20 @@ if _TRITON_AVAILABLE:
 
             BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM = Q.shape
 
-            assert res_mask.shape == (BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
-            assert pair_bias.shape == (BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
+            assert res_mask.shape == (
+                BATCH_SIZE,
+                N_SEQ,
+                1,
+                1,
+                SEQ_LEN,
+            ), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
+            assert pair_bias.shape == (
+                BATCH_SIZE,
+                1,
+                HEAD,
+                SEQ_LEN,
+                SEQ_LEN,
+            ), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
 
             softmax_scale = DIM**-0.5
             BLOCK_DIM = max(triton.next_power_of_2(DIM), 32)
--- a/openfold3/core/kernels/triton/evoformer.py
+++ b/openfold3/core/kernels/triton/evoformer.py
@@ -904,6 +904,10 @@ if _TRITON_AVAILABLE:
             ).contiguous()  # (BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM)
 
             BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM = Q.shape
+
+            assert res_mask.shape == (BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
+            assert pair_bias.shape == (BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
+
             softmax_scale = DIM**-0.5
             BLOCK_DIM = max(triton.next_power_of_2(DIM), 32)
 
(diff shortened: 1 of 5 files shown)


GMNGeoffrey:
I think removing the `+=4` on the largest chunk size is pretty uncontroversial. Should I split that out into a separate PR so it can be merged?

GMNGeoffrey:
I think it's ok to close this. Like I said, I think the important things are covered in #226 and #228. We can explore getting rid of `apply_per_sample` entirely in #213, but I think you're right that for now in practice it kicks in before we get to a sequence length where it's going to matter anyways

PR #208, opened 2026-05-01
Add .venv to gitignore

GMNGeoffrey:
This is a common path for virtual environments and actually documented in installation.md


GMNGeoffrey:
Whoops, looks like I just needed to pull main 😅

PR #213, opened 2026-05-05
Avoid flattening and expanding broadcast dimensions when chunking

### LINKED ISSUE #203: [BUG/QUESTION] Chunk size incremented by 4 off largest power of 2

### LINKED ISSUE #206: Restoring chunking for batch_size > 1 with optimized kernels

GMNGeoffrey:
**Summary**
This is an alternative, more complicated fix for #206 instead of #207. To avoid the expansion of an implicit broadcast dimension (e.g. in pair bias), if one exists it chunks across each batch dimension discretely. If the innermost batch dimension (usually N_seq) is larger than chunk_size then this is equivalent to splitting per sample. If it's smaller though, then we might fit multiple samples in a single chunk.

So as an example when we have diffusion_samples 5 and sequence length 76 in the pairformer embedding we have batch dims [5, 76] for QKV but for the pair bias it's [5, 1]. Rather than flattening to [380], we chunk each dimension separately. If chunk size were 128 then we'd split into 5 chunks of [1, 76]; if it were 64, then we'd split into 5 chunks of [1, 64] and 5 chunks of [1, 12]; if chunk size were 256, then we'd split into chunks of [3, 76] and [2, 76], etc.

Doing this in the `chunk_layer` function allows us to be chunk-size-aware, which we can't be at higher levels when chunk size has potentially not yet been determined due to chunk size tuning. I think this could completely replace `apply_per_sample` and `per_sample_token_cutoff` as well and is more robust.

**Changes**
- Add new function for dimension-aware chunking employed when inputs would be broadcast in their batch dimensions 
- Remove workarounds that avoid chunking when using optimized kernels and batch size > 1
- Allow kernel tests to use batch_size > 1

**Related Issues**
- Fixes https://github.com/aqlaboratory/openfold-3/issues/206
- Fixes https://github.com/aqlaboratory/openfold-3/issues/203

**Testing**
- Unit tests for chunking logic
- Integration tests for kernels now work with batch==2

**Other Notes**
- I have not done performance testing on this yet, which I think we may want to do before merging.
- This is based on #207 but reverts some of the changes it makes. If we want to go this direction then we could close #207 and use this instead.
- This was partially authored by Claude

### DIFF
--- a/openfold3/core/kernels/triton/evoformer.py
+++ b/openfold3/core/kernels/triton/evoformer.py
@@ -905,8 +905,20 @@ if _TRITON_AVAILABLE:
 
             BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM = Q.shape
 
-            assert res_mask.shape == (BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
-            assert pair_bias.shape == (BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
+            assert res_mask.shape == (
+                BATCH_SIZE,
+                N_SEQ,
+                1,
+                1,
+                SEQ_LEN,
+            ), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
+            assert pair_bias.shape == (
+                BATCH_SIZE,
+                1,
+                HEAD,
+                SEQ_LEN,
+                SEQ_LEN,
+            ), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
 
             softmax_scale = DIM**-0.5
             BLOCK_DIM = max(triton.next_power_of_2(DIM), 32)
--- a/openfold3/core/kernels/triton/evoformer.py
+++ b/openfold3/core/kernels/triton/evoformer.py
@@ -904,6 +904,10 @@ if _TRITON_AVAILABLE:
             ).contiguous()  # (BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM)
 
             BATCH_SIZE, N_SEQ, HEAD, SEQ_LEN, DIM = Q.shape
+
+            assert res_mask.shape == (BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN), f"{tuple(res_mask.shape)} != {(BATCH_SIZE, N_SEQ, 1, 1, SEQ_LEN)}"
+            assert pair_bias.shape == (BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN), f"{tuple(pair_bias.shape)} != {(BATCH_SIZE, 1, HEAD, SEQ_LEN, SEQ_LEN)}"
+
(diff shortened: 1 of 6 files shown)


christinaflo:
I am in favor of this kind of fix vs #207, I always wanted something like this to avoid the workarounds but never got to it lol, thank you for this! Would you be able to do the performance testing? I'm wondering how high it could go compared to apply_per_sample + offloading.

GMNGeoffrey:
Yeah I'm happy to do some performance testing :-) Would be helpful to merge my other chunking PRs so I can get this on a stable base.

This solution still does leave something to be desired because a chunk of 1024 when it includes the batch dimension is not the same as one where it's just the sequence count, as pair bias scales with the former but not the latter. But I guess that's ok because we're caching chunk size based on the argument shapes anyways.

jandom:
Sorry @GMNGeoffrey it looks like some of my changes gave conflicts here – hopefully won't be a nightmare to resolve

GMNGeoffrey:
> Sorry @GMNGeoffrey it looks like some of my changes gave conflicts here – hopefully won't be a nightmare to resolve

NBD. I think most of the merge conflicts are with my own PRs :-)

PR #216, opened 2026-05-06
Fix SMILES ligand seq IDs in CIF output

### LINKED ISSUE #214: [BUG] SMILES based ligands lack useful auth_seq_id/label_seq_id
**Describe the bug**
When doing structure prediction with a ligand based on a SMILES entry, the `label_seq_id`  and `auth_seq_id` entries of the `atom_site` table are listed as '.'

This makes working with the produced structures in downstream contexts more difficult, as other programs assume that this column will contain integers (e.g. Attempting to load the structure in AtomWorks results in an crash), and even if they don't crash, most programs which interact with protein structures use the chain letter and seq_id combination to work with and identify specific residues. Having the seq_id be unspecified results in difficulties in working with and specifying the ligand residue in downstream applications.

**Hoped for Behavior**

Instead of the '.' entry, ligands would be given a valid numeric seq_id. (i.e. if they're the only residue in the chain, a seq_id value of 1 would be reasonable.)

**To Reproduce**
Example input showing the issue with OpenFold3 version 0.4.0

```
{
"queries": {
"test":{
    "chains": [
        {
            "molecule_type": "protein",
            "sequence": "ACDEFGHIKLMNPQRSTVWY",
            "chain_ids": "A"
        },
        {
           "molecule_type": "ligand",
           "smiles":"NCCc1cc(O)c(O)cc1",
           "chain_ids": "X"
        }
    ]
}
}
}
```

ryanhulke:
**Summary**

Fix missing CIF sequence IDs for SMILES-based ligands (described in #214). Previously, SMILES ligands were written with `.` for `label_seq_id` and `auth_seq_id`, which can break downstream tools that expect numeric residue IDs.

**Changes**

- Fill missing ligand `label_seq_id` and `auth_seq_id` values from the AtomArray residue ID during CIF writing in `openfold3/core/data/primitives/structure/metadata.py`
- Added a test to `openfold3/tests/test_structure_from_query.py` for a protein plus SMILES ligand query matching the reported issue

**Related Issues**

- Fixes https://github.com/aqlaboratory/openfold-3/issues/214

**Testing**

- all tests in `openfold3/tests/test_structure_from_query.py` pass, including the new `test_smiles_ligand_cif_seq_ids_are_numeric` 

**Other Notes**

- My first contribution


jnwei (changes requested):
Thank you @ryanhulke for the thorough fix and the additional tests.

I have one comment: Would it make sense for the label_seq_id to remain "." while updating the author_seq_id to be numerical

Having Claude look over a few mmcif examples, for non-polymer ligands, the label_seq_id is usually kept as "." Some examples below

@roccomoretti @ljarosch do you have an opinion about whether we can keep the label_seq_id as "."?




Residue | Type | label_seq_id | auth_seq_id
-- | -- | -- | --
HEM (43 atoms, heme group) | non-polymer ligand | "." | numeric (142, 147)
EPE (HEPES buffer) | non-polymer ligand | "." | numeric (3022)
SO4 (sulfate) | non-polymer ligand | "." | numeric
HOH (water) | non-polymer ligand | "." | numeric
MSE (selenomethionine) | modified polymer residue | numeric | numeric
5IU (nucleotide analog) | part of RNA chain | numeric | numeric

ryanhulke:
@jnwei yeah that makes sense. Updated the fix so non-polymer ligands keep label_seq_id as "." while only filling missing auth_seq_id from the AtomArray residue ID. Also updated the test accordingly.

jnwei (approved):
LGTM thanks for the fix @ryanhulke !

roccomoretti:
It does look like only setting `auth_seq_id` is sufficient to make AtomWorks not crash, and it also works as expected with PyMol.  (And it does indeed look like from-wwPDB mmCIF structures leave the `label_seq_id` column as `.` for ligands.)

ryanhulke:
New commit specifically overwrites label_seq_id to "." in ligands/non-polymers, and I ran the ruff check. Should pass tests now.

jnwei:
Excellent, thank you @ryanhulke for the fixes and @roccomoretti for confirming the `label_seq_id` column values may be left as `.`

Will merge once CI tests pass

jnwei (approved):

jandom:
Great fix, thanks all!

PR #220, opened 2026-05-11
Monkeypatch trunc_normal_init_ to noop during inference

rizsp:
**Summary**

I noticed that the random initialization takes up half a minute on model load and tried to optimize it in #178 by moving it from the CPU to the GPU.

In this PR, I eliminate random initialization during inference because it is actually not necessary at all...the model weights just replace these randomly initialized weights anyway. I do this with a simple monkeypatch of `trunc_normal_init_` (since this is the only initialization function which popped out during profiling) to a noop inside the inference runner.

This saves half a minute during model load.

**Changes**

Inside `experiment_runner.py:InferenceExperimentRunner` I override the `lightning_module` method using a context manager which monkey patches `trunc_normal_init_` to a noop. 

**Related Issues**

This PR is related to #178. 

**Testing**

I added a simple test of the monkey patching in a separate test file since I didn't know which existing test would be most suitable. I also copied over the benchmark test from #178 to test for model load time regression.

**Other Notes**


openfold3/entry_points/experiment_runner.py
```diff
@@ -535,6 +536,16 @@ def callbacks(self):
...
 
 
+@contextlib.contextmanager
+def skip_random_init():
+    from unittest.mock import patch
+    def noop_init(*args, **kwargs):
+        pass
+    target_path = "openfold3.core.model.primitives.initialization.trunc_normal_init_"
+    with patch(target_path, noop_init):
+        yield
```
jnwei, line 549: Nit: Prefer not to use the unittest module in non-test code.

We can avoid using unittest.mock.patch here if we use the following wrapper code instead:

```
@contextlib.contextmanager
def skip_random_init():
    import openfold3.core.model.primitives.initialization as m
    original = m.trunc_normal_init_
    m.trunc_normal_init_ = lambda *args, **kwargs: None
    try:
        yield
    finally:
        m.trunc_normal_init_ = original
```

openfold3/tests/test_skip_random_init.py
jnwei, line 1: This is a nice test! 

I think this test would fit inside openfold3/tests/test_entry_point.py

openfold3/tests/test_skip_random_init.py
```diff
@@ -0,0 +1,16 @@
+import openfold3.core.model.primitives.initialization as initialization
+from openfold3.entry_points.experiment_runner import skip_random_init
+
+
+def test_skip_random_init_context_manager():
+    original_func = initialization.trunc_normal_init_
+
+    with skip_random_init():
+        # function should be nop
```
jnwei: typo: noop

openfold3/tests/test_model_checkpoint.py
```diff
@@ -169,3 +172,32 @@ def test_load_model_ckpt_with_missing_fields_fails(
         inference_runner = InferenceExperimentRunner(inference_config)
         with pytest.raises(RuntimeError):
             inference_runner.setup()
+
+    def test_inference_load_state_dict_benchmark_under_ten_seconds(
```
jnwei, line 176: nice benchmark test!

jnwei (changes requested):
Thank you for this contribution @rizsp ! This will certainly help speed up the start of inference.

A few small stylistic nits and we should be good to go.

rizsp:
Thank you for the review, @jnwei! The newest commit should resolve the outstanding issues:
* Moved the skip random init test into test_entry_points
* Replaced the unittest.mock patch following your suggestion

rizsp (commented):
Resolved issues

jnwei (approved):
Great, thank you @rizsp !

rizsp:
I assume due to xdist the pytest-benchmark plugin is disabled, this is from the log:

> /opt/conda/envs/openfold3/lib/python3.12/site-packages/pytest_benchmark/logger.py:44: PytestBenchmarkWarning: Benchmarks are automatically disabled because xdist plugin is active.Benchmarks cannot be performed reliably in a parallelized environment.

This is why the benchmark object is missing the stats attribute.

Maybe the test can be guarded against a missing pytest-benchmark by checking if the benchmark fixture is an actual benchmark fixture? Something like:

```python
if benchmark.stats is None:
    # optionally emit some warning
    return
```

jnwei:
Yes, I think you're right @rizsp . I've added the fix you suggested and one other to help with envrionment variable leakage between tests. Hopefully should be good after this.

GMNGeoffrey:
You might want to look at https://docs.pytorch.org/docs/2.12/generated/torch.nn.utils.skip_init.html. Requires threading through the device kwarg for the nn modules, but that's pretty easy and nice to have anyways (allows initializing directly on GPU)

PR #223, opened 2026-05-18
Make pairformer_emb_per_sample call into pairformer_emb

GMNGeoffrey:
**Summary**
The current code has some duplication that also causes the two paths to diverge. In particular, the batched path currently flattens the batch dimensions whereas the per_sample path does not. This makes some downstream things confused (the issue I ran into was the chunk tuner expecting the same function to always be called with the same rank inputs, which is probably a bug as well). But also, to the extent that the flattening needs to happen, then I think it should happen for all paths (there's a disadvantage in that it can force implicit broadcasts to become explicit).

**Changes**
- Make `per_sample_pairformer_emb` a thin wrapper around `pairformer_emb`
- Add validation that all inputs to PairformerEmbedding have the same number of batch dimensions
- Fix slicing of `x_pred` to not assume that num_samples is the 1st dimension
- Fix slicing of `zij` in a couple other places that assumed num_samples is the first dimension
- Add tests for equivalence of per_sample and batched pairformer embedding paths

**Related Issues**
- Fixes #199

**Testing**
- Added unit tests for per_sample path equivalence
- I marked this as fixing #199 in that it documents the input expectations. The example reproducer there would immediately throw a `ValueError` now.

Issue #199, opened 2026-04-30
[BUG] per_sample_pairformer_emb passes wrong-shaped x_pred to embed_zij


jnwei:
**Describe the bug**
In openfold3/core/model/heads/prediction_heads.py, line 149, The PairformerEmbedder passes `x_pred` with the wrong dimensions to the pair embedder.

inside per_sample_pairformer_emb, each sample is indexed as:

```
zij_chunk = self.embed_zij(
    si_input=si_input, zij=zij, x_pred=x_pred[:, i : i + 1]
)
```
`x_pred[:, i:i+1]` has shape `[batch, 1, n_token, 3]` — it retains the sample dimension. embed_zij expects x_pred shaped `[batch, n_token, 3]`. With the extra leading 1, the distance computation inside embed_zij produces dij shaped `[batch, 1, n_token, n_token, 1]`, so `self.linear_distance(dij)` is `[batch, 1, n_token, n_token, c_z]`. Adding this to `zij [batch, n_token, n_token, c_z]` broadcasts the result to 5D `[batch, 1, n_token, n_token, c_z]`. Downstream attention then fails when trying to add a 5D bias into 4D attention scores.

**To Reproduce**
Below is a sample script that uses PairformerEmbedding, resulting in the following error

RuntimeError: output with shape [1, 16, 10, 10] doesn't match the broadcast shape [1, 1, 16, 10, 10]

<details> <summary>Minimal reproducer</summary>

import torch
from openfold3.core.model.heads.prediction_heads import PairformerEmbedding
from openfold3.projects.of3_all_atom.project_entry import OF3ProjectEntry

proj_entry = OF3ProjectEntry()
config = proj_entry.get_model_config_with_presets()

c_s_input = config.architecture.shared.c_s_input
c_s = config.architecture.shared.c_s
c_z = config.architecture.shared.c_z

batch_size = 1
n_sample = 2
n_token = 10

pair_emb = PairformerEmbedding(
    **config.architecture.heads.pairformer_embedding
).eval()

si_input = torch.randn(batch_size, n_token, c_s_input)
si = torch.randn(batch_size, n_token, c_s)
zij = torch.randn(batch_size, n_token, n_token, c_z)
x_pred = torch.randn(batch_size, n_sample, n_token, 3)  # [B, N_sample, N_token, 3]
single_mask = torch.ones(batch_size, n_token)
pair_mask = torch.ones(batch_size, n_token, n_token)

with torch.no_grad():
[...]

PR #224, opened 2026-05-18
Treat arg-cache length mismatch as a cache miss in ChunkSizeTuner

GMNGeoffrey:
**Summary**
The chunk size tuner currently throws if the function it's tuning for receives arguments with different ranks in different calls. Instead, treat this as a cache miss. I don't think the tuner should be forcing invariants on the functions it's tuning for.

I encountered an issue with this when doing https://github.com/aqlaboratory/openfold-3/pull/213. That enables `batch > 1` and chunking to happen together with Triton kernels (you hit the same bug with kernels off without that change). When you run inference with samples both above and below `per_sample_token_cutoff`, the small inputs use the batched/normal pairformer embedding path which flattens the batch dimensions (i.e. batch and num_samples), but the larger inputs follow the per-sample path, which *doesn't* flatten their batch dimensions and results in higher-rank inputs (https://github.com/aqlaboratory/openfold-3/pull/223 aims to fix this mismatch).

`_compare_arg_caches` recurses on tensor shapes (`torch.Size` is a tuple subclass), so when the same `ChunkSizeTuner` instance was invoked with tensors of different ranks across calls, the inner `zip(..., strict=True)` raised. Instead in this change we treat any length mismatch as a cache miss so the caller re-tunes instead. The top-level length assert in `tune_chunk_size` is now handled the same way and removed.

Also adds dtype element size to the cache key for a tensor argument. We could use the entire dtype, but I think in terms of what matters for chunking, the element size is the key factor.

**Changes**
- Remove assert that tuned function always called with same number of args
- Add `dtype.itemsize` to tensor cache key along with shape
- Record a cache miss when cache key lengths differ rather than raising from `zip(...,strict=True)`
- Unit tests for the same

**Related Issues**
- #199 in that it's also related to the rank changes in the pairformer per-sample path.

**Testing**
- Unit tests verifying cache miss on arg count, arg rank, and arg dtype size changes.

### DIFF
--- a/openfold3/core/utils/chunk_utils.py
+++ b/openfold3/core/utils/chunk_utils.py
@@ -388,6 +388,9 @@ class ChunkSizeTuner:
         return candidates[lo]
 
     def _compare_arg_caches(self, ac1, ac2):
+        # When recursing this tests that tensors have the same rank
+        if len(ac1) != len(ac2):
+            return False
         consistent = True
         for a1, a2 in zip(ac1, ac2, strict=True):
             assert type(a1) is type(a2)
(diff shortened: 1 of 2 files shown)


GMNGeoffrey:
@christinaflo @jnwei PTAL

jandom (commented):
Those are all good – all 3 test for the negative case, would it make sense to test for the positive as well, or do we think that's overkill?

Meta-comment, and outside of scope, we should really migrate all these to pytest just to be consistent

GMNGeoffrey:
> Those are all good – all 3 test for the negative case, would it make sense to test for the positive as well

Done

christinaflo (approved):
LGTM!

GMNGeoffrey:
> LGTM!

Can you mark safe to test and merge? Or were you waiting for @jandom?

christinaflo:
> > LGTM!
> 
> Can you mark safe to test and merge? Or were you waiting for @jandom?

Yeah I was going to see if @jandom had other comments but I think it's fine to merge after the tests finish running.

Issue #225, opened 2026-05-19
Inference OOM from `get_token_frame_atoms`


GMNGeoffrey:
When running a collection of queries of increasing size, I'm hitting an OOM in `get_token_frame_atoms` with a 5,247 residue monomer (41,842 atoms). The OOM doesn't happen when this query is run on its own, so I think there must be some memory fragmentation going on (unfortunately `clear_cache_between_steps` didn't seem to help). But also this is just a pretty large materialization since the pairwise distance computation here is `torch.sum(eps + (x[..., None, :] - x[..., None, :, :]) ** 2, dim=-1) ** 0.5`, materializing an 3*N_atom^2 intermediate (I'm seeing 42GB of peak memory from this call alone)

Some potential fixes:

- I think we can use `torch.cdist` here. I believe by default this uses the identity `||a - b||^2 = ||a||^2 + ||b||^2 - 2 a.b` for more than 25 points (or always if you set `use_mm_for_euclid_dist`). This has some negative consequences for precision as it ends up subtracting large numbers to get a small result, but given that we then feed it into topk, this shouldn't have that big an effect. It's also about 4x faster (just for the distance computation), though I don't think that really matters here.
- The same identity can also be implemented manually with torch ops, but there's not really an advantage I can see.
- We could chunk this function as well. Gets bit-exact output, but somewhat more complex, though less so because we already have all the chunking machinery. No significant slowdown relative to current implementation, but massively reduced memory.
- Compute per coordinate separately and use a running sum. Same memory usage as cdist, without the catastrophic cancellation issues so much more precise results. Slightly faster than current implementation. Not bit-exact like chunking by atoms.

I had Claude compare the approaches. Tested by capturing snapshots of the tensors going into this function for the 14 inputs I used (49-4563 residues, 366-36,356 atoms), so we have realistic coordinates. But because distances are only used for frame atoms when is_atomized is true, I artificially flipped all of them to set them as atomized (would've been better to use actual atomized inputs, but this is what I had on hand). cdist changes the output of `phi` for four atoms total in the dataset (per input mean 0.05%, max 0.5%). None of the others change it at all and none of them change the `valid_frame_mask`. cdist and per-coord drop the memory requirement for 42k atoms from 42GB to 13GB and chunking in chunks of 1024 atoms drops it to 7GB. For any of those, something else in the function ends up dominating peak memory, which drops from 47GB to 33GB.

[...]

PR #227, opened 2026-05-20
Make config chunk_size maximum rather than minimum

GMNGeoffrey:
**Summary**
The `chunk_size` specified in the config is currently the minimum value that the tuner will go down to. I don't think there's a clear use case for this to be the minimum and users were confused. Instead, we can make this the *maximum*. This also allows us to remove the hardcoded values for the max with default and flash-style kernels: these instead are just config settings. I removed the minimum chunk size entirely, so the tuner will search all powers of 2 up to the maximum. I'm struggling to think of a use case where someone would want to limit chunks size to a minimum of 4 (I think this number just came from the original suggested chunk size in the AlphaFold 3 paper).

Note that this changes the hardcoded different chunk size for diffusion conditioning. I'm not sure how important this is to preserve and whether we saw actual latency improvements from raising the chunk size here vs just observing that it was possible to do so. If we want different chunk sizes for different modules, I think we should probably thread through a config setting per module because having a hardcoded value here is a bit surprising IMO (and note that it only gets applied when tuning is turned on). Probably the best analog in the config would be `offload_inference` as a per-module setting that isn't an init arg. If more things got added here, IDK whether we'd want to keep this structure or perhaps mirror the structure of architecture.

**Changes**
- Make config `chunk_size` indicate the maximum chunk size when tuning is active
- Remove hardcoded min and max chunk size constants
- Update default config to specify chunk size 1024 (since flash kernels are used by default)

**Testing**
- Added unit tests for the chunk tuner


GMNGeoffrey:
@christinaflo PTAL

examples/example_runner_yamls/cuequivariance.yml
```diff
@@ -1,10 +1,11 @@
...
     - low_mem  # to use low memory settings
   custom:
     settings:
       memory:
         eval:
           use_cueq_triangle_kernels: true
-          use_deepspeed_evo_attention: true
\ No newline at end of file
+          use_deepspeed_evo_attention: true
+          chunk_size: 1024
```
christinaflo: I was thinking more that we would keep these defaults and allow people to change them if they wanted to for max chunk size. like whatever kernel we make the default should default to the associated max chunk size

openfold3/core/model/latent/base_stacks.py
```diff
@@ -155,7 +147,7 @@ def block_with_cache_clear(block, *args, **kwargs):
                     chunk_size=tuned_chunk_size,
                     # A temporary measure to address torch's occasional
                     # inability to allocate large tensors
-                    _attn_chunk_size=max(chunk_size, attn_chunk),
+                    _attn_chunk_size=attn_chunk,
```
christinaflo, line 150: If it picks 1 or 2 for the chunk size since we dont set a min, this will become 0 if no kernel is used

examples/example_runner_yamls/cuequivariance.yml
GMNGeoffrey: Not sure if I'm understanding you. Did you want to keep the constants in the code? Or you're saying that the default config (not this example) should have chunk size 1024? Right now I think the default based on model_config.py is to have all the kernels off so I made the default there 512. If we swapped one of them to on by default we could update the default chunk size as well.

openfold3/core/model/latent/base_stacks.py
GMNGeoffrey: Good catch! Added an e2e test of chunk size 1 and fixed (lemme know if you think the test is overkill)

examples/example_runner_yamls/cuequivariance.yml
christinaflo: The default is now to have DS on for NVIDIA (we are going to retire DS at some point tho) and triton on for AMD for inference, so we could make it 1024 in the model config. Originally I wanted it to keep the constants so the user doesnt have to know to specify the chunk size when they enable/disable a kernel. I guess we could do the reverse and lower it automatically if no kernels are enabled. But I'm kind of overcomplicating it, probably best to just bump the default value.

openfold3/tests/test_of3_model.py
```diff
@@ -59,8 +60,12 @@ def run_model(
...
         )
 
-        if use_triton_triangle_kernels:
-            config.settings.memory.eval.use_triton_triangle_kernels = True
+        config.settings.memory.eval.use_triton_triangle_kernels = (
+            use_triton_triangle_kernels
+        )
+        if chunk_size is not None:
+            config.settings.memory.eval.chunk_size = chunk_size
+            config.settings.memory.train.chunk_size = chunk_size
```
christinaflo: We never really run chunking with training because the activations stack up anyway in the backward pass so it doesnt save you much. I havent actually run this but i think it may fail some assert. Diffusion conditioning at least has a `assert not self.training` for its chunking function

christinaflo: ah okay i see now you only have a test for eval mode anyway

openfold3/tests/test_of3_model.py
```diff
@@ -153,6 +158,25 @@ def test_shape_small_fp32(self, model_phase):
             use_deepspeed_evo_attention=False,
         )
 
+    def test_shape_small_chunk_size_one(self):
```
christinaflo, line 162: Actually this reminds me, chunk_size=4 is probably hardcoded in a bunch of tests like here:
openfold3/tests/test_msa_module.py
openfold3/tests/test_evoformer.py
openfold3/tests/test_kernels.py
openfold3/tests/test_heads.py
openfold3/tests/test_pairformer.py
openfold3/tests/test_utils.py

these should also be updated

examples/example_runner_yamls/cuequivariance.yml
GMNGeoffrey: Oh, where is that set? I don't see anywhere that deepspeed is defaulted to on other than cuequivariance.yml, which is where I'm setting chunk size to 1024.

openfold3/tests/test_of3_model.py
GMNGeoffrey: We don't want to keep it small so we exercise the chunking path even for small inputs?

examples/example_runner_yamls/cuequivariance.yml
christinaflo: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/projects/of3_all_atom/config/model_config.py#L116
the default config for train/eval is in this file and model presets have some settings on top of that for inference/low mem settings (this should really be condensed)

openfold3/tests/test_of3_model.py
christinaflo: ah no yeah you're right, i'd only care about it being changed if tuning is enabled but I think it's not for any of these anyway

examples/example_runner_yamls/cuequivariance.yml
GMNGeoffrey: Ah, thank you. I actually did know about that file, but I think was reading the train section by mistake and then was searching for `use_deepspeed_evo_attention.*true` and not finding anything. So you'd like me to set 1024 here and leave it off in the individual yaml files?

christinaflo: lets set it here: https://github.com/GMNGeoffrey/openfold-3/blob/d89aeae80e0211a342037786e266118082211ed7/openfold3/projects/of3_all_atom/config/model_setting_presets.yml#L18 bc it's more an inference than validation setting
and leave it out of the kernel configs

openfold3/tests/test_of3_model.py
christinaflo: can we just delete this line anyway since it cant run

GMNGeoffrey: In that case, shouldn't we remove `chunk_size` from the train settings entirely? I set it here because if chunk size is overridden for the test it would be surprising if it weren't overridden for training as well IMO. I can delete it if you prefer though

examples/example_runner_yamls/cuequivariance.yml
GMNGeoffrey: Done

openfold3/tests/test_of3_model.py
christinaflo: I had it there originally because I needed chunking enabled during validation for some large samples but not for training, so it'll pick what to use in the model like: `chunk_size=mode_mem_settings.chunk_size` depending on the stage it's in. I guess there's nothing stopping you from doing it during training, it's just not really worth it ever. 

We could fix the assert in diffusion conditioning to match the other modules so it runs:
```
if chunk_size is not None and self.chunk_size_tuner is not None:
    assert not self.training
```
or change model.py to always set it to None for training and not reference the config, i thought it was easier to distinguish in the config though

GMNGeoffrey: Yeah it looks like all the others just assert no chunk size *tuning* during training. It would maybe be nice if they matched.

So the reason we have chunk_size defined in the training config is just because mode_mem_settings is agnostic to the actual mode so they need to have the same schema? I see now. I've replaced this line with `assert not train` so we don't have a case where someone sets chunk size and train=True and gets unexpected behavior.

christinaflo: Yeah I can update that diffusion conditioning one later to match. But yes it's because it's agnostic to the mode. All looks good now aside from merge conflicts. One thing that I want to do is have a benchmark we run with the old chunk tuning logic and the updated ones (just this one and #213 really) that checks that the speed and memory usage at different seq lengths is comparable or better than the old one, like making sure we reach the same max seq length also. For this one, I don't really see a reason it would be any different, and maybe can get to larger seqs without the minimum, but still would make me feel better to double check. Maybe for this one we can just check max seq length is the same for both and then we can do the benchmarking just for #213.

GMNGeoffrey: Yeah I'll fix the merge conflicts once https://github.com/aqlaboratory/openfold-3/pull/224 lands.

I think this is really just a configuration change that takes us from hardcoded maximum and configurable minimum to configurable maximum and hardcoded minimum (albeit minimum 1). At least for a server-class GPU like MI300X, I think these allocations are actually not in the peak memory path at the maximum token count that fits (https://github.com/aqlaboratory/openfold-3/issues/225 is) and they just always use the max chunk size. So I don't think it's really possible for this to change the memory usage or performance in that scenario. I'm happy to run a sanity check though if you have something in mind. Do you have some standardized benchmarks/inputs? I've been validating things with some example inputs, but it would be nice to have something standard. I know there's `random_of3_features`, but I'm not sure of my ability to pick `n_msa` and `n_templ` realistically, so I've been sticking to real inputs.

christinaflo: Oh yeah thats a lot of memory. on 80gb gpus with no kernels I can see the chunking take effect with seq lengths > 1500 tokens. I have a really old and messy script for benchmarking inference speed + mem using random_of3_features. I can clean it up and share it, but I have a bit of a backlog this week so I can just send it for testing #213 so I don't block this PR. As you said, this is really just a config change. Btw realistic values are `n_msa=16384` (this will get subsampled to 1024 each recycle) and `n_templ=4`.

GMNGeoffrey: Perfect, thanks! I think that gives me enough to do testing without being limited by what I can find in the pdb. A standardized script in-repo would be nice, but you don't need to rush to clean up yours 🙂

So those don't vary much with input or scale with n_tok? I found that homoers used significantly less memory at very large n_tok I think due to msa reuse for the shared sequences, but I didn't dig in.

I can test for smaller memory caps and the chunk tuner behavior by limiting torch mem_fraction. The only problem is the combinatorial explosion of options. If 80gb is of particular interest (H100?) I can test it as well. The other option is to test fixed chunk sizes and just track memory. I've got a memory snapshotting callback (happy to clean up and upstream if it would be generally useful). The only issues there are that tuning itself can affect peak due to some clones and then diffusion conditioning getting chunk size 2048 is guarded by the tuner getting on (is the diff there worth it?)

christinaflo: It's just the max allowable input `n_msa` and `n_templ`, there could be less but it'll get capped at those numbers. I'm not sure why large homomers use less memory either off the top of my head, I'd have to look into that as well, since msas should still be capped at 1024 per recycle I'd expect they would still reach that threshold. 
Yeah 80 gb h100 is generally what we have access to so that is of particular interest. Even on mi300A I try to limit model gpu memory quite a bit anyway since our data loader processes take up a lot of memory, so I kind of treat it like an h100. 
For the memory snapshotting, I have a callback for this also internally that will get merged in at some point, I assume they're probably the same. 
For the diffusion conditioning chunk size, this PR changes it to the global max but I think it's fine, I highly doubt it's that much slower and I eyeballed that number to begin with :).

GMNGeoffrey: Ah I looked closer and found the issue. One of the homomers I had Claude dig up for me (6R7M-1) is a 40-chain homomer with an MSA depth of only 122. This was throwing things off and I thought it would be an issue with all homomers to a lesser extent, but it's really just this one that is weird. If n_msa gets subsampled to 1024 every recycle, does it actually make a difference if the random input has n_msa=1024 or n_msa=16384?

> For the diffusion conditioning chunk size, this PR changes it to the global max but I think it's fine, I highly doubt it's that much slower and I eyeballed that number to begin with :).

Oh right, I forgot that I did that already :-D

christinaflo: No actually it doesn't make a difference, it's only if you want to exercise the subsampling logic which doesn't matter here.

GMNGeoffrey:
Validated that this doesn't impact correctness or performance and doesn't reduce the size of the largest input we can fit. I did single-batch runs using synthetic inputs created from `random_of3_features` with the Triton kernels and stock PyTorch, with and without an 80GiB memory cap to simulate a smaller GPU. I left the configured chunk size at 1024 even with the stock operations. The graphs here are based on running the full experiment runner with a dataloader as opposed to just invoking the model, which is why there's more memory usage and fragmentation and we can't get quite as large as the analysis I did in #248. I can share my synthetic dataset loader as well.

In general, the largest input that fit was the same before and after this patch. In one case: Triton kernels with no memory cap, the largest fit rose from 4690 to 4750 residues, perhaps due to the smaller default diffusion chunk size leaving slightly more headroom.

With the Triton kernel, we don't hit any OOMs and the default chunk size is used, so the only difference is lowering diffusion to 2048→1024.

With the stock torch ops, the chunk sizes chosen were (base→patched where they differ):

full GPU:

| n_token   | Templ    | MSA      | PairF    | DiffCond  |
| --------- | -------- | -------- | -------- | --------- |
| 1000      | 512→1024 | 512→1024 | 512→1024 | 2048→1024 |
| 2000      | 512→1024 | 512→1024 | 512→1024 | 2048→1024 |
| 3000      | 512→1024 | 512→1024 | 512→1024 | 2048→1024 |
| 4000      | 512→256  | 512      | 512→256  | 2048→1024 |
| 4500–4590 | 256      | 256      | 256      | 2048→1024 |

80 GiB cap:

| n_token | Templ    | MSA      | PairF    | DiffCond  |
| ------- | -------- | -------- | -------- | --------- |
| 1000    | 512→1024 | 512→1024 | 512→1024 | 2048→1024 |
| 2000    | 512→1024 | 512→1024 | 512      | 2048→1024 |
| 2500    | 512      | 512      | 256      | 2048→1024 |
| 2750    | 512      | 512      | 256      | 2048→1024 |
| 2875    | 256      | 256      | 128      | 2048→1024 |
| 2910    | 256      | 256      | 256      | 2048→1024 |

I also tracked allocated memory up to the largest input that fit. Note that with fragmentation, the memory reserved is larger, so we hit OOMs even below the full memory capacity.

image

With the Triton kernel, this is not the memory bottleneck and the chunk size stays at the max all the way through. The slightly smaller diffusion chunk size allows us to squeeze in a slightly larger input at uncapped memory.

With the stock kernel, the patched version initially uses more memory for smaller inputs due to the larger default chunk size, but as we near the ceiling, the tuner kicks in and picks smaller chunk sizes so things ultimately fit. One interesting behavior at 4k input tokens is that the patched version chooses a smaller chunk size. I think this is probably because testing 1024 uses some memory so then 512 doesn't fit. Also interesting is that for some reason in both cases it uses chunks of 128 for the second PairFormer input at 2875 tokens but goes back to 256 for 2910. I blame slightly different memory fragmentation

None of this appears to have a significant impact on performance. Mostly a +/-3% variation. Where the patched version tries 1024 and OOMs (stock 4k or 2k with the memory cap) there's about a 10% overhead. I think this is just inherent to how the tuner works. As discussed, a smaller chunk size is probably better for the stock pytorch kernels on a smaller GPU.

The output is mostly identical. There's some non-determinism where different chunk sizes are chosen leading to different accumulation order and bf16 imprecision; max coord diff of 0.009 Angstroms.

@christinaflo I think this is good to go. 1024 isn't ideal for stock kernels. If we wanted to preserve the different default there, I think we could look at whether kernels are turned on and set the default different in model_config.py. But also at some point this just becomes a configuration option that people can choose. We could provide a "no kernels" example yaml that sets it smaller, since the default now uses the kernels in some way

christinaflo:
Thanks for checking all this @GMNGeoffrey! Looks good to merge after removing the old arg from the unit tests.

GMNGeoffrey:
Oh whoops! Done

jandom:
@GMNGeoffrey thanks for this PR – @christinaflo has asked me to unblock the CI/CD on this PR, and I think it's picking up some real test failures. Would you have bandwidth to take a look this week?

GMNGeoffrey:
The test is just asserting what the value of the chunk config parameter is, which has now changed. Updated.

PR #228, opened 2026-05-21
Add checks against expanded pair bias

GMNGeoffrey:
**Summary**
The Triton kernel assumes that pair bias has size 1 in the N_seq dimension and does an implicit broadcast (`pair_bias.stride(1)` is not even passed to the kernel), but this isn't actually checked. This is the reason chunking is turned off when batch size > 1 and we're using optimized kernels, because the chunker expands the implicit broadcast dimensions. Added asserts for this and documented the issue.

**Changes**
- Assert on expected bias shapes in Triton triangle attention
- document why chunking gets turned off in prediction_heads.py when using optimized kernels
- Update tests for Triton kernels in test_kernels.py to avoid batch_size > 1 + chunking (was only set for deepspeed). Cuequivariance has a workaround, so its tests still use the larger batch.

**Related Issues**
- Related to #206

**Testing**
- Updated the tests to not use batch size > 1 (existing tests would have thrown)
- I don't think a test checking for the assert is very useful

**Other Notes**
This, along with #226, contains the parts of #207 that I think we still want even if the approach in #213 is preferred overall. If we merge this, then I'll close #207 and iterate on #213 instead.

### DIFF
--- a/openfold3/tests/test_kernels.py
+++ b/openfold3/tests/test_kernels.py
@@ -438,8 +438,7 @@ class TestKernels(unittest.TestCase):
         if chunk_size is not None and (
             use_deepspeed_evo_attention or use_triton_triangle_kernels
         ):
-            # Chunk tuning is not supported with batch size > 1 for DeepSpeed kernel
-            # Triton kernel works with these shapes but has some numerical differences
+            # Chunk tuning is not supported with batch size > 1 for these kernels
             batch_size = 1
 
         n_res = 200  # Avoid cuEq seq len constraints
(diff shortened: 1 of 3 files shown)


GMNGeoffrey:
@christinaflo PTAL

openfold3/tests/test_kernels.py
```diff
@@ -719,13 +719,13 @@ def _compare_template_stack(
...
-        attention kernel. Kernel can be used for Triangle Attention in the Template Pair
-        Stack.
+        Compare Template Stack output with and without using different optimized
+        attention kernels. Kernel can be used for Triangle Attention in the
+        Template Pair Stack.
         """
         batch_size = consts.batch_size
-        if chunk_size is not None and use_deepspeed_evo_attention:
-            # Chunk tuning is not supported with batch size > 1 for DeepSpeed kernel
+        if chunk_size is not None:
```
christinaflo: our cueq function does have a fix in place for this (#181, #204), it's not ideal in that it broadcasts the bias dim, but it does work. for cueq at least we should leave the test as is

GMNGeoffrey: Thanks for flagging; fixed.

christinaflo (approved):
LGTM!

PR #35, opened 2025-11-12
Preserve Raw ColabFold MSA Output Files

qurat-ul-ain95:
**Summary**

These changes preserve raw ColabFold MSA output files (`.a3m` files) if the user disables `cleanup_msa_dir`. Previously, the `raw` directory containing batch MSA files was always deleted, regardless of the `cleanup_msa_dir` setting, preventing users from accessing the raw colabfold MSA data for inspection or reuse.

The pipeline sends batched requests to Colabfold containing multiple sequences. Suggested change in this PR separates out the resulting files by the unique sequence identifier `rep_id` and stores them in a new `raw_colabfold_output` folder which persists between runs.

**Changes**

1. **`_parse_a3m_file_by_m()` function**: util function to parse batch A3M files and extract individual MSA sections by M value (ColabFold's internal sequence identifier). This function handles:
   - Null byte delimiters (`\x00`) that separate M sections in batch files
   - UniRef headers vs. M value headers
   - Optional filtering by M values

2. **`_organize_raw_main_outputs_by_query()` method**: Extracts individual MSA sections from batch A3M files and saves them in a persistent directory `raw_colabfold_output`:
   - **Directory structure**: `raw_colabfold_output/{rep_id}/{filename}.a3m`
   - **Files preserved**: `uniref.a3m` and `bfd.mgnify30.metaeuk30.smag30.a3m` for each `rep_id`
   - **Called after**: Main MSA processing completes, before cleanup

3.  **`class TestParseA3mFileByM`** contains a few pytests to check `_parse_a3m_file_by_m()` functionality. 

Closes #31

PR #233, opened 2026-05-28
Raise FileExistsError if colabfold raw directory is found before performing colabfold alignments

### LINKED ISSUE #39: Always delete `raw` Colabfold folder before processing
Hi Team,

This is not an issue per se but  is definitely a strong suggestion by an end user. It's possible for an openfold job to fail for N number of reasons and a lot of the times this happens before cleanup: i.e. the `raw` colabfold output folder and stale out.tar.gz files remain in the MSA dir. Subsequent runs skip hitting the Colabfold API because of 

```python
`if not os.path.isfile(tar_gz_file):` 
```  
in core/data/tools/colabfold_msa_server.py 

which leads to the pipeline using older colabfold raw output folder for the new query as this check only checks file existence, not whether it matches the current query. This results in nondescript downstream errors which are very difficult for the end user to investigate.

This check for tar_gz_file is redundant anyway because if the query runs properly, this tar_gz_file is deleted along with the entire raw folder. On rerun, the pipeline always calls the API again.

My suggestion is to always delete the `raw` colabfold folder before this line runs. What do you think? Thanks!

jnwei:
**Summary**
The existance of the `raw` colabfold directory, in particular, the `raw/out.tar.gz` file from a previous run can cause the colabfold pipeline to assume that the colabfold job has been previously completed and skips submitting a new job. 

In the worst case scenario, if a previous inference job crashes, and a new one is submitted, the wrong alignments from the outdated `out.tar.gz` file will be parsed and create potentially silent issues for prediction.

This PR raises an FileExistsError in the case that the colabfold raw directory still exists, and instructs the user to delete the directory. At this time, we leave the task of deleting the directory on the user in case they want to inspect or save the raw files.

**Changes**
- In `preprocess_colabfold_msas` add an error if the `raw/` output directory exists
- Add a new test `TestColabFoldQueryRunner.test_preprocess_raises_if_raw_dir_exists` to check for FileExistsError
- Updates two previous tests as a result of the check:
   - `TestColabFoldQueryRunner.test_empty_m8_file_handling `: Replace the full `preprocess_colabfold_msas` call with `add_msa_paths_to_iqs`. I believe the intended purpose is to check the assignment of the templates alignments / entry ids upon receiving an empty template pdb70.m8 file.  The rest of `preprocess_colabfold_msas` is already performed separately in this test in order to add additional assertions and checks.
  - `TestColabFoldQueryRunner.test_features_on_multiple_queries_with_same_name`: Delete the raw directory upon completion of the test loop.

**Related Issues**
#39

**Testing**
- [x] Passes all tests in `openfold3/tests/core/data/tools/test_colabfold_msa_server.py`

**Other Notes**
- We may still want a more extensive fix as proposed in #35. This is a more minimal fix to help navigate around the footgun that arises in #39

### DIFF
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -1141,6 +1141,16 @@ def preprocess_colabfold_msas(
     if compute_settings.save_mappings:
         save_colabfold_mappings(colabfold_mapper, output_directory)
 
+    # Abort early if a stale raw directory exists — it contains unvalidated
+    # out.tar.gz files that would be silently reused for the wrong query.
+    raw_dir = output_directory / "raw"
+    if raw_dir.exists():
+        raise FileExistsError(
+            f"ColabFold raw directory already exists: {raw_dir}\n"
+            "This is likely left over from a previous failed run. "
+            "Please remove it before starting a new run."
+        )
+
     # Run batch queries for main and paired MSAs
     colabfold_query_runner = ColabFoldQueryRunner(
         colabfold_mapper=colabfold_mapper,
--- a/openfold3/tests/core/data/tools/test_colabfold_msa_server.py
+++ b/openfold3/tests/core/data/tools/test_colabfold_msa_server.py
@@ -15,6 +15,7 @@
 """Tests for the ColabFold MSA server module."""
 
 import json
+import shutil
 import textwrap
 from pathlib import Path
 from unittest.mock import patch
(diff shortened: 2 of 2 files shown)


jnwei:
> `TestColabFoldQueryRunner:test_empty_m8_file_handling`: Replace the full `preprocess_colabfold_msas` call with `add_msa_paths_to_iqs`. I believe the intended purpose is to check the assignment of the templates alignments / entry ids upon receiving an empty template `pdb70.m8` file. The rest of `preprocess_colabfold_msas` is already performed separately in this test in order to add additional assertions and checks.

@qurat-ul-ain95 Could you verify that the above change to the test `test_empty_m8_file_handling` is correct. I think I have understood the logic of that test, but I'd appreciate your validation as well.

jandom (approved):

PR #238, opened 2026-05-30
fix: trailing slash in colabfold URL

### LINKED ISSUE #237: [BUG] colabfold return invalid ID for ubiquitin example
**Describe the bug**

`run_openfold predict` fails at MSA generation with a misleading error:

```
ERROR: Server didn't reply with json: invalid ID
Exception: MMseqs2 API is giving errors. Please confirm your input is a valid protein sequence.
```

The server is healthy and the sequence is valid. The real cause is a doubled slash in the request URL: `MsaComputationSettings.server_url` is a pydantic `Url`, whose `str()` includes a trailing slash. Combined with the `f"{host_url}/{submission_endpoint}"` f-string in `query_colabfold_msa_server`, this produces `https://api.colabfold.com//ticket/msa`. The server 301-redirects it; `requests` follows the redirect but downgrades `POST` → `GET` and drops the body, so the server returns `400 invalid ID`, which the code cannot parse as JSON and converts to `{"status": "ERROR"}`.

Regression introduced in commit `f958eeb9` ("add msa settings to experiment config") when `server_url` was changed from `str` to `Url`. Existing tests mock `query_colabfold_msa_server` entirely, so URL construction was never tested against the live server.

**To Reproduce**

```bash
run_openfold predict \
  --query-json examples/example_inference_inputs/query_ubiquitin.json \
  --runner-yaml examples/example_runner_yamls/low_mem.yml
```

**Expected behavior**

MSA submission proceeds with a single-slash URL and returns a valid ticket.

**Stack trace**

```
File ".../colabfold_msa_server.py", line 282, in query_colabfold_msa_server
    raise Exception(
Exception: MMseqs2 API is giving errors.Please confirm your input is a valid protein sequence.
```

**Configuration:**
 - OS: Ubuntu 24.04.4 LTS
 - GPU: NVIDIA GB10
 - Python: 3.14.5
 - Installation: build from repo (pixi, `openfold3-cuda13-pypi`)

**Additional context**

Fix: add `host_url = str(host_url).rstrip("/")` in `query_colabfold_msa_server` before the nested helpers, so all request URLs are built without a doubled slash regardless of input type.

jandom:
Fixes https://github.com/aqlaboratory/openfold-3/issues/237

### DIFF
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -112,6 +112,12 @@ def query_colabfold_msa_server(
 
     submission_endpoint = "ticket/pair" if use_pairing else "ticket/msa"
 
+    # Normalize the host: a pydantic Url stringifies with a trailing slash, which
+    # would produce a doubled slash (e.g. ".../com//ticket/msa") in the f-strings
+    # below. The server 301-redirects the doubled slash, and requests downgrades
+    # the POST to GET while dropping the body, yielding a misleading "invalid ID".
+    host_url = str(host_url).rstrip("/")
+
     headers = {}
     if user_agent != "":
         headers["User-Agent"] = user_agent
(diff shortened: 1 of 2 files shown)


jnwei (approved):
LGTM! thanks for identifying the issue

PR #242, opened 2026-05-31
fix: sys.getrefcount on py3.14

jandom:
# Fix `sys.getrefcount` assertions broken by Python 3.14 LOAD_FAST optimisation

## Problem

`run_openfold predict` with the `low_mem` preset (or any config that enables both
`offload_inference` and `per_sample_token_cutoff: 0`) crashes during inference with:

```
File "prediction_heads.py", line 165, in per_sample_pairformer_emb
    assert sys.getrefcount(si_chunk) == 2
AssertionError
```

The assertion is a sanity check that a tensor about to be offloaded to CPU has no
unexpected Python references holding it on the GPU. It was working correctly, but
Python 3.14 changed how `sys.getrefcount` counts references for function-local
variables, causing false failures.

## Python 3.14 `sys.getrefcount` change

Python 3.14 optimised the `LOAD_FAST` bytecode instruction — loading a local
variable no longer pushes a temporary reference onto the evaluation stack before
a function call. This affects `sys.getrefcount` in two ways that interact:

- **`LOAD_FAST` (function-local variable):** Python 3.14 uses a "steal" calling
  convention — the caller's reference is transferred rather than copied, and
  `getrefcount` itself no longer adds a temporary. For a sole holder, the count
  drops from `2` (old) to `1` (3.14+).

- **`BINARY_SUBSCR` (subscript like `input_tensors[1]`):** the subscript
  operation still creates a counted temporary on the evaluation stack in all
  Python versions. The `== 2` expectation (1 in container + 1 subscript temp)
  is unchanged and these assertions did not need fixing.

The assertions in this codebase fell into two groups — one group broke, the other
did not — purely because of this bytecode-level distinction.

## What broke

```python
# FAST_LOCAL pattern — variable is the sole Python holder
si_chunk, zij_chunk = self.pairformer_emb(...)
assert sys.getrefcount(si_chunk) == 2  # BROKEN in Python 3.14: returns 1, not 2
```

These are in `prediction_heads.py` (`si_chunk`, `zij_chunk`) and
`template_module.py` (`t`), where the assertion was added alongside per-sample
offload logic.

## What did NOT break

```python
# BINARY_SUBSCR pattern — tensor held in a list; subscript temp still counted
assert sys.getrefcount(input_tensors[1]) == 2  # still returns 2 in Python 3.14
```

These are in `base_blocks.py`, `evoformer.py`, and `msa_module.py` and are left
unchanged.

## Fix

Add `assert_sole_holder(t)` in a new `openfold3/core/model/utils.py`. The
expected `getrefcount` value differs across Python versions **and** between
inline use and helper-function use, because each Python version handles the
calling convention differently:

| Python | `getrefcount(t)` inline | via helper |
|---|---|---|
| ≤ 3.13 | 2 (sole + call temp) | 3 (caller frame + param + call temp) |
| 3.14+  | 1 (sole, no temp)   | 1 (param only, caller ref stolen, no temp) |

The helper uses `expected = 1 if sys.version_info >= (3, 14) else 3`.

> **Why not use the helper for `input_tensors[i]` subscripts?**
> A subscript expression creates a reference that persists into the helper's
> frame in all Python versions, shifting the expected count. The `== 2` inline
> pattern is correct and simpler there.

### Files changed

| File | Change |
|---|---|
| `openfold3/core/model/utils.py` | **New file.** `assert_sole_holder` with docstring explaining the LOAD_FAST vs BINARY_SUBSCR distinction and why it must not be called with subscript expressions. |
| `openfold3/core/model/heads/prediction_heads.py` | Replace broken `sys.getrefcount(si_chunk/zij_chunk) == 2` with `assert_sole_holder`. |
| `openfold3/core/model/latent/template_module.py` | Replace prior ad-hoc inline fix with `assert_sole_holder`. |

`base_blocks.py`, `evoformer.py`, `msa_module.py` — no change (subscript
pattern, already correct in all Python versions).

## Testing

```bash
python -m pytest openfold3/tests/test_embedders.py::TestTemplateEmbedders::test_template_module_offload

run_openfold predict \
  --query-json examples/example_inference_inputs/query_ubiquitin.json \
  --runner-yaml examples/example_runner_yamls/low_mem.yml
```


GMNGeoffrey:
It doesn't look like the fix described in the pr description matches the files changed. I don't see the helper function added at all, just two inline edits. I would add the helper function and include in the docstring that it doesn't work for container subscript references and then use that without comment at all the locations that need it

jandom:
That's fair – claude dropped before i made the edits, the function itself added a refcount, so that became pointless. Now we just have this on repeat in a few places. Not ideal. 

> I would add the helper function and include in the docstring that it doesn't work for container subscript references and then use that without comment at all the locations that need it

That's probably the least bad option honestly, sorry for the low quality PR – hard to focus properly during a hackathon :P

openfold3/core/model/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Used in offload-inference loops to verify that a tensor has no unexpected
+    references before it is moved to CPU, ensuring in-place memory reuse works
+    as intended.
+
+    **Calling convention:** pass a plain function-local variable, never a
+    container subscript expression.  A subscript like ``container[i]`` creates
+    an additional counted reference on the evaluation stack, so the expected
+    count differs — use ``assert sys.getrefcount(container[i]) == 2`` inline
+    for that case instead.
```
GMNGeoffrey: Nit: I wouldn't say this is really a "calling convention", since it's a hard constraint.

openfold3/core/model/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+    The expected count changes across Python versions because of the LOAD_FAST
+    optimisation introduced in Python 3.14: in older versions the function call
+    itself keeps an extra reference on the caller's stack and getrefcount adds
+    its own temporary, whereas Python 3.14 steals the caller's reference and no
+    longer creates a temporary for FAST_LOCAL arguments.  See:
+    https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-refcount
+    """
+    expected = 1 if sys.version_info >= (3, 14) else 3
+    assert sys.getrefcount(t) == expected
```
GMNGeoffrey: We could have a single function that works for container subscripts as well, with  something like `def assert_sole_holder(t, *, in_container=False)`. I think that's slightly cleaner (it's possible to have Python introspect to figure this out itself, but it would be slow and gross).


```suggestion
def assert_sole_holder(t, *, in_container=False) -> None:
    """Assert that `t` is the only live Python reference to this object.

    Used in offload-inference loops to verify that a tensor has no unexpected
    references before it is moved to CPU, ensuring in-place memory reuse works
    as intended.
    
    ``in_container`` is required to indicate if the reference is a container
    subscript like ``container[i]``, which creates an additional counted
    reference on the evaluation stack, so the expected
    count differs.

    The expected count changes across Python versions because of the LOAD_FAST
    optimisation introduced in Python 3.14: in older versions the function call
    itself keeps an extra reference on the caller's stack and getrefcount adds
    its own temporary, whereas Python 3.14 steals the caller's reference and no
    longer creates a temporary for FAST_LOCAL arguments.  See:
    https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-refcount
    """
    expected = 3
    if sys.version_info >= (3, 14):
        expected = 2 if in_container else 1

    assert sys.getrefcount(t) == expected
```

GMNGeoffrey (approved):
I have a suggestion for how we could use a helper in the container subscript case as well and a nit about wording, but generally LGTM.

openfold3/core/model/utils.py
jnwei: I like @GMNGeoffrey's fix. I think it is cleaner to apply `assert_sole_holder` to all such asserts, rather than just to asserts that are not in containers.

jandom: Agreed

jnwei (approved):
LGTM thanks @jandom for the fix and @GMNGeoffrey for the suggestion re in_container

PR #245, opened 2026-06-03
refactor: migrate test_kernel.py to pytest

jandom:
**Summary**

We're working a lot on kernels this month, and one thing that's a silly blocker to some of this work is that these are the legacy unittest.TestCases

**Changes**

Mostly mechanical migration of test_kernel.py

**Related Issues**

N/A

**Testing**

Passes locally

**Other Notes**

N/A

### DIFF
--- a/openfold3/tests/test_kernels.py
+++ b/openfold3/tests/test_kernels.py
@@ -100,7 +100,7 @@ class TestKernels(unittest.TestCase):
             ).cpu()
 
         err = torch.max(torch.abs(kernel_out - real_out))
-        self.assertTrue(err < eps, f"Error: {err}")
+        assert err < eps, f"Error: {err}"
 
     @compare_utils.skip_unless_ds4s_installed()
     def test_dsk_forward_bf16(self):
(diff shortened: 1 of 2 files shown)


openfold3/tests/test_kernels.py
jnwei, line 1: At some point, could we change the kernel test error messages to be something a bit more informative? As one option, you can use the utility functions `assert_max_abs_diff_small` and `assert_mean_abs_diff_small` in `compare_utils`, which will provide a small wrapper around `torch.testing_assert_close`

https://github.com/aqlaboratory/openfold-3/blob/2172402c96ca3ed112d8ff413a31429579276e94/openfold3/tests/compare_utils.py#L72-L77

Maybe updating the error messages will be a part of later PRs, in which case, no need to update anything here.

jnwei (approved):
Looks good! Nice to migrate tests to use pytest. One small suggestion regarding error messages in thread, but it can be addressed later.

openfold3/tests/test_kernels.py
jandom: No, this will have to change – these messages are currently not super helpful, intentionally left for later

PR #248, opened 2026-06-08
Reduce peak memory usage

### LINKED ISSUE #71: [BUG] Cuda Out of Memory on large structure even low_mem is set
**Describe the bug**
low_mem setting does not work, a large structure will run into CUDA out of memory issue during predicting dataloader step.

**To Reproduce**
Our runner.yml:
```
model_update:
  presets: 
    - "predict"
    - "low_mem"  # for lower memory systems
    - "pae_enabled"
  custom:
    settings:
      memory:
        eval:
          use_deepspeed_evo_attention: true  # set this to False to use cueq only
output_writer_settings:
  structure_format: pdb
msa_computation_settings:
  server_url: http://[local server ip]:8080/api
```
command:

`run_openfold predict --query_json EVO42BM50_P19.json --num_diffusion_samples 100 --runner_yaml configurations.yml`

**Expected behavior**
Successfully run the job

**Stack trace**
```
ERROR:openfold3.projects.of3_all_atom.runner:OOM for query_id(s) BM50-P19. See logs/predict_err_rank0.log for details.
/opt/miniconda3/envs/openfold3/lib/python3.13/site-packages/pytorch_lightning/loops/prediction_loop.py:257: predict returned None if it was on purpose, ignore this warning...
Predicting DataLoader 0: 100%|██████████| 1/1 [09:10<00:00,  0.00it/s]
==================================================
    PREDICTION SUMMARY (COMPLETE)    
==================================================
Total Queries Processed: 1
  - Successful Queries:  0
  - Failed Queries:      1

Failed Queries: BM50-P19
==================================================

Predicting DataLoader 0: 100%|██████████| 1/1 [09:10<00:00,  0.00it/s]
Cleaning up MSA directories...
```

```
 File "/opt/miniconda3/envs/openfold3/lib/python3.13/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
  File "/opt/miniconda3/envs/openfold3/lib/python3.13/site-packages/openfold3/core/model/heads/head_modules.py", line 228, in forward
    zij = zij.to(device=out_device)
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 75.94 GiB. GPU 0 has a total capacity of 47.38 GiB of which 39.47 GiB is free. Including non-PyTorch memory, this process has 7.90 GiB memory in use. Of the allocated memory 6.68 GiB is allocated by PyTorch, and 787.35 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

 - OS: Ubuntu 22.04
 - a 4090 with 48g RAM

Would really appreicate the help. Don't know if it is we set the num_diffusion_samples too large.

### LINKED ISSUE #225: Inference OOM from `get_token_frame_atoms`

GMNGeoffrey:
**Summary**
This builds on https://github.com/aqlaboratory/openfold-3/pull/244 to further reduce peak memory usage across a range of input sizes. The changes are most noticeable at 3k+ residues.

I had Claude Code iteratively look for the peak memory usage point and come up with a plan to reduce it. Then I cleaned up and extracted some of its commits (some I deemed excessively invasive)

**Changes**
- Use torch.cdist in `get_token_frame_atoms` (from https://github.com/aqlaboratory/openfold-3/pull/244)
- Use inplace bool masks in `get_token_frame_atoms` (from https://github.com/aqlaboratory/openfold-3/pull/244)
- Under offload, move zij to GPU in slices for PAE/PDE computation (cause of the OOM reported in https://github.com/aqlaboratory/openfold-3/issues/71)
- Under offload, defer moving the PAE/PDE/distogram results to the GPU until they are actually consumed
- Use in-place adds to fold in linear projections in template_embedders
- Chunk the diffusion_conditioning embedding

**Related Issues**
- Fixes https://github.com/aqlaboratory/openfold-3/issues/225. My report about an OOM that motivated the cdist change.
- Fixes https://github.com/aqlaboratory/openfold-3/issues/71. User report about

**Testing**
- I've done e2e runs showing no measurable impact on latency and identical outputs as well as graphed the memory usage.

**Other Notes**
I'm bundling these because the effects are sort of interdependent (e.g. one change might not reduce peak memory usage because there's another binding peak) and because thoroughly testing them is a bit time consuming, so I think it might be more efficient to show that the whole stack doesn't affect outputs or latency while reducing memory usage rather than doing it for each one individually. But I can break this up. A draft PR currently because I wanted something to point to.

Some of this might also change if we get rid of `apply_per_sample` as tentatively discussed in https://github.com/aqlaboratory/openfold-3/pull/213.

### DIFF
--- a/openfold3/core/model/feature_embedders/template_embedders.py
+++ b/openfold3/core/model/feature_embedders/template_embedders.py
@@ -23,6 +23,7 @@ from ml_collections import ConfigDict
 
 import openfold3.core.config.default_linear_init_config as lin_init
 from openfold3.core.model.primitives import LayerNorm, Linear
+from openfold3.core.utils.tensor_utils import add
 
 
 class TemplatePairEmbedderAllAtom(nn.Module):
@@ -67,7 +68,7 @@ class TemplatePairEmbedderAllAtom(nn.Module):
         self.layer_norm_z = LayerNorm(c_in)
         self.linear_z = Linear(c_in, c_out, **linear_init_params.linear_z)
 
-    def _embed_feats(self, batch: dict):
+    def _embed_feats(self, batch: dict, inplace_safe: bool = False):
         dtype = batch["template_unit_vector"].dtype
 
         # [*, N_token, N_token]
@@ -102,24 +103,30 @@ class TemplatePairEmbedderAllAtom(nn.Module):
             *template_restype.shape[:-2], n_token, -1, -1
         )
 
-        # Build the accumulator from the first projection, then fold each
-        # subsequent linear in-place. Out of place adds (a = a + linear(x)) keep
-        # a, linear(x), and a+linear(x) all live during the addition; that's
-        # three [*, N_templ, N, N, C] tensors co-resident. add_ frees the new
-        # sum's separate tensor (it's written into a) so peak per step drops
-        # from 3x slice to ~2x slice.
         a = self.dgram_linear(template_distogram)
-        a.add_(self.pseudo_beta_mask_linear(pseudo_beta_pair_mask))
-        a.add_(self.aatype_linear_1(template_restype_ti.to(dtype=dtype)))
-        a.add_(self.aatype_linear_2(template_restype_tj.to(dtype=dtype)))
-        a.add_(self.x_linear(x[..., None]))
-        a.add_(self.y_linear(y[..., None]))
-        a.add_(self.z_linear(z[..., None]))
-        a.add_(self.backbone_mask_linear(backbone_frame_pair_mask))
+        a = add(
+            a, self.pseudo_beta_mask_linear(pseudo_beta_pair_mask), inplace=inplace_safe
+        )
+        a = add(
+            a,
+            self.aatype_linear_1(template_restype_ti.to(dtype=dtype)),
+            inplace=inplace_safe,
+        )
+        a = add(
+            a,
+            self.aatype_linear_2(template_restype_tj.to(dtype=dtype)),
+            inplace=inplace_safe,
+        )
+        a = add(a, self.x_linear(x[..., None]), inplace=inplace_safe)
+        a = add(a, self.y_linear(y[..., None]), inplace=inplace_safe)
+        a = add(a, self.z_linear(z[..., None]), inplace=inplace_safe)
+        a = add(
+            a, self.backbone_mask_linear(backbone_frame_pair_mask), inplace=inplace_safe
+        )
 
         return a
 
-    def forward(self, batch, z):
+    def forward(self, batch, z, inplace_safe: bool = False):
         """
         Args:
             batch:
@@ -129,7 +136,7 @@ class TemplatePairEmbedderAllAtom(nn.Module):
         Returns:
             # [*, N_templ, N_token, N_token, C_out] Template pair feature embedding
         """
-        a = self._embed_feats(batch=batch)
+        a = self._embed_feats(batch=batch, inplace_safe=inplace_safe)
 
         # [*, N_templ, N_token, N_token, C_out]
         z = self.linear_z(self.layer_norm_z(z))
--- a/openfold3/core/model/feature_embedders/template_embedders.py
+++ b/openfold3/core/model/feature_embedders/template_embedders.py
@@ -102,14 +102,20 @@ class TemplatePairEmbedderAllAtom(nn.Module):
             *template_restype.shape[:-2], n_token, -1, -1
         )
 
+        # Build the accumulator from the first projection, then fold each
+        # subsequent linear in-place. Out of place adds (a = a + linear(x)) keep
+        # a, linear(x), and a+linear(x) all live during the addition; that's
+        # three [*, N_templ, N, N, C] tensors co-resident. add_ frees the new
+        # sum's separate tensor (it's written into a) so peak per step drops
+        # from 3x slice to ~2x slice.
         a = self.dgram_linear(template_distogram)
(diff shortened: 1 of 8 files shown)


GMNGeoffrey:
Here's a plot showing how this changes peak memory usage

image

Note that the changes especially with per-slice zij moves do introduce more memory fragmentation, which is why the end of the series with these changes is significantly below the full memory capacity

openfold3/core/model/feature_embedders/template_embedders.py
```diff
@@ -102,14 +102,20 @@ def _embed_feats(self, batch: dict):
...
+        # from 3x slice to ~2x slice.
         a = self.dgram_linear(template_distogram)
-        a = a + self.pseudo_beta_mask_linear(pseudo_beta_pair_mask)
-        a = a + self.aatype_linear_1(template_restype_ti.to(dtype=dtype))
-        a = a + self.aatype_linear_2(template_restype_tj.to(dtype=dtype))
-        a = a + self.x_linear(x[..., None])
-        a = a + self.y_linear(y[..., None])
-        a = a + self.z_linear(z[..., None])
-        a = a + self.backbone_mask_linear(backbone_frame_pair_mask)
+        a.add_(self.pseudo_beta_mask_linear(pseudo_beta_pair_mask))
```
christinaflo: this should only run if inplace_safe=True, similar to other places in the code. we have a function
`from openfold3.core.utils.tensor_utils import add` that passes the flag like `a = add(a, self.pseudo_beta_mask_linear(pseudo_beta_pair_mask), inplace=inplace_safe)`. it's always true for inference, not training

GMNGeoffrey: Ah sorry, I didn't look closely enough at how inplace_safe was being used. I think these particular inplace adds are safe and that test_of3_model.py would catch if they weren't. But it looks like the convention in the project is to guard all inplace operations within the model proper behind `inplace_safe` to avoid having to reason about each individual usage site. Is that right? I don't think the usage is here is special enough to be worth breaking that if so. But there could be potential memory savings during training by relaxing this constraint (and I think it's something that would be caught by unit tests, so not super dangerous). That's dependent on what the particular layers hang on to for the backward pass, but at least with linear layers there should be a training-time memory savings.

This would potentially affect the `masked_fill_` I used as well. I think that one is safe by construction given that we're only using it to compute topk indices, which we can't differentiate through anyways (could actually wrap it in `torch.no_grad()`, I think).

I'll go ahead and change this to plumb through the flag and use the utility

PR #249, opened 2026-06-09
Add callback for profiling GPU memory usage

GMNGeoffrey:
**Summary**
It's useful to be able to keep track of where GPU memory is being used in the model and where the peaks are so we can try to fit larger inputs (or on smaller GPUs).

**Changes**
- Adds an optional callback that tracks GPU memory usage with `torch.cuda.memory` utilities. This outputs a pkl dump of all allocations over time as well as logging the peak memory usage.

**Testing**
- Basic unit test matching the ones for other experiment runner config settings.

**Other Notes**
My editor was stripping trailing whitespace from the files I touched and introducing distracting diffs. Rather than fight with it, I pre-factored that change, so you can just view the last two commits for the substantive change (or use the view that hides whitespace diffs). I can revert that if you don't like formatting changes of untouched lines mixed into PRs though.

### DIFF
--- a/examples/reference_full_config/full_config.yml
+++ b/examples/reference_full_config/full_config.yml
@@ -20,6 +20,7 @@ experiment_settings:
   use_msa_server: true
   use_templates: true
   skip_existing: false
+  record_memory_snapshot: false
 
 # PLTrainerArgs: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/entry_points/validator.py#L121
 # Configuration settings for pytorch lightning
--- a/examples/reference_full_config/full_config.yml
+++ b/examples/reference_full_config/full_config.yml
@@ -3,16 +3,16 @@
 # Most experiments do not require a runner.yml
 # When a runner.yml is used, it is not required to specify all configuration settings.
 
-# For typical usage, please refer to the example runner yamls in examples/example_runner_yamls/, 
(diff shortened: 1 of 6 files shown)


GMNGeoffrey:
@christinaflo Sorry, I think I misremembered and thought your comment in https://github.com/aqlaboratory/openfold-3/pull/227 said I should share my version, but you actually said you already had one 😬 If this isn't helpful or you like yours better, we can just close this 😄 or if you want to share the version you've got, I'm happy to see if there's any meaningful diff between them and clean up or test

christinaflo:
Yeah i just wanted to avoid the conflicts for when mine eventually gets merged in, but i can share mine and you can add some edits on top of it? mine has some extra functionality id like to keep

jnwei (commented):
This is a nice Callback to have, especially for profiling! 

One nit: In general, we try to limit the available configuration settings available from the command line to only those which we expect the user to need frequently (e.g. templates, colabfold settings).

Since most users will probably not use memory profiling with their inference run, can we keep the settings configurable only through the `experiment_settings` header in the runner_yaml?

jandom:
Very handy!

christinaflo:
@GMNGeoffrey Ill tag you in the PR later today, mine was mostly training focused so I think we can add in the additions on top of the general snapshot that you have in this PR

PR #251, opened 2026-06-09
Profiling tools

christinaflo:
Added general profiling stuff and other utils. Includes:

1. SecondsPerIterationProgressBar() callback to flip the progress bar to seconds/iteration
2. MemorySnapshot() callback to create the torch pkl mem dumps. Also has an oom_observer to save a snapshot of memory right after OOM (useful for backward pass where error message is ambiguous about location)
3. PredictTimer(): Add iteration time step logging for training. More lightweight than profiling to see spikes/stalls in iteration time at scale and over time.
4. ProfilerConfig to add torch profiler to lightning module when enabled

@GMNGeoffrey This is a copy of my internal PR, it has a some extra stuff outside of the memory snapshot callback. LMK what changes you'd like to add from your PR.

### DIFF
--- a/examples/example_runner_yamls/profiling.yml
+++ b/examples/example_runner_yamls/profiling.yml
@@ -1,0 +1,21 @@
+profiler:
+  enabled: true
+  # Default values:
+  dirpath: logs/profiler_traces
+  filename: profile_trace
+  skip_first: 0
+  wait: 1
+  warmup: 1
+  active: 3
+  repeat: 0
+  record_shapes: true
+  profile_memory: false
+  with_stack: true
+
+memory_snapshot:
+  enabled: true
+  # Default values:
+  output_path: memory_snapshot.pickle
+  start_step: 0
+  dump_on_oom: false
+  stacks: null  # Defaults to 'all' for x86_64, 'python' for ARM
--- a/examples/example_runner_yamls/profiling.yml
+++ b/examples/example_runner_yamls/profiling.yml
@@ -16,6 +16,6 @@ memory_snapshot:
   enabled: true
   # Default values:
   output_path: memory_snapshot.pickle
-  start_step: 0
+  step: 0
   dump_on_oom: false
   stacks: null  # Defaults to 'all' for x86_64, 'python' for ARM
--- a/openfold3/core/utils/callbacks.py
+++ b/openfold3/core/utils/callbacks.py
@@ -29,27 +29,188 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
+class SecondsPerIterationProgressBar(pl.callbacks.TQDMProgressBar):
+    """Progress bar that displays s/it instead of it/s.
+
+    Lightning hardcodes {rate_noinv_fmt} in BAR_FORMAT which forces it/s.
+    Overriding with {rate_inv_fmt} to always show s/it.
+    """
+
+    BAR_FORMAT = (
+        "{l_bar}{bar}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, "
+        "{rate_inv_fmt}{postfix}]"
+    )
+
+
+class MemorySnapshot(pl.Callback):
+    """
+    Records memory history and dumps a snapshot pickle.
+    The pickle can be visualized at https://pytorch.org/memory_viz
+
+    Args:
+        output_path: Path to save the .pickle snapshot file.
+        start_step: Step (batch_idx) at which to start recording and dump a single-step
+            snapshot. Set to `None` to disable step-based snapshotting (useful
+            when only dump_on_oom is needed). Defaults to 0.
+        dump_on_oom: If True, attaches an OOM observer that dumps the snapshot
+            on out-of-memory. History is reset each step to keep the snapshot
+            small.
+        stacks: Stack trace mode for record_memory_history. Defaults to "all".
+    """
+
+    def __init__(
+        self,
+        output_path: str = "memory_snapshot.pickle",
+        start_step: int | None = 0,
+        dump_on_oom: bool = False,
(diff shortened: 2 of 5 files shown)


GMNGeoffrey (commented):
Nice, obviously a lot more features here :-)

I think potential limitations vs my implementation are:

- limited to single step snapshot with global file path
- Registered before the predict timer, so I think it would put the memory snapshotting inside of the timer's timing. IMO that isn't ideal since it means you have to do separate runs to see latency and memory. In my runs I found that the memory tracking otherwise added minimal overhead, so it was nice to capture both at once.

Both of these are tradeoffs/judgement calls.

We could switch to the file per-step I have, although I see why especially in training that might be undesirable. My main use case was running a single batch anyways: I just made it per batch because it seemed natural for how it was constructed. So I think it's fine to leave it single-step only. The one change I would make is renaming start_step because that strongly suggests to me that subsequent steps will *also* get recorded and I don't think that's true.

Switching the order of the callbacks getting registered wouldn't be hard (add superclass callbacks after the subclass ones), though it is a bit awkward because it means having to reason about what gets added when across the hierarchy. I think you could also argue for putting the output writing outside of the predict step timing, depending on what exactly you're interested in timing, but that is a more core operation, so including it makes sense as well.

I can propose changes if that sounds good to you.

christinaflo:
> Nice, obviously a lot more features here :-)
> 
> I think potential limitations vs my implementation are:
> 
> * limited to single step snapshot with global file path
> * Registered before the predict timer, so I think it would put the memory snapshotting inside of the timer's timing. IMO that isn't ideal since it means you have to do separate runs to see latency and memory. In my runs I found that the memory tracking otherwise added minimal overhead, so it was nice to capture both at once.
> 
> Both of these are tradeoffs/judgement calls.
> 
> We could switch to the file per-step I have, although I see why especially in training that might be undesirable. My main use case was running a single batch anyways: I just made it per batch because it seemed natural for how it was constructed. So I think it's fine to leave it single-step only. The one change I would make is renaming start_step because that strongly suggests to me that subsequent steps will _also_ get recorded and I don't think that's true.
> 
> Switching the order of the callbacks getting registered wouldn't be hard (add superclass callbacks after the subclass ones), though it is a bit awkward because it means having to reason about what gets added when across the hierarchy. I think you could also argue for putting the output writing outside of the predict step timing, depending on what exactly you're interested in timing, but that is a more core operation, so including it makes sense as well.
> 
> I can propose changes if that sounds good to you.

Yeah we can rename start_step, I originally had start and end to visualize memory over 2-3 steps but those files get big and slow and I also generally do single batch anyway so I took it out. I left this as a var so you can leave some warm up steps. But yeah I agree that doing it per batch makes more sense for inference. Same with the callback order, we could get rid of the hierarchy and just have the desired order in the training/inference callback sections if that's easier to read.

Yes feel free to propose changes!

GMNGeoffrey:
Ah if you also didn't have a specific use for running this with multiple batches and warmup steps and such, then I think "just do it every batch" simplifies the API a bit. Is there a realistic case where you *don't* want to record ooms? And then if it's per batch it's somewhat natural to put the dump in the batch output directory (but maybe this is just my aversion to scripts putting stuff in the working directory 😁). I think in that case the API could just be an enable switch like in my pr (except no command line flag, per Jennifer's comment) plus the stacks option (assuming that was an option you found it useful to modify): every batch gets a snapshot that gets recorded and dumped on batch end or on oom at `<output_dir>/mem_snapshot.pkl` (can also keep the full `.pickle` extension and the `_oom` suffix in the latter case if you prefer). If we wanted more control, I'd be inclined to mostly mirror the torch profiler options, but I would say yagni on that right now. How does that sound? I'll send a PR against your branch if that sounds good. To be clear, I'm happy to defer to you on any of this: this pr pretty much satisfies what I was trying to accomplish; I'm just trying to suggest something helpful for the project and be specific enough that you can say yes or no 🙂

christinaflo:
> Ah if you also didn't have a specific use for running this with multiple batches and warmup steps and such, then I think "just do it every batch" simplifies the API a bit. Is there a realistic case where you _don't_ want to record ooms? And then if it's per batch it's somewhat natural to put the dump in the batch output directory (but maybe this is just my aversion to scripts putting stuff in the working directory 😁). I think in that case the API could just be an enable switch like in my pr (except no command line flag, per Jennifer's comment) plus the stacks option (assuming that was an option you found it useful to modify): every batch gets a snapshot that gets recorded and dumped on batch end or on oom at `<output_dir>/mem_snapshot.pkl` (can also keep the full `.pickle` extension and the `_oom` suffix in the latter case if you prefer). If we wanted more control, I'd be inclined to mostly mirror the torch profiler options, but I would say yagni on that right now. How does that sound? I'll send a PR against your branch if that sounds good. To be clear, I'm happy to defer to you on any of this: this pr pretty much satisfies what I was trying to accomplish; I'm just trying to suggest something helpful for the project and be specific enough that you can say yes or no 🙂

For training I just want to be able to take a snapshot at least a step in so that I get the optimizer states also, there's never really a need for me to have every batch for that case. And I generally wouldnt run this with oom dump unless it came up within a a relatively small amount of steps for training and I was debugging. So as long as I can run that without generating a bunch of files I'm happy :). 
re: stack option, yes i need this for gh200s

GMNGeoffrey:
Ah, sounds like you do need the features you have right now then. Regarding running with OOM dump, I didn't mean that you would always run with OOM dump in production. Rather that if you were already doing memory snapshotting of a step, it's unlikely that OOM dump on the other steps would hurt. But I think what you've got here is good and has features you need, so I'll stop bikeshedding :-)

jandom:
Confirmed with @christinaflo that this is still of interest but dropping to draft

PR #252, opened 2026-06-09
fix: use-msa-server and use-templates defaults

### LINKED ISSUE #250: [Bug] Inference yaml use_templates = False overwritten unexpectedly
**Problem**
Inference runs with use_templates = True in cases where the user intends to set use_templates = False. 
- If the user sets use_templates = False in their yaml and passes no arguments to predict, the behavior is to use templates in inference. 
- Additionally, while most CLI argument overwrite the yaml, but if a user explicitly passes use_templates = False to predict, this does not overwrite a yaml setting for use_templates = True.

**Current code logic**
The config default is use_templates = False and a user can also set that explicitly in their yaml.

However, the default for predict in run_openfold.py is True https://github.com/aqlaboratory/openfold-3/blob/792fa5df46dab62cf3e197cd7488dfad8b9abb04/openfold3/run_openfold.py#L135-L141 (the help description is also inaccurate -- this sets whether templates are used for prediction not whether MSA server is used to align templates).

Downstream, in the experiment runner, the yaml/config value is only overwritten if use_templates = True https://github.com/aqlaboratory/openfold-3/blob/792fa5df46dab62cf3e197cd7488dfad8b9abb04/openfold3/entry_points/experiment_runner.py#L628-L630

**Recommendation**
I would recommend changing the default arg for predict use_templates=None and overwrite the yaml if it's explicitly set to True or False. If the default behavior for inference should be True, that could be set in the config preset 'predict'.

jandom:
**Summary**

Fixes a bug where the inference `use_templates` (and `use_msa_server`) yaml setting was silently overwritten. Previously, a user setting `use_templates: false` in their runner yaml would still run inference with templates enabled, and an explicit `--use_templates False` on the CLI did not override a yaml `use_templates: true`.

The root cause was a two-part mismatch:
1. The CLI options defaulted to `True`, so even when the user passed nothing, `True` was handed to the runner.
2. The runner only applied the override when the value was truthy (`if use_templates:`), so it (a) clobbered a yaml `False` with the `True` CLI default and (b) ignored an explicit CLI `False`.

The flags are now tri-state: `None` means "not specified" (the yaml / config value is used), while an explicit `True` or `False` overrides it. To preserve today's behavior when nothing is specified, the inference config defaults are set to `True` (templates + MSA server on).

**Changes**

- `openfold3/run_openfold.py` — `--use-msa-server` and `--use-templates` now default to `None` instead of `True`; updated the `predict()` signature accordingly. Also corrected the inaccurate `--use-templates` help text (it previously described MSA-server template alignment; it actually controls whether templates are used for prediction).
- `openfold3/entry_points/experiment_runner.py` — `InferenceExperimentRunner.__init__` and `update_config_with_cli_args` default `use_msa_server` / `use_templates` to `None`; the override now uses `if x is not None:` and assigns the actual value, so explicit `True` and `False` both override the yaml while `None` leaves it untouched.
- `openfold3/entry_points/validator.py` — `InferenceExperimentSettings` defaults for `use_msa_server` / `use_templates` changed from `False` to `True` to preserve the current "on" behavior when neither yaml nor CLI specifies a value. This class is used only by the `predict` path (training/validation use `TrainingExperimentSettings`), so the change is scoped to inference.
- `openfold3/tests/test_entry_points.py` — replaced the two-state CLI tests with a single readable, fully-parametrized resolution matrix.

**Resulting behavior**

| yaml setting | CLI arg | result |
|---|---|---|
| unset | omitted | `True` (unchanged from today) |
| `False` | omitted | `False` (bug fixed) |
| `True` | `--use_templates False` | `False` (bug fixed) |
| `False` | `--use_templates True` | `True` |

**Related Issues**

Closes #250

**Testing**

Added `TestInferenceCommandLineSettings::test_cli_and_yaml_resolution`, a 3×3 matrix (CLI omitted / `True` / `False` × yaml unset / `True` / `False`) parametrized over both `use_templates` and `use_msa_server`, with named `pytest.param` ids and hardcoded expected values. The two issue-#250 regression rows are tagged `BUG#250` in their ids.

```
.pixi/envs/openfold3-cuda13-pypi/bin/pytest \
  openfold3/tests/test_entry_points.py::TestInferenceCommandLineSettings -v
```

All 18 resolution cases pass (plus the existing seeding tests in the class). Also verified `run_openfold predict --help` shows the corrected help text and no longer forces a `True` default.

**Other Notes**

This changes the no-input default source from the CLI to the config (both still resolve to "on"), so behavior is unchanged for users who pass nothing. Users who explicitly set `use_templates: false` in their yaml will now correctly get template-free inference.

### DIFF
--- a/openfold3/tests/test_entry_points.py
+++ b/openfold3/tests/test_entry_points.py
@@ -18,6 +18,7 @@ import shutil
 import tempfile
 import textwrap
 import unittest
+from dataclasses import dataclass
 from pathlib import Path
 from unittest.mock import patch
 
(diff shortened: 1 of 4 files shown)


openfold3/tests/test_entry_points.py
```diff
@@ -425,22 +426,87 @@ def test_store_configs_creates_files(self, mock_os_system):
...
+
+    yaml: bool | None
+    cli: bool | None
+    expected: bool
+
+
+# Resolution rule: the CLI arg wins when provided; otherwise the yaml value
+# wins; otherwise the config default (True). The two cases marked "BUG #250"
+# are the regressions this matrix guards against.
+_FLAG_RESOLUTION_CASES = [
```
jnwei, line 446: Nice! I like this way of organizing the test cases

jnwei (approved):
LGTM! thanks for the quick fix on this @jandom

openfold3/tests/test_entry_points.py
jandom: I know, Claude really doesn't like doing this :P

ECalfeeAdaptive:
Thanks for the super fast fix!

PR #256, opened 2026-06-11
fix: fixed flaky test, test_template_module_offload

tytrdev:
**Summary**

Fixed a flaky test, `test_template_module_offload`. This test was failing on my machine and would pass only intermittently. This makes two small changes to get deterministic behavior and allow for some numerical tolerance in assertions.

**Changes**

- Uses the existing `seeded_rng` test fixture for determinism
- Compare with a realistic tolerance `atol=1e-5, rtol=1e-4` to absorb rounding difference while still catching any real regression

**Related Issues**

N/A

**Testing**

Verified the flakiness is gone with a before/after run, 20 iterations each.
Pre-fix version passed `9/20` runs, fix passes `20/20`.

<details>
  <summary>Before vs. after, 20 consecutive runs each</summary>

  > T="openfold3/tests/test_embedders.py::TestTemplateEmbedders::test_template_module_offload"
  
  # BEFORE (parent commit: bare allclose, unseeded weights)
  > git checkout HEAD~1 -- openfold3/tests/test_embedders.py
  > for i in $(seq 1 20); do pytest "$T" -q -p no:cacheprovider; done
  BEFORE: 9 passed, 11 failed (out of 20)
  
  # AFTER (this fix: seeded_rng + tolerance)
  > git checkout HEAD -- openfold3/tests/test_embedders.py
  > for i in $(seq 1 20); do pytest "$T" -q -p no:cacheprovider; done
  AFTER: 20 passed, 0 failed (out of 20)
  </details>
  
  <details>
  <summary>Single run (verbose) + full module under xdist</summary>

  $ pytest openfold3/tests/test_embedders.py::TestTemplateEmbedders::test_template_module_offload -v
  ============================= test session starts ==============================
  platform linux -- Python 3.14.5, pytest-9.0.3, pluggy-1.6.0
  collected 1 item
  openfold3/tests/test_embedders.py::TestTemplateEmbedders::test_template_module_offload PASSED [100%]
  ======================== 1 passed, 15 warnings in 3.13s ========================

  $ pytest openfold3/tests/test_embedders.py -n auto
  ============================= test session starts ==============================
  platform linux -- Python 3.14.5, pytest-9.0.3, pluggy-1.6.0
  16 workers [9 items]
  ======================= 9 passed, 254 warnings in 27.15s =======================
  </details>

**Other Notes**

Seeding fixes the weight draw, so the test now checks one representative config rather than re-sampling each run. I confirmed the offload vs batched difference is stable across different seeds. That said, using `seeded_rng` here may have consequences that are non-obvious to me.

### DIFF
--- a/openfold3/tests/test_embedders.py
+++ b/openfold3/tests/test_embedders.py
@@ -189,7 +189,7 @@ class TestTemplateEmbedders:
 
         assert emb.shape == (batch_size, n_templ, n_token, n_token, c_t)
 
-    def test_template_module_offload(self, template_batch):
+    def test_template_module_offload(self, template_batch, seeded_rng):
         batch_size = template_batch["batch_size"]
         n_token = template_batch["n_token"]
         batch = template_batch["batch"]
@@ -213,7 +213,8 @@ class TestTemplateEmbedders:
                 batch=batch, z=z, pair_mask=pair_mask, offload_inference=True
             )
 
-        assert torch.allclose(t_no_offload, t_offload)
+        # Batched vs. per-template offload paths round differently; compare within tolerance
+        assert torch.allclose(t_no_offload, t_offload, atol=1e-5, rtol=1e-4)
 
 
 class TestMSAModuleStack:


openfold3/tests/test_embedders.py
```diff
@@ -213,7 +213,8 @@ def test_template_module_offload(self, template_batch):
                 batch=batch, z=z, pair_mask=pair_mask, offload_inference=True
             )
 
-        assert torch.allclose(t_no_offload, t_offload)
+        # Batched vs. per-template offload paths round differently; compare within tolerance
+        assert torch.allclose(t_no_offload, t_offload, atol=1e-5, rtol=1e-4)
```
jnwei, line 217: confirming that atol=1e-5 is within reason. atol is set to 5e-4 for similar tests.

https://github.com/aqlaboratory/openfold-3/blob/fc9b7394c402803f919c38ee3cce55f77c7dda2e/openfold3/tests/config.py#L30

jnwei (approved):
LGTM, thanks for the fix @tytrdev !

PR #257, opened 2026-06-12
Change CI cronjob timing to time with (hopefully) less traffic 

jnwei:
**Changes**

We observe a high failure rate on our daily GPU CI tests. The error is always due to unavailability of GPU resources on the instance that we've been granted, possibly due to noisy neighbors.

This PR makes two changes:
- ~Adds a simple smoke test (`nvidia-smi`) on the CI runners to ensure that nvidia-smi is available before running the tests.~
   - This smoke test was not effective in catching an unavailable GPU early. A heavier smoke test is unlikely to be more efficient than the current time to failure of 4-5 min, so I am removing the smoke test.

- Changes the cronjob run time to be 3:17am rather than 3:00am, in hopes of finding a lighter traffic period to run the tests.

**Related Issues**
Recording of CI tests failures is available here: https://github.com/aqlaboratory/openfold-3/blob/claude/ecstatic-faraday-kdx091/ci-monitoring/aws-outage-failures.md

As of June 6, on 6 / 7 days have failed with AWS job busy errors. The true error rate is somewhat obscured because we have 3 parallel runs (conda, pixi cuda 12, pixi cuda 13), so to get the actual failure rate, we'd have to start recording the instances with failures as well.

**Testing**
- [x] Trigger a CI testing run manually and verify tests pass.

**Other Notes**

### DIFF
--- a/.github/workflows/ci-integration-test-conda-reusable.yml
+++ b/.github/workflows/ci-integration-test-conda-reusable.yml
@@ -12,6 +12,11 @@ on:
         description: 'CUDA base image tag (e.g., 12.2.2-cudnn8-devel-ubuntu22.04)'
         required: true
         type: string
+      gpu_smoke_image:
+        description: 'Small CUDA base image used by the pre-build GPU smoke check (only needs nvidia-smi).'
+        required: false
+        type: string
+        default: 'nvidia/cuda:12.4.1-base-ubuntu22.04'
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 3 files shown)


jandom (approved):

jnwei:
Hm it looks like we'll have to try running these tests at a less busy time, I couldn't schedule the test: https://github.com/aqlaboratory/openfold-3/actions/runs/27404204546

PR #259, opened 2026-06-12
MemorySnapshot: rename step param, register after PredictTimer

GMNGeoffrey:
**Summary**
A couple of edits to the MemorySnapshot callback in #251 (this is deliberately against that branch, not the main branch). Rename `start_step` for clarity and run memory tracking outside of the timed part of the step.

**Changes**
- Rename `start_step` -> `step`. The previous name implied later steps would also be recorded, which is not the case; only the single named step is dumped.
- Flatten callback registration: drop the base-class MemorySnapshot construction and append it inline in each runner after PredictTimer. This keeps the memory snapshot dump outside the predict timer's measurement window so a single run can capture both memory usage and a relatively-close-to-production latency.

### DIFF
--- a/examples/example_runner_yamls/profiling.yml
+++ b/examples/example_runner_yamls/profiling.yml
@@ -16,6 +16,6 @@ memory_snapshot:
   enabled: true
   # Default values:
   output_path: memory_snapshot.pickle
-  start_step: 0
+  step: 0
   dump_on_oom: false
   stacks: null  # Defaults to 'all' for x86_64, 'python' for ARM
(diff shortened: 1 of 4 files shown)


GMNGeoffrey:
@christinaflo PTAL. Just the basic changes we discussed for #251.

christinaflo:
LGTM! Ah I cant make up my mind, I liked your inference per prediction mem dump idea (and saving it to the prediction folder), just not for training. Idk if there's a way to marry the two. I guess it's really not the biggest deal to just do the first step for training. Anyway we can change it later on if we end up needing it.

GMNGeoffrey:
Yeah I liked that better for inference too, but decided it was better to stop bike shedding, since I know you all have limited bandwidth 😁

PR #255, opened 2026-06-11
Add CIF template-based generating example notebook 

sukritsingh:
**Summary**

Following up on conversations with @jandom and @jnwei, this PR provides an example notebook intended to install and run OpenFold3 given a template CIF structure of a kinase (the underlying intention is to generate template-based structures of homologs that may be under-characterized). Ideally, this example notebook can serve two purposes: 
1. How to setup and run OpenFold3 in a colab notebook (although you need a Google subscription to access the needed A100 GPU).
2. How to generate a templated structure using OpenFold3 given a desired conformational template CIF file (using kinases as an example).

The associated citation for this notebook is a preprint that should be released eminently. Perhaps we can coordinate merging of this notebook in once the DOI is added into the notebook? 

**Changes**

In the `examples/` directory there is now a new notebook `template_kinase_example_notebook` that contains the example `.ipynb` file! 

**Related Issues**

Two potential flags here:

1. The comments might be a bit verbose and chaotic as I wrote out issues I found and debugged them! Definitely open to streamlining them!
2. The only way to run this notebook on Colab is using an A100 instance, which is only available if you have a paid Google subscription that accesses an A100 (and also uses a Triton kernel)

**Testing**

You could just open this notebook in Colab and run it. I have not added a "Open in Colab" badge yet at the top of the notebook, but that might be the last thing to add (alongside the relevant DOI?)

That said, any other edits and suggested reviews are welcome!

PR #264, opened 2026-06-17
Create '--non-interactive' option for setup_openfold

jnwei:
**Summary**
`setup_openfold` is currently an interactive script that asks 4 questions regarding where to download the model parameters, which model parameter(s) to download, and whether to run integration tests.

For automated workflows, it is convenient to provide a "--non-interactive" option for the script such that all default options are automatically selected. One immediate use case is in TestSetupOpenfold (in test_entry_points.py), where we need to provide fake values for setup_openfold.main(). Another use case is in example colabfold notebooks, such as #255 

**Changes**
- Refactor setup_openfold to use a `OpenFoldSetupConfig` pydantic model to record options, and store default settings. Interactive mode now populates OpenFoldSetupConfig with user input through the same set of queries provided previously.
- Updates setup_openfold to include `--non-interactive` option, which will use the default settings of `OpenFoldSetupConfig`
- Updates `test_entry_points.TestSetupOpenFold` with additional test to test default settings and incorporate cli_runner.
- [x] update documentation

**Related Issues**
#255 

**Testing**
- [x] `test_entry_points.py` tests pass
- [x] `test_inference_full.py` passes
- [x] Manually run setup_openfold on existing .openfold3 cache and new cache directory

### DIFF
--- a/openfold3/setup_openfold.py
+++ b/openfold3/setup_openfold.py
@@ -22,12 +22,16 @@ import logging
 import os
 import sys
 from pathlib import Path
+from typing import Literal
 
 import biotite.setup_ccd
 import click
+import pydantic
+from pydantic import Field
 
 from openfold3.core.utils.s3 import download_s3_file, s3_file_matches_local
 from openfold3.entry_points.parameters import (
+    CHECKPOINT_ROOT_FILENAME,
     DEFAULT_CHECKPOINT_NAME,
     LEGACY_CHECKPOINTS,
     OPENFOLD_MODEL_CHECKPOINT_REGISTRY,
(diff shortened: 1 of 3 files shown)


openfold3/tests/test_entry_points.py
```diff
@@ -735,9 +736,30 @@ def test_remove_duplicates(self, dummy_ckpt_file, dummy_output_path):
 
 
 class TestSetupOpenFold:
+    def test_run_with_defaults(self, tmp_path, monkeypatch):
+        monkeypatch.delenv("OPENFOLD_CACHE", raising=False)
+        monkeypatch.setenv("HOME", str(tmp_path))
```
jandom: nit: I generally don't try to mix different testing apis/styles, so either monkey patch or patch but not both

openfold3/setup_openfold.py
```diff
@@ -231,23 +240,38 @@ def run_integration_tests() -> None:
     logger.info("Integration tests passed!")
 
 
-def main():
-    """Main execution."""
+@click.command()
+@click.option(
+    "--run-with-defaults",
```
jandom: a more typical flag would be '--quiet' or '--non-interactive'

openfold3/setup_openfold.py
```diff
@@ -231,23 +240,38 @@ def run_integration_tests() -> None:
...
     if should_download:
-        download_parameters(param_dir)
+        download_parameters(param_dir, run_with_defaults=run_with_defaults)
 
     # Step 4: Setup CCD with biotite
     setup_biotite_ccd(ccd_path=biotite.setup_ccd.OUTPUT_CCD, force_download=False)
 
     # Step 5: Run tests (always run regardless of download status)
-    run_integration_tests()
+    run_integration_tests(run_with_defaults=run_with_defaults)
```
jandom: This is perhaps the most structural comment (and serious) rather than replicating the boolean flag (these tend to multiply) we pull this out and create a config object, if it's interactive, it asks user how to configure, otherwise it constructs a default one, and we just pass the config object down?

Will break if we need to do a lot of checking if things exist/don't exist when constructing the config

jandom (approved):

openfold3/tests/test_entry_points.py
jnwei: replaced with patch

openfold3/setup_openfold.py
jnwei: replaced with `--non-interactive`

openfold3/setup_openfold.py
jnwei: PTAL, based on this comment, I made a heavier refactor of setup_openfold. I think it will be worth it / make the options more clear.

openfold3/tests/test_entry_points.py
```diff
@@ -735,56 +737,107 @@ def test_remove_duplicates(self, dummy_ckpt_file, dummy_output_path):
...
-            ]
+    def test_non_interactive(self, tmp_path):
+        env_patch = patch.dict(os.environ, {"HOME": str(tmp_path)}, clear=False)
+        s3_patch = patch(
+            "openfold3.setup_openfold.download_s3_file",
+            side_effect=_fake_download_s3_file,
         )
+        with env_patch, s3_patch:
+            os.environ.pop("OPENFOLD_CACHE", None)
+            result = CliRunner().invoke(setup_openfold.main, ["--non-interactive"])
```
jandom, line 748: Awesome

openfold3/setup_openfold.py
```diff
@@ -222,32 +224,69 @@ def run_integration_tests() -> None:
...
+    help="Path to a JSON file containing an OpenFoldSetupConfig.",
+)
+def main(non_interactive: bool = False, config_path: Path | None = None):
+    """Set up OpenFold3 model parameters."""
+    if config_path is not None:
+        config = OpenFoldSetupConfig.model_validate_json(config_path.read_text())
+    elif non_interactive:
+        config = OpenFoldSetupConfig()
+    else:
+        config = _prompt_for_config()
```
jandom, line 283: This is really nice

jandom (approved):
Excellent imo

PR #271, opened 2026-06-23
Triton attn backward: atomic-free d_pair_bias via dedicated kernel

GMNGeoffrey:
**Summary**
Using the Triton kernels, training is currently significantly slower and `_attn_bwd_dq` is ~35% of total training GPU time on MI300X. The culprit is the atomic add to compute dpair_bias. This PR replaces it with a separate kernel that reduces over the msa dimension internally (one plain store per tile, no atomics).

This gives a ~20-30x speedup of the Triton attention backward in microbenchmarks. Full e2e training speedup is ~2-3.5x and attention backward is no longer the bottleneck (<5% of total GPU time). It brings the Triton path down to ~10-25% faster than stock torch using the crop sizes in the examples. The performance improvement is not as dramatic as in inference, likely because of the much smaller token counts.

**Changes**
Add new `_attn_bwd_dpair_bias` Triton kernel for computing pair bias gradient

**Testing**
Passes the kernel unit tests (fixed in #258, so actually testing something).

Ran 8 steps of training using both random inputs and a subset of real inputs with both random initialized weights and fine tuning from the published checkpoint. Generally in bf16 gradients match with >0.99 cosine similarity. When fine-tuning from the existing checkpoint, gradient norms sometimes blow up, all in diffusion, when QK is very large and the true gradient is very small (<1e-6). The wrong gradients can be up to ~1e0, so a huge multiple, but not large in absolute terms. These all come from dV and dK, so I think the issue is not with the new kernel. This appears to be a feature of bf16 specifically: running the Triton kernel in fp32 does not exhibit it (gradients match to within 0.15%).

Verified DDP works (adds 10% step overhead)

**Other Notes**
I didn't do full e2e training to test convergence. I think this is sufficient evidence that the replacement kernel is generally correct and significantly better than the current Triton one (which is unusably slow). I'm not sure whether these grad blow-ups would actually cause problems in practice since they start small and there's clipping.

The bf16 gradient blow-up also points to a weirdness in the way we're currently force-casting to bf16 for these kernels. Contrary to the comment there, the kernel runs fine with fp32; it's just slightly slower. The Triton kernel also isn't a huge win for diffusion anyways (since it doesn't have the same N_seq scaling), so we could turn it off there if it proves to be the primary 

These concerns seem somewhat orthogonal to the kernel itself though, and I thought it was better to share this as it stands and we can continue doing further validation.

This PR includes changes from PR #258 to make the kernel tests not degenerate, since otherwise the CI won't really be validating much. I can rebase or update when that is merged. Only the last commit is part of this PR.

### DIFF
--- a/openfold3/core/kernels/triton/evoformer.py
+++ b/openfold3/core/kernels/triton/evoformer.py
@@ -387,7 +387,6 @@ if _TRITON_AVAILABLE:
         dQ,
         dK,
         dV,
-        d_pair_bias,
         M,
         D,
         stride_batch,
@@ -401,10 +400,6 @@ if _TRITON_AVAILABLE:
         stride_mask_batch,
         stride_mask_msa,
         stride_mask_seq,
-        stride_d_pair_bias_batch,
-        stride_d_pair_bias_head,
-        stride_d_pair_bias_seq1,
-        stride_d_pair_bias_seq2,
         HEAD,
         N_SEQ,
         SEQ_LEN,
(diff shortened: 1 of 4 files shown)


christinaflo:
We're aware of this issue, @singagan is working with us internally to fix it.

PR #274, opened 2026-06-23
Fixes errors from incompatibilities with biotite=1.7

jnwei:
## Summary
The recent update to [biotite=1.7](https://github.com/biotite-dev/biotite/releases/tag/v1.7.0) broke some openfold3 tests. There were 3 error modes, all of which are addressed in this PR

### Error 1: TypeError: BondList.__new__() missing 1 required position 
Example Failure: `openfold3/tests/test_structure_from_query.py::test_structure_from_query[standard_peptide] -`
Explanation: This failure arose due to a change in definition of the BondList constructor. This meant our previous snapshots which were pickled atom arrays had a different constructor.
Solution: Change test case records to use pytest-regression to record the previous run, and then use an explicit deserializer to inspect every field rather than assert_atom_arrays_equal. This solution is agnostic to the choice of biotite 1.6 or 1.7. _No changes to the codebase required._

### Error 2: ValueError: An atom cannot be bonded to itself
Example Failure: `test_permutation_alignment.py:: test_chain_connected_molecule_iter`
Explanation: The calling function `chain_connected_molecule_iter` creates a pseudo molecule which includes an artificial root atom intended to be bonded to all other atoms in the chain. This bond list included a bond to the root atom itself, which triggers an error in BondList validation in biotite=1.7
Solution: [Amend the artificial bond list](https://github.com/aqlaboratory/openfold-3/pull/274/changes#diff-3b4b1cab1a6f23074117d342009befa5b51d6e41e67dfda112e7afc0ad942057) to skip self-connections. **Requires a change to `openfold3.core.data.primitives.permutations.mol_labels.chain_connected_molecule_iter`**

### Error 3: AttributeError: 'list' object has no attribute 'array_length'
Example Failure: `openfold3/tests/core/data/primitives/structure/test_query.py::test_consistent_structure_from_smiles_and_ccd_code`
Explanation: These tests all call the helper function `processed_reference_molecule_from_mol`, which uses the method `biotite.structure.create_atom_names`. Previously, this method would work with a pure list of str. In biotite=1.7, a length check was introduced that only works for AtomArrays [reference commit](https://github.com/biotite-dev/biotite/pull/897/changes#diff-b02a833aa17be1bfabac982859f3ba99c6c12ebcf13056c2fc9ea947295ab816R147) which means that this function can no longer be used with a pure list.
Solution: [Create a temporary AtomArray](https://github.com/aqlaboratory/openfold-3/pull/274/changes#diff-975fc0a76594394c54ca4322e8483f74c6d3ba9c6d367173353e0f851184a5ae) from the rdkit Mol generated in `processed_reference_molecule_from_mol`, and pass that in to `biotite.structure.create_atom_names`. **Requires a change to openfold3.core.data.primitives.structure.query.processed_reference_molecule_from_mol`**

## Changes
- Updates root atom bond list in pseudo molecule used in `chain_connected_molecule_iter`
- Materialize a temporary AtomArray in processed_reference_molecule_from_mol
- Updates test case generation of test_structure_from_query examples.

## Testing
Tested faililng tests with both biotite 1.6 and biotite 1.7:
- openfold3/tests/test_structure_from_query.py 
- openfold3/tests/test_permutation_alignment.py 
- openfold3/tests/core/data/primitives/structure/test_query.py

## Other Notes
I am less sure about the solution for Error 3 that generates a temporary AtomArray just to use `create_atom_names`. We could consider pushing a fix upstream in biotite for this issue instead. My understanding is that this function is intended to only be used for small molecules / reference molecules, so hopefully this doesn't create a memory or runtime bottleneck.

Update: I have opened https://github.com/biotite-dev/biotite/issues/915 to discuss possible solutions


openfold3/core/data/primitives/permutation/mol_labels.py
```diff
@@ -192,8 +192,12 @@ def chain_connected_molecule_iter(
...
     root_atom_repeated = np.repeat(root_atoms, np.diff(chain_starts))
 
-    # Like [(0, 0), (0, 1), ..., (N, N), (N, N+1), ...]
-    root_atom_bond_pairs = np.column_stack((root_atom_repeated, np.arange(n_atoms)))
+    # Exclude self-bonds (i, i) that occur at each chain's first atom
+    all_indices = np.arange(n_atoms)
+    non_self = root_atom_repeated != all_indices
+    root_atom_bond_pairs = np.column_stack(
+        (root_atom_repeated[non_self], all_indices[non_self])
+    )
```
jandom, line 201: This certainly does give me head-spin (maybe it's the temperature, probably not :P)

I also don't think we had any tests for this function?

jandom: Might be worth keeping some comments – it's not always obvious when we're dealing with fully-dense matrices [I-elements, j-elements], or lists of [i, j] tuples like here

openfold3/core/data/primitives/structure/query.py
```diff
@@ -262,9 +262,10 @@ def processed_reference_molecule_from_mol(
     if atom_names is not None:
         mol = set_atomwise_annotation(mol, "atom_name", atom_names)
     else:
-        elements = [atom.GetSymbol().upper() for atom in mol.GetAtoms()]
-        atom_names = struc.create_atom_names(elements)
+        temp_struc = from_mol(mol, conformer_id=0, add_hydrogen=False)
+        atom_names = struc.create_atom_names(temp_struc)
```
jandom, line 268: maybe a comment here why this is happening? it's obvious in your PR description but the PR will get merged away and this could stick around

jandom (approved):
Awesome, this is pretty surgical – minor comments

openfold3/core/data/primitives/permutation/mol_labels.py
jnwei: Restored the old comment after some modifications. 

We actually do have a test for specifically `chain_connected_molecule_iter`, but not for the root atom creation. We can add a smaller test if that would be helpful?

openfold3/core/data/primitives/structure/query.py
jnwei: Done, referring to biotite issue I created.

PR #277, opened 2026-06-24
Pypi 0.4.2 rc candidate

jandom:
**Summary**

bump version numbers before release

**Changes**

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ If you use OpenFold3-preview in your research, please cite the following:
   title = {OpenFold3-preview},
   author = {{The OpenFold3 Team}},
   year = {2025},
-  version = {0.4.1},
+  version = {0.4.2},
   doi = {10.5281/zenodo.19001000},
   url = {https://github.com/aqlaboratory/openfold-3},
   abstract = {OpenFold3-preview is a biomolecular structure prediction model aiming to be a bitwise reproduction of DeepMind's AlphaFold3, developed by the AlQuraishi Lab at Columbia University and the OpenFold consortium.}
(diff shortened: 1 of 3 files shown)


jnwei:
This branch is the release candidate  for the next pypi release of openfold3.

In order to successfully build and use deepspeed:
- `nvidia-cutlass` is required to build deepspeed. This is not a stable long term solution as support for nvidia-cutlass is deprecated, with no replacement on pypi.

Then for user installation:
- Install the deepspeed optional dependency, e.g. `pip install openfold3[deepspeed]`
- Set `export CUTLASS_PATH=DS_USE_CUTLASS_PYTHON_BINDINGS`

Given the number of extra steps / deprecation of `nvidia-cutlass`, perhaps this is the right time to move away from deepspeed all together.

jnwei:
@sdvillal Do we want to incorporate https://github.com/aqlaboratory/openfold-3/pull/284 into this RC?

Looking to make this release today

sdvillal:
Regarding #284, I would not wait for it: for a pypi release the pixi workspace does not really matter, plus it makes me uneasy to ship things that are fairly untested. Note that #284 is quite disruptive as it forces users to update pixi (I have opened an issue in pixi itself and maybe we can wait to see if it is picked-up).

I would move the nvidia-cutlass dependency to its own group, separated from deepspeed. Or simply document the  options people have to setup cutlass towards compilation of deepspeed's Evoformer , as we discussed offline.

jnwei (approved):

PR #282, opened 2026-06-27
Do boolean mask sums directly in fp32

GMNGeoffrey:
**Summary**

In a few loss computations, we sum over a mask. These sums get computed in i64 by default, which is a bit slow. Then when we add epsilon (and even if we didn't, when we use them as a divisor), they get converted to fp32. The precision loss of just doing this in fp32 in the first place is negligible.

The diffusion loss sums over all atoms showed up on a training profile I did on MI300X. I saw ~5% training step speedup with that change. I did a sweep for the same pattern elsewhere and updated those as well, as I think it's just a good practice, but those are smaller and didn't show up in the profile (OTOH because they're smaller, the fp32 will be bit-exact). I could revert those commits if you prefer.

**Changes**
- Change sums over boolean mask tensors used in loss denominators to use float32 when the result is immediately added to a float epsilon.
- Dropped mask sums that cancel entirely.

**Testing**
- Did some sanity checking here that torch.sum in fp32 doesn't lose significant precision vs using int64 and then casting.
- Did a training step starting from trained weights and confirmed no major disruption (cosine similarity >0.999, relative grad norm difference <1%)
- Unit tests for large edge case inputs

### DIFF
--- a/openfold3/core/loss/confidence.py
+++ b/openfold3/core/loss/confidence.py
@@ -252,7 +252,7 @@ def all_atom_plddt_loss(
         * ((d < 0.5).int() + (d < 1).int() + (d < 2).int() + (d < 4).int())
         * pair_mask,
         dim=-1,
-    ) / (torch.sum(pair_mask, dim=-1) + eps)
+    ) / (torch.sum(pair_mask, dim=-1, dtype=torch.float32) + eps)
 
     # Compute binned lddt
     # [*, N_atom, no_bins]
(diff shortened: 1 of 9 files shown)


jandom:
Nice work @GMNGeoffrey – any chance we could bolt that behavior in with tests? I know this maybe ballooning the scope significantly, i not sure if we have a great harness for this part of the codebase

GMNGeoffrey:
Yeah it's a bit tricky because I need to use cases where we know what the answer should be a priori but any possible error from this only appears when the inputs get very large. So I kind of can only test somewhat degenerate cases, which maybe aren't the most useful. And to test cases where fp32 starts being imprecise, we need really large inputs that are impractical in unit tests. I actually wrote some tests that would fail if we changed the accumulator type to bf16, but wasn't sure if they were worth the complexity of setting them up. I'll push them and let you decide.

One thing I noticed as part of that: in a couple of these it's actually not possible to get a failure because the mask sum cancels in the algebra. So I've just removed them entirely and taken the ratio of the sums instead.

Issue #283, opened 2026-06-28
[BUG]pixi.toml CUDA environments are broken with pixi 0.71.1


ai-bits:
Tried to install with pixi latest on Ubuntu 24.04 latest, 256GB Xeon 3435X, 2x 32GB RTX Pro 4500 Blackwell, 20GB RTX 4000 Ada

Claude says: see title
deprecated system-requirements syntax on features is causing pixi to compute an empty platform set
CPU env works fine. The fix is to migrate the deprecated system-requirements syntax to the new platforms syntax with virtual packages, as pixi 0.71.1 suggests.

```
(openfold3) gy@gnu24:~/dl/openfold-3$ pixi run -e openfold3-cuda13 setup_openfold
 WARN Encountered 2 warnings while parsing the manifest:
  ⚠ the `[system-requirements]` table is deprecated in favor of virtual packages on `platforms`
     ╭─[/home/gy/dl/openfold-3/pixi.toml:276:3]
 275 │     
 276 │ ╭─▶   [feature.cuda12.system-requirements]
 277 │ ├─▶   cuda = "12.0"
     · ╰──── declare these on the `platforms` entries instead
 278 │     
     ╰────
  help: e.g. platforms = [{ platform = "linux-64", cuda = "12" }]

  ⚠ the `[system-requirements]` table is deprecated in favor of virtual packages on `platforms`
     ╭─[/home/gy/dl/openfold-3/pixi.toml:290:3]
 289 │     
 290 │ ╭─▶   [feature.cuda13.system-requirements]
 291 │ ├─▶   cuda = "13.0"
     · ╰──── declare these on the `platforms` entries instead
 292 │     
     ╰────
  help: e.g. platforms = [{ platform = "linux-64", cuda = "12" }]

Error: unsupported-platform

  × the environment 'openfold3-cuda13' does not support 'linux-64'
  help: supported platforms are
        Or install for a target this machine cannot run with `pixi install --platform <platform>`.

```
Claude wants to edit pixi.toml line 290:
```
[feature.cuda13]

  platforms = [
    { platform = "linux-64", cuda = "13" },
    { platform = "linux-aarch64", cuda = "13" },
  ]
```

sdvillal:
Thanks for reporting @ai-bits ! I thought that pixi had [deprecated but not yet disabled this syntax](https://github.com/prefix-dev/pixi/releases/tag/v0.71.0). I suspect this breaking change is a bug to report to the pixi folk.

[...]

PR #284, opened 2026-06-28
Update to pixi 0.72.0

sdvillal:
-----------------------------------------

**Summary**

Updates pixi.toml to use the new "rich platforms" and other features from pixi 0.72

**Breaking**

People will have to update their pixi: min-pin on pixi 0.72

```
requires-pixi = ">=0.72.0"
```

Pixi can be updated with `pixi self-update`

**Changes**

See summary

**Related Issues**

#283 https://github.com/prefix-dev/pixi/issues/6462

**Testing**

I have not yet tested this but the new environments solve. Waiting for CI and/or time to run tests locally.

**Other Notes**

This of course would force users to update to pixi 0.71. I believe we are hitting a bug in pixi (see related issues), as the old syntax should still be allowed.


sdvillal:
Another thing we probably should do is to disable 0.71 in the old workspaces.

sdvillal:
Also we might need to update pixi in the docker images fot the tests here to pass.

sdvillal:
For openfold3-cuda12-pypi and openfold-cuda13-pypi envs, we get upon creation:

```
WARN PyPI package files will overwrite files installed by conda packages:
  - PyPI package 'ninja' overwrites conda package 'ninja':
  - bin/ninja
```

This comes from deepspeed and should not be the case, as pypi ninja (referring to the python ninja bindings) and CF ninja (referring to ninja itself) are different packages. To investigate and maybe report to parselmouth.

jandom:
~~@sdvillal can you add me and jennifer as collaborators on your repo, otherwise we can't push~~

All working now, thx - i'm bumping pixi version for the docker builds

jandom:
@sdvillal should we instead go for 0.70.2 to avoid these problems for the time being?

https://github.com/aqlaboratory/openfold-3/issues/283#issuecomment-4830059570

sdvillal:
> @sdvillal should we instead go for 0.70.2 to avoid these problems for the time being?
> 
> [#283 (comment)](https://github.com/aqlaboratory/openfold-3/issues/283#issuecomment-4830059570)

Yes, a quick patch release with pixi-version = ">=0.70, <0.71" should give a better user experience for the time being.

I run tests with the new environment and got some failures, will need to figure later but I am swamped until Thursday.

jandom:
> Yes, a quick patch release with pixi-version = ">=0.70, <0.71" should give a better user experience for the time being.

Yeah that works but this manifest is not compatible with 0.70 (it has all the goodies coming from 0.71). 

Dropping this to draft, let's not worry about 0.71 for this release

sdvillal:
I meant indeed to pin pixi to <0.71 **in main in a hotfix**. This PR will need to wait for more testing - thanks for marking it as draft.

They seem to have [fixed the bug upstream](https://github.com/prefix-dev/pixi/pull/6471), once a new version of pixi is released I hope no users will experience this.

jandom:
> Yet another failure to convert system-requirements

Lol, they seem to be having a great time with this feature :D 

Let's not hotfix, i don't think this impacts too many people (probably N<10), the release prep is keeping us sufficiently busy

sdvillal:
FYI They just released the new version of pixi. I would from now on pin to minimum 0.71.3 (will do in this or other PR, or feel free to do anywhere).

FYI I suspect that updating to pytorch 2.12 may break some tests due to small tolerance violations, will report back when I find a minute.

sdvillal:
In the cuda-pypi environments, now pixi warns upon creation:

```
WARN PyPI package files will overwrite files installed by conda packages:
  - PyPI package 'ninja' overwrites conda package 'ninja':
    - bin/ninja
```

Seems that this still works even after clobbring

jandom:
Yeah, that's an annoying but benign warning

sdvillal:
The failing tests also happen in main, so I will report in a separate issue. Other than that, this should be ready to go (please review).

To watch:
- This does a general update, e.g., to torch or deepspeed.
- Results in clobbering of the ninja binary in pypi-centric environments, as commented above.
- Pyproject cuequivariance pins are now consistent with pixi.toml.

This will force users to update pixi, but also remove annoying deprecation warnings or risk of hitting the bug in #283.

jandom:
> The failing tests also happen in main, so I will report in a separate issue. Other than that, this should be ready to go (please review).

Thanks Santi, i bumped the docker stuff in this PR – main seems green so let's see if CI/CD runs clear or reproduces the errors you mentioned.

sdvillal:
Thanks Jan! We should just merge this.

The errors I observe, only in cuda envs, are likely due to testing context differences:

- Caching of MSAs, because I run the tests in parallel? Maybe we should ensure better isolation or more appropriate handling of caches.

```
FAILED openfold3/tests/test_inference_full.py::TestInferenceRun::test_protein_and_ligand
FAILED openfold3/tests/test_inference_full.py::TestInferenceRun::test_protein_only
```

```
        if raw_dir.exists():
>           raise FileExistsError(
                f"ColabFold raw directory already exists: {raw_dir}\n"
                "This is likely left over from a previous failed run. "
                "Please remove it before starting a new run."
            )
E           FileExistsError: ColabFold raw directory already exists: /tmp/of3-of-xxx/colabfold_msas/raw
E           This is likely left over from a previous failed run. Please remove it before starting a new run.
```

- Borderline numerical regression tests, likely because I use different GPU setup (linux-x64 A100, B300, software versions)?
```
FAILED openfold3/tests/test_triangular_attention.py::test_shape[cuda-True] - ...
FAILED openfold3/tests/test_triangular_attention.py::test_shape[cuda-False]
FAILED openfold3/tests/test_triangular_multiplicative_update.py::test_shape[cuda]
```

They all look like:

```
        # Snapshot regression: output must be numerically identical across runs.
        # CUDA tolerances are looser to accommodate hardware-level differences.
        # Regenerate with: pytest --force-regen
        tolerances = dict(atol=1e-6, rtol=1e-5)
>       ndarrays_regression.check(
            {"output": x.cpu().numpy()},
            default_tolerance=tolerances,
        )
E       AssertionError: Values are not sufficiently close.
E       To update values, use --force-regen option.
E
E       output:
E         Shape: (2, 22, 22, 128)
E         Number of differences: 26 / 123904 (0.0%)
E         Statistics are computed for differing elements only.
E         Stats for abs(obtained - expected):
E           Max:     1.8249847926199436e-06
E           Mean:    1.1431402526795864e-06
E           Median:  1.0728836059570312e-06
E         Stats for abs(obtained - expected) / abs(expected):
E           Max:     0.0034163997042924166
E           Mean:    0.002049802802503109
E           Median:  0.0018927527125924826
```

We can either relax tolerance from 1e-6 to 1e-5 or regenerate the expectations and detect the system where we are running.

jandom:
> Caching of MSAs, because I run the tests in parallel? Maybe we should ensure better isolation or more appropriate handling of caches.

The tests run in parallel okay and cleanup after themselves, unless there are certain types of failures. You can imagine that with certain failure modes we never get a chance to cleanup properly. We can probably improve on this – but i can't reproduce this failure and neither can CI/CD, so that's a challenge.

jandom:
Just to confirm, the only breaking change here is that our users (small number that rely on pixi, mostly devs) will have to bump pixi

```
requires-pixi = ">=0.72.0"
```

jnwei (approved):

PR #291, opened 2026-07-02
refactor: silence benign nvcc warning

jandom:
**Summary**

Removes the repeated nvcc warning : incompatible redefinition for option 'compiler-bindir' messages emitted when the DeepSpeed evoformer_attn CUDA extension is JIT-compiled (at first inference) in the CUDA pixi environments.

**Changes**

pixi.toml: in [feature.deepspeed-build.activation], set env.NVCC_PREPEND_FLAGS = "-ccbin=$CC" so the prepended -ccbin matches the one PyTorch injects. This is a no-op functionally — nvcc already used $CC (the last value) — it just stops the two from disagreeing.

**Related Issues**

**Testing**

**Other Notes**


jnwei (approved):
LGTM

PR #292, opened 2026-07-02
refactor: prepare deepspeed for ejection

jandom:
**Summary**

Slowly moving away from deepspeed, keeping it as dep for now but making a couple of changes 
- ~~changing the default inference kernel ~~ already done by @jnwei 
- moving into more lazy-import territory, rather than global imports

**Changes**

**Related Issues**

**Testing**

```bash
pixi run -e openfold3-cuda12       pytest openfold3/tests/test_kernels.py -v [passed]
pixi run -e openfold3-cuda12-pypi       pytest openfold3/tests/test_kernels.py -v [passed]
pixi run -e openfold3-cuda13       pytest openfold3/tests/test_kernels.py -v [passed]
pixi run -e openfold3-cuda13-pypi       pytest openfold3/tests/test_kernels.py -v [passed]
```

**Other Notes**


openfold3/core/utils/deepspeed_utils.py
```diff
@@ -0,0 +1,40 @@
...
+return False without importing it.
+"""
+
+import sys
+
+
+def deepspeed_is_initialized() -> bool:
+    """Whether DeepSpeed distributed comm is initialized. Never imports deepspeed."""
+    ds = sys.modules.get("deepspeed")
+    return ds is not None and ds.comm.comm.is_initialized()
```
jandom, line 31: This checks if the module is available without importing

openfold3/projects/of3_all_atom/config/model_config.py
```diff
@@ -113,8 +113,10 @@
                 "eval": {
                     "chunk_size": None,
                     "tune_chunk_size": tune_chunk_size,
-                    "use_deepspeed_evo_attention": not _is_rocm,
-                    "use_cueq_triangle_kernels": False,
+                    # Default inference kernels: cuEquivariance on CUDA, Triton on
+                    # ROCm. DeepSpeed remains available as an explicit opt-in.
+                    "use_deepspeed_evo_attention": False,
+                    "use_cueq_triangle_kernels": not _is_rocm,
```
jandom: This is the most consequential change for inference. I don't think we want to rush this PR out, so keeping it as draft

pyproject.toml
```diff
@@ -27,7 +27,9 @@ urls.Documentation = "https://openfold-3.readthedocs.io/en/latest/"
 urls.Repository = "https://github.com/aqlaboratory/openfold-3"
 urls.Issues = "https://github.com/aqlaboratory/openfold-3/issues"
 
-requires-python = ">=3.10"
+# Upper-bounded at <3.15: TorchScript (torch.jit.script/script_method, still used by
+# deepspeed) is unsupported and breaks on Python 3.15+, so 3.14 is the max runnable.
+requires-python = ">=3.10,<3.15"
```
jandom, line 32: drive-by fix – we're getting all these torch.jit.script deprecation warnings, things just won't work with py315

sdvillal: I think deepspeed newer version is almost free of script uses ([see this commit](https://github.com/deepspeedai/DeepSpeed/commit/d2ca6e75be18b278f0925720d9d50cc714c5e61a))

openfold3/core/model/primitives/attention.py
```diff
@@ -558,11 +552,19 @@ def _deepspeed_evo_attn(
...
         raise ValueError(
             "_deepspeed_evo_attn requires that DeepSpeed be installed "
             "and that the deepspeed.ops.deepspeed4science package exists"
-        )
+        ) from e
+
+    # Prevent deepspeed from doing triton matmul autotuning, which can hang if
+    # libaio is not installed and cause restart errors on preemption. Set here
+    # (rather than at import time) so it only runs when the DS kernel is used.
+    deepspeed.HAS_TRITON = False
```
sdvillal, line 567: Is there a way we could simply not do this? Or at least have a global config to allow the user to disable it.

openfold3/core/utils/deepspeed_utils.py
sdvillal: Do we require that ds comm is initialized for every use of deepspeed, also in simple inference workloads?

jandom: That is a fair question...

pyproject.toml
jandom: hopefully won't be an issue either way soon :P

openfold3/core/model/primitives/attention.py
jandom: Agreed

PR #300, opened 2026-07-09
Replace default inference kernel

jnwei:
**Summary**
Changes default kernel behavior to use triton attention kernels instead of base pytorch. 

**Changes**
- Default model_config kernel settings for inference
- `examples/example_runner_yamls/cuequivariance.yml` to use only cuequivariance, as deepspeed is no longer primarily supported.
- Update snapshots for triangle attention and triangle_multiplicative tests.
- Relax tolerance threshold for triangular multiplication tests: `tolerances = dict(atol=1e-6, rtol=1e-5) -> tolerances = dict(atol=1e-4, rtol=1e-4)`

**Related Issues**

**Testing**
- [x] All GPU tests locally
- [x] Run sample inference examples
- [x] Numerical similarity of current and new kernels are similar (<2Å)

**Other Notes**


jandom (approved):

jandom:
Sorry my revert broke this :(

openfold3/tests/test_triangular_attention.py
jnwei, line 1: @jandom nb I had to loosen the tolerance settings for this and the openfold3/tests/test_triangular_multiplicative_update.py to get the snapshot to work across GB10 and H100 architectures

sdvillal: Note, we also discussed about this [here](https://github.com/aqlaboratory/openfold-3/pull/284#issuecomment-4884025420). It should be fine as we are accounting for differences we cannot easily control.

PR #303, opened 2026-07-10
fix: make ColabFold MSA client robust to double-slash URL and unmappable template chains

muhammedhunaid:
## Summary
Two independent robustness bugs in the ColabFold MSA client (`openfold3/core/data/tools/colabfold_msa_server.py`) cause hard failures during MSA preprocessing. Both are fixed here, each with a fail-verified test.

## Bug 1 — double-slash request URL → HTTP 400
On the pipeline path the host is a `pydantic_core.Url` (`MsaComputationSettings.server_url`, `ColabFoldQueryRunner.host_url`). Stringifying a bare-authority `Url` yields the spec-canonical trailing slash: `str(Url("https://api.colabfold.com")) == "https://api.colabfold.com/"`. The request URLs are built as `f"{host_url}/{endpoint}"`, so the pipeline path produces `https://api.colabfold.com//ticket/msa` (double slash), which the server rejects with `400 "invalid ID"`. Plain string-host callers were unaffected (no trailing slash), which is why this only surfaces via the settings/pipeline path.

Fix: normalize the host once at the top of `query_colabfold_msa_server` — `host_url = str(host_url).rstrip("/")`. This covers both `Url` and `str` inputs and all four join sites (submit / status / download / template).

## Bug 2 — one unmappable template chain aborts the whole job
`remap_colabfold_template_chain_ids` raises `RuntimeError` when a pdb100 template hit references an author chain absent from the RCSB author->label map (observed: `7t3m_J`; RCSB 7t3m exposes A/C/D/G/H/I). It runs inside `query_format_main` during MSA preprocessing, so a single stale template hit aborts the entire prediction — even when templates are disabled downstream.

Fix: skip-and-drop unmappable hits (with a warning) instead of raising, filtering the alignment rows so the remapped column stays row-aligned. Mappable hits are preserved.

## Tests
- Replaced `test_unknown_author_chain_raises` (asserted the old raise) with `test_unmappable_author_chain_is_dropped` and `test_unmappable_hit_dropped_mappable_kept`.
- Added `TestQueryUrlConstruction`: `Url` and `str` hosts both yield single-slash submit/download URLs.

All new tests fail on the pre-fix code and pass after. `ruff check` and `ruff format --check` are clean.

## Out of scope
Noticed but intentionally not changed here: in `submit`/`status`/`download`, `error_count = 0` is reset inside the retry `while True` loop, so the `if error_count > 5: raise` guard never fires (unbounded retry on persistent errors). Happy to send that as a separate focused PR.

### DIFF
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -110,6 +110,13 @@ def query_colabfold_msa_server(
             is True, also returns a list of template paths for each sequence.
     """
 
+    # host_url may arrive as a pydantic Url (or a str). pydantic normalizes a bare
+    # authority to a trailing slash ("https://api.colabfold.com/"), so building request
+    # URLs as f"{host_url}/{endpoint}" would produce a double slash
+    # ("https://api.colabfold.com//ticket/msa"), which the server rejects with HTTP 400.
+    # Normalize to a plain string without a trailing slash so exactly one slash appears.
+    host_url = str(host_url).rstrip("/")
+
     submission_endpoint = "ticket/pair" if use_pairing else "ticket/msa"
 
     headers = {}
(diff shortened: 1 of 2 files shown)


jandom:
This is partially out of date with master, the ColabFold URL bug has long been fixed. @muhammedhunaid can you confirm this still happens on current 'main' branch?

jandom:
Closing as duplicate (we've already solved both of these on main, please re-open if needed)

PR #305, opened 2026-07-11
ci: use spot aws runners

jandom:
**Summary**

I don't know if this will work but i'm curious if we can switch to spot runners for our builders. 

**Changes**

- switch to `omsf/start-aws-gha-runner@v1.3.0` everywhere 
- switch to spot market for t3.2xlarge runs (these are probably v abundant) 

**Related Issues**

**Testing**

**Other Notes**


jnwei (approved):
Let's give this a shot. My concern is that some of our CI tests (the conda CI) typically run for ~45 min, and so it may get preempted before completing. We can monitor the test to see how frequently this occurs.

Issue #235, opened 2026-05-28
CIF Direct templates have apparent negligible influence on predicted conformations/steering


sukritsingh:
## High level summary
After the merge of PR #37 I decided it would be fun to try and replicate a [Colab Notebook I use for conformational seeding from AF2](https://colab.research.google.com/drive/1Dcb5ZouYJQfyq1np-TBmKJE8DOAbA9dU?usp=sharing), and wrote a [colab notebook for conformation-based seeding for OF3](https://colab.research.google.com/drive/1Kne4ZNCrgYRQNGMkU899IY7wJuTNw3V6?usp=sharing) to see how well OF3 can generate conformations based on given templates. 

**TLDR:** Unfortunately, it does not seem to have much effect of biasing the conformations in useful directions for seeding new simulations

## Ideally:
Ideally, I would love to use OF3 to try and do this "Multistate" type of conformational seeding that is very useful for parallel simulations of kinase structures for biophysical analysis (like state-breakdown, conformational analysis upon activation, etc.)

I have written out the full "how to reproduce" below, but I wanted to summarize my protocol and findings first:

### Protocol:
1. Using a given kinase sequence, and 20 chosen CIF file templates (10 each for one of two state labels). This gives us two "states" to compare against that should diverge drastically 
2. I generated 4 sets of predictions, varying both the template state, and whether or not an MSA was used.

### Result summary 
Summary: MSA matters for even getting a correct kinase fold. Passing different template structures/state does not seem to make a difference to the kinase state generated.

Table below:

| run | MSA mode | templates (DFG state) | top-sample mean pLDDT | qualitative fold | conformation observed |
|---|---|---|---|---|---|
| **A** | `msa_free` | BLBplus (DFGin) | **32.2** | incorrect non-kinase | (n/a — no kinase fold) |
| **B** | `msa_server` | BLBplus (DFGin) | **81.0** | kinase domain  | **BLAminus** (not BLBplus) |
| **C** | `msa_free` | BBAminus (DFGout) | **32.2** | incorrect non-kinase | (n/a — no kinase fold) |
| **D** | `msa_server` | BBAminus (DFGout) | **81.1** | kinase domain  | **BLAminus / BLBplus hybrid** (not DFGout) |

Images below compare "correct" output (first), "with MSA + templates" output (second image - all overlaid kinase structures) versus "No MSA" OF3 outputs (completely wrong - doesn't even generate a kinase structure) to highlight the need for MSA and that the templates make no difference

[...]

Issue #234, opened 2026-05-28
[QUESTION] Minimal template influence with single sequence?


rostro36:
**Describe the issue**
I want to test the effect of some changes to protein templates. AF2 had the nice trick, that if you provide just single sequence and template and for normal sequences, [it would basically just follow the template with reasonably high confidence](https://ibmm-unibe-ch.github.io/template-analysis/), since it can't learn anything from the MSA.

When I do the same single seq experiment with OpenFold3, the difference between with and without template is minimal, both in confidence and actual structure. The attached outputs below show that the templates get correctly provided, but it seems to me like the models actively ignore the templates instead of leaning into them. 

Am I being blonde and there is something wrong with my input or installation?
Is that expected behaviour from the change to diffusion or has template influence been tested in an end-to-end fashion?
Is there some other way/hack that you can think of to increase template influence?

Let me know if I can help you by providing more information or running some tests.
Thanks for your help! 
 
**To Reproduce**
OpenFold3 compiled from source from a [pull this week](https://github.com/rostro36/openfold-3) and `of3-p2-155k.pt`. 
- CLI:
> run_openfold predict --query_json=1A8Q_template.json --use-msa-server=False --use-templates=True --output-dir 1A8Q/masked
- query JSON:
```
     {
	"queries":{
		"monomer": {
			"chains": [
				{
					"molecule_type": "protein",
					"chain_ids": "A",
					"sequence": "PICTTRDGVEIFYKDWGQGRPVVFIHGWPLNGDAWQDQLKAVVDAGYRGIAHDRRGHGHSTPVWDGYDFDTFADDLNDLLTDLDLRDVTLVAHSMGGGELARYVGRHGTGRLRSAVLLSAIPPVMIKSDKNPDGVPDEVFDALKNGVLTERSQFWKDTAEGFFSANRPGNKVTQGNKDAFWYMAMAQTIEGGVRCVDAFGYTDFTEDLKKFDIPTLVVHGDDDQVVPIDATGRKSAQIIPNAELKVYEGSSHGIAMVPGDKEKFNRDLLEFLNK",
                    "template_cif_paths": ["1A8Q_A.cif"] #with or without this line makes minimal difference to output
				}
			]
		}
	}
}
```

The provided cif is the downloaded file from the PDB.
I also tried 1A03 and 10DJ .

**Expected behavior**
Structure predictions very close to the provided template.

**Stack trace**
[...]

PR #306, opened 2026-07-11
Fix templates during inference

rostro36:
**Summary**
Fix templates during inference
**Related Issues**
#234 #235 
**Testing**
#234 as expected


jandom:
Hah, i'll be damned - could that be it? Thanks for this, we'll take a look for sure!

jandom:
@rostro36 thanks so much for this – I think this is real, the value is just used as sentinel – without it the whole templating branch is dropped. 

Our template code is a royal mess, imo, I've taken a few stabs at it already.

jandom (approved):
I'm leaning to just merge it – this codepath is uncovered by tests, so everything is deceptively green before and after the merge. 

The code lives on a fork so it's not possible to push to this remote branch, and elaborate a bit more on the PR beyond the minimal 1-line fix. We'll have to do that in follow-up PRs anyway.

PR #308, opened 2026-07-12
Making `max_seq_count` actually truncate parsed MSAs

### LINKED ISSUE #307: [BUG] max_seq_counts is a no-op — MSA per-file caps never applied, altering profile features
_I have a fix and regression tests ready for this problem._

**Describe the bug**
`parse_a3m()` and `parse_stockholm()` (openfold3/core/data/io/sequence/msa.py) accept a `max_seq_count` argument, but it has no effect. Both call` MsaArray.truncate(max_seq_count)` as a bare statement. truncate defaults to `inplace=False`, so it computes a truncated copy and returns it, but the caller discards the return value and keeps the original, full-length MSA. The result: every configured cap in `MSASettings.max_seq_counts` (uniref90_hits=10000, uniprot_hits=50000, mgnify_hits=5000, …) is inert, and all parsed sequences are retained.

**To Reproduce**
```
from openfold3.core.data.io.sequence.msa import parse_a3m

a3m = ">q\nABCD\n>a\nABCD\n>b\nABCD\n>c\nABCD\n>d\nABCD\n"
print(len(parse_a3m(a3m, max_seq_count=1)))   # prints 5 --> should have been 1.
```
parse_stockholm behaves identically.

**Expected behavior**
`parse_a3m `/ `parse_stockholm` return at most max_seq_count sequences.

**Additional context**
This is not masked by the downstream max_rows subsampling. In pipelines/sample_processing/msa.py, the profile and deletion-mean features are computed from the full concatenated MSA (main_msa_redundant, lines ~324–329) before max_rows is applied, so those features currently reflect all sequences and would change once the caps take effect. The caps also change which databases fill the max_rows budget. Fixing this alters model inputs, so I'd like to confirm the intended semantics (are the max_seq_counts values meant to bind?) before opening the fix.

While in `MsaArray.truncate`, two adjacent minor issues: a ValueError that is constructed but never raised (so wrong types pass silently), and a slice(min(row_slice.stop, self.__len__())) line that is redundant for int callers (numpy/pandas already clamp out-of-range slices) and lossy for slice callers (discards start/step).

I have a fix and regression tests ready, and can open the PR once the intended behavior is confirmed.

Pana-TsK:
**Summary**

_Behavior change_: Previously all parsed sequences were kept. The final MSA-rows feature was then bounded by the downstream max_rows (16384) subsampling, but the profile and deletion-mean features are computed on the full pre-subsample MSA (main_msa_redundant), so they were never bounded. This fix therefore changes those features for any chain whose source MSAs exceed their caps, and also changes which databases populate the max_rows budget. Chains with shallow MSAs (already under every cap) are unaffected.

`parse_a3m()` and `parse_stockholm()` called `MsaArray.truncate()` without `inplace=True`, leading to Msa truncation being calculated, but set to return a new MsaArray object, which was then discarded.

Additionally, The method also constructed a ValueError without ever raise-ing it (the same discarded-expression bug), so a wrong-typed input passed silently instead of erroring, which I removed. The line `row_slice = slice(min(row_slice.stop, self.__len__()))` is redundant. It attempts to clamp the count so the caller can never ask for more rows than exist, but numpy and pandas already clamp out-of-range slices, indexing past the end just returns everything, without error. the regression test test_truncate_over_length_keeps_all_rows covers this.

**Changes**
`parse_a3m()` and `parse_stockholm()` now use  `inplace=True`, so the methods return truncated MSAs as intended. Additionally, the mentioned valueError has been deleted (it would never fire), the line  `row_slice = slice(min(row_slice.stop, self.__len__()))` has also been deleted.

**Testing**
AI-generated regression tests (simple, easy to check & reason about), have been written, to show that this issue was real and now has been resolved.


jnwei (approved):
This is a great contribution, thank you @Pana-TsK !

My only suggestion would be to move the test files to where our other parser tests are located (e.g. `openfold3/tests/core/data/io/sequence/template`) , but I can handle that move directly

Pana-TsK:
Thank you for the review, I will keep this in mind for next time and will put the tests in their appropriate folders. While it is not a correctness fix, the MSA truncation is a recurring cost per epoch (if I am not mistaken), so I hope with this to have made the OpenFold fine-tuning and training more efficient for deep MSAs.

Again, thank you very much!

PR #312, opened 2026-07-16
fix: correct column misattribution in calculate_profile's chunked bincount

Pana-TsK:
**Summary**
Currently, `calculate_profile` flattens each chunk with `msa.ravel()` (row-major: walks columns fastest within each row), but built
col_indices_local with np.repeat(np.arange(blocks_n_cols), n_rows). The two orderings disagree, so bincount attributes each row's residue to the wrong column.

Every misattributed entry is still counted under some column; so `counts[col, :].sum() == n_rows`, regardless of whether the attribution is correct — this is why the bug produces a plausible, correctly-shaped profile and wasn't being caught by the sum check. At the configured chunk_size = 1000, only ~.1% of the entries would land in their correct column at realistic MSA depths (I check n_rows 10 through 16384).

**Changes**

Fix:[ `np.repeat`](https://numpy.org/doc/stable/reference/generated/numpy.repeat.html) -> [`np.tile`](https://numpy.org/doc/stable/reference/generated/numpy.tile.html), which reproduces `.ravel()`'s row-major order with no extra copy (unlike switching .ravel() to Fortran order instead).

**Testing**

I added (AI-generated but simple to follow & understand) regression tests in tests/core/data/primitives/sequences/msa/: a hand-computed example (three distinct, assymetric residues so it can't pass by coincidence), a randomized cross-check against an independent reference implementation across 10 shapes (two of them deliberately degenerate, where reap and
tile collapse to the same array), a chunk-size invariance check, and a test documenting why a column-sum check alone would not have caught this.

**Other Notes**

Impact: this affects only the 'profile' feature (a 32-dim per token, computed once per chain from the full pre-subsample main MSA). Does not affect the per-row MSA tensor (msa/has_deletion/deletion_value) that` MSAModuleEmbedder` attends over, which is built via a separate code path. Profile does however food InputEmbedderAllAtom's s_input (same channel budget at restype, seeding both s and z), so in my opinion this is not negligible; actual effect on trained accuracy is unmeasured here and
would need an ablation.

### DIFF
--- a/openfold3/core/data/primitives/sequence/msa.py
+++ b/openfold3/core/data/primitives/sequence/msa.py
@@ -1213,7 +1213,7 @@ def calculate_profile(
         # Flatten subarray (size = n_rows * block_n_cols)
         val_indices = msa_chunk.ravel()  # row-major flatten by default
         # Build local col_indices of the same flattened shape
-        col_indices_local = np.repeat(np.arange(block_n_cols), n_rows)
+        col_indices_local = np.tile(np.arange(block_n_cols), n_rows)
         # Now each col in this chunk is offset from the "absolute" col_start, but for
         # bincount we just care about "relative" indexing from 0...(block_n_cols-1). We
         # combine into a single 1D array: offset + val Where offset = col_indices_local
--- a/openfold3/tests/core/data/primitives/sequence/msa/test_calculate_profile.py
+++ b/openfold3/tests/core/data/primitives/sequence/msa/test_calculate_profile.py
@@ -1,0 +1,215 @@
+# Copyright 2026 AlQuraishi Laboratory
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


jnwei:
Thank you very much @Pana-TsK for the finding this bug and submitting a bugfix! This is indeed a serious bug; we do not yet have the ablation studies to measure its impact.

We had caught this bug internally. The fix is incorporated for the next set of OpenFold3 models, which estimated to be released soon.

For `openfold3<=0.4.3` with the preview 2 weights (e.g. checkpoint `of3-p2-155k.pt`) the model has been trained with the incorrect version of the MSA profile features. If we were to update the MSA profile computation in the codebase now, while keeping the model weights as is, the model would have much worse inference performance since the MSA profile features no longer match the distribution of features that the model was trained on. 

I will close this PR for now. We really do appreciate your careful review of the codebase, it goes a long way towards improving our models and our understanding of the model's performance. We would warmly welcome future contributions.

Pana-TsK:
Hi @jnwei , I am glad to hear it was already caught! I have 2 questions; 

1. Is there a list of other bugs that are already caught? I found some other ones, it would be good to cross-check other bugs I might find? If it's in the openfold-3 slack channel, I can find them.
2. Regarding contribution etiquette, should I submit an issue next to PRs, or can I just submit PRs?

In the meanwhile, when I have time, I will work on the documentation inconsistency as promised.

jnwei:
Thank you @Pana-TsK !

1. I will ask the team to prepare a list of bugs to share in the slack. The MSA profile error is the biggest bug by far. At the very least, a list of changes will reported as part of the report for the next model release.
2. Happy to review a PR directly or to discuss as an issue first, whatever is easier for the topic in question.

PR #315, opened 2026-07-16
Multimer templates inference by unmasking

rostro36:
**Summary**
Use [unmasking](https://github.com/clami66/AF_unmasked) (also check their [paper](https://www.nature.com/articles/s41467-024-52951-w)) to support multimer templates during inference.

Especially compare with [this file](https://github.com/clami66/AF_unmasked/blob/af3/AF3_cross_chain_templates.patch) that shows the most important changes.
**Changes**
1. Move template masking away from feature creation and implement it in template embedder (similar to AlphaFold) (commit 4d38c6afe4dc1976657be4bc1436fdda57b0e587)
2. Add a CLI that adds a unmasked parameter to the model (commit 045b5d515c51e563ea4213b721e713319c838f14)
3. Add example input/output from (see testing), will be removed (commit b4f80c6e2137f615510319a51e8f277abdf30383)
4. Some mistakes fixed to get testing (commit db38d056b697207b2aa1db0ea88a3e859dc2fc75)

**Testing**
`python run_openfold.py predict --query-json=../examples/example_inference_inputs/query_multimer_templates.json --use-msa-server=False --use-templates=True --output-dir ../example_dir/ --unmasked True`
The mask in `example_dir/unmasked_True.txt` has the size of the full structure.
vs.
`python run_openfold.py predict --query-json=../examples/example_inference_inputs/query_multimer_templates.json --use-msa-server=False --use-templates=True --output-dir ../example_dir/`
The mask in `example_dir/unmasked_False.txt` has the size of the different monomers.
The features also notably change; more checks can be done this way, but they are harder to look at.
**Other Notes**
The printing to the file will be removed, together with `example_dir`. The CLI is my first idea, but can be changed. If you are interested in the feature, you can direct me where to add more documentation.

The differences are smaller than I anticipated from the AF2_unmasked results, where probably the template chain alignment was most useful. Also looking more closely at the paper, the difference starts to shine when forcing more samples.


jandom:
This is really awesome, multi-chain templates – thank you @rostro36 ! It's protein only right? No small molecule templates?

rostro36:
Great to hear that you are interested! I also thought if it is possible to implement molecules, but it seems it is not the case out of the box and in general needs a bigger change with probably rather negative effect on post-hoc inference.

Explanation:
I do not actively add anything to template preprocessing. The four template features are `pseudo_beta_atom_coords`, `template_unit_vector`, `template_restype` and `template_distogram` ([computed using `pseudo_beta_atom_coords`](https://github.com/rostro36/openfold-3/blob/98b143ccdcb71ea1cc7c156c113a395d90cf9bb4/openfold3/core/data/primitives/featurization/template.py#L195)).  [`template_unit_vector` needs `frame_atom_coords` which has coordinates of N, CA, and C](https://github.com/rostro36/openfold-3/blob/98b143ccdcb71ea1cc7c156c113a395d90cf9bb4/openfold3/core/data/primitives/featurization/template.py#L259). [They get collected using TemplateSlices from process_template_structures_of3](https://github.com/rostro36/openfold-3/blob/98b143ccdcb71ea1cc7c156c113a395d90cf9bb4/openfold3/core/data/framework/single_datasets/inference.py#L256), [which only allows protein chains](https://github.com/rostro36/openfold-3/blob/98b143ccdcb71ea1cc7c156c113a395d90cf9bb4/openfold3/core/data/pipelines/sample_processing/template.py#L98).

Also, the [current documentation](https://openfold-3.readthedocs.io/en/latest/template_how_to.html) says the same.

**To newly implement molecule support**, N, CA and C from `frame_atom_coords` are ill defined for arbitrary molecules, also beta atoms are ill defined for molecules on a theoretical level. So, not obvious to implement during inference and I think this unmasking hack might not work for a pre-trained template-embedder that has never seen molecule templates, if I am not mistaken. For unmasking on the other hand, the model might interpret at certain points that the unmasked mulitmer is just a very long monomer, which it can handle.

rostro36:
I removed my debug code and wrote some tests instead. You can still access the debug code on an older commit.
Further, I added some documentation and did linting.

Now, this could maybe be merged squashed and merged after review (and I did not miss a test).
But already from the public side, I see that it is a busy week with many interesting PRs, so no need to stress from my side.

siyuanj:
An independent replication of the same idea on a different base model, since the result lines up with your note that "the differences are smaller than I anticipated".

We ported the same change to AlphaFold 3 v3.0.4 — remove the `multichain_mask` in `evoformer.py` so cross-chain template pairs survive into the template distogram and backbone unit-vector features — and benchmarked it on protein–nucleic-acid complexes with **oracle** templates (the experimental structure of the target itself, split into one single-chain file per protein chain), so these numbers are an upper bound on what any template could do. Two paired comparisons, each isolating the mask and nothing else:

| comparison | n | interface lDDT (protein–NA, 30 Å) | targets above the success threshold |
|---|---|---|---|
| byte-identical input, mask on → off | 15 | median Δ +0.0010, p=0.83 | 3 → 3 (1 up, 1 down) |
| per-chain templates, mask on → off | 172 | median Δ +0.0001, mean +0.0001, 33 better / 30 worse / 109 unchanged, p=0.62 | 48 → 46 (6 up, 8 down) |

Two-sided Wilcoxon signed-rank on paired targets; threshold is interface lDDT ≥ 0.236, the DockQ > 0.23 equivalent from AF3 Extended Data Fig. 9.

It is not a no-op. In the n=15 comparison all 15 predictions changed substantially between the two arms (4.5–59.6 Å unsuperposed Cα RMSD), so the cross-chain geometry does reach the model — it just does not translate into interface accuracy. The one mildly positive signal was on the protein side: protein assembly Cα RMSD mean Δ −1.39 Å, 78 better / 48 worse, p=0.004 (not surviving correction for ~100 tests, but the sign was stable across a preliminary n=22 and the full n=172). The protein–nucleic-acid interface itself did not move.

Two caveats before anyone reads this as a verdict on this PR: different base model, and our target set is hard cases (targets where at least one co-folding model we benchmark gets the assembly wrong) in a protein–nucleic-acid setting, not the protein–protein multimer setting this PR is aimed at. On AF2-Multimer, unmasking demonstrably does pay off there (AF_unmasked; also gapTrick), so a null on AF3 protein–NA says little about OF3 on protein multimers.

Three explanations our design cannot separate, in case they are useful for whatever benchmark gets run here:

1. **Out of distribution.** AF3's first author states the model was never trained with multimer templates (google-deepmind/alphafold3#251). But that is equally true of AF2-Multimer, where unmasking works, so OOD cannot be the whole story.
2. **MSA depth — the one we would test first, and it is cheap.** Augustin-Zidek (DeepMind): "AlphaFold 2 and 3 don't pay much attention to templates unless the MSA is shallow" (google-deepmind/alphafold3#464), and the official multimer-template workaround explicitly includes clearing the MSA "to force the model to pay more attention to the template(s)" (google-deepmind/alphafold3#346). Every arm above carried a full MSA. A 2×2 of {mask on, mask off} × {full MSA, empty MSA} separates "the model cannot use cross-chain templates" from "the model was not looking at templates at all". If the effect appears only in the empty-MSA row, this PR is worth more than the current headline number suggests.
3. **Sampling** — you already spotted this; part of AF_unmasked's gain comes from forcing more samples.

One documentation suggestion. Whether cross-chain geometry survives at all depends on the template chains staying in their parent structure's frame: if a user superposes or recenters each template chain onto its own query chain, unmasking exposes pair features that no longer encode the correct relative placement, which is worse than masking them. Your example JSON does the right thing (one cif, different `template_cif_chain_ids` per chain), and OF3's per-chain `template_cif_paths` schema makes it natural — in AF3 we had to patch the template feature cache, which is keyed on entity (sequence), so every copy of a homo-oligomer was handed the first copy's template coordinates; 226 of our 256 targets have a repeated sequence, so with the mask off that silently tells the model the copies are superimposed. Worth one line in `template_how_to.md` so nobody reintroduces it on the input side.

Happy to share the AF3 patch and the scoring scripts if they would help with a comparison.

rostro36:
Good day @siyuanj ,
Thanks for making all of these tests! Sounds interesting to me.
May I ask a couple questions? Don't feel bad for not knowing or not making the effort in checking, it is just some discussion in basically an internet forum.
1. Did I get it right, that your unmasking only applied to protein chains? Or can it also directly unmask RNA structures?
2. How many structures per target did you predict? Is it one prediction per target, per mask (e.g. 15x2=30 for the first dataset). Can you give a feeling how big the differences are between predictions of the same protein with the same mask, but different random seed or different AlphaFold3 models?
3. You mention above or below threshold. Was it a big improvement or decline for those that flipped above/below threshold or was it closer to a rounding error?
4. Did you control for random seed of AF3? Since we slightly change the input, the random seed might be less relevant though.
5. It sounds like you ran with full MSA. You don't have to specifically check, but would you say the MSAs were deep at the interface (roughly>128 non-gap rows at the interface)?

I guess my take-away is that multimer templates can have a positive effect, but this effect is pretty niche for most tasks. Proteins with shallow MSAs profit more from it and the more you sample, the higher the chance you pick up a prediction with this signal. As far as I can tell, there is no downside to using multimer templates (but the upside is also generally pretty small).
Would you agree?

PR #317, opened 2026-07-17
Standalone script for running training integration test

jnwei:
**Summary**
A standalone training integration script that can be invoked with `./scripts/datasets/train_on_subsets.sh`

The steps include:
- Creates a mini PDB dataset using slices of the OpenFold3 PDB training set. This includes the dataset.json, the MSA, template, and original structure files
- Creates a `runner.yaml` whose fields correspond with the mini PDB dataset
- Runs a small training run.

**Changes**
- Expands the `./scripts/dataset` directory to include scripts and helper to set up a training set directory.

**Related Issues**

**Testing**
- Ran this script to test normal behavior and help debug training errors (e.g. #281)

**Other Notes**

Several changes would need to be made before this could be used as a full integration test

- Integration with pytest / adding sensible asserts (e.g. training loss decreases)
- Stopping the training run upon errors in the training run (by default, these errors are recorded and then skipped)


scripts/datasets/train_on_subset.sh
```diff
@@ -0,0 +1,22 @@
...
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+RUNNER_YAML="${SCRIPT_DIR}/train_pdb_subset.yaml"
+
+echo "Downloading PDB training subset (train=32, val=16 by default)..."
+pixi run --manifest-path "${REPO_ROOT}/pixi.toml" -e openfold3-cuda12 \
+  python "${SCRIPT_DIR}/download_subset.py"
+
+echo "Starting training run with ${RUNNER_YAML} ..."
+pixi run --manifest-path "${REPO_ROOT}/pixi.toml" -e openfold3-cuda12 \
+  run_openfold train --runner-yaml="${RUNNER_YAML}" "$@"
```
jandom, line 22: My main comment and it's a fairly annoying one – would be to do everything programmatically via a pytest harness. This way we get this running in all sorts of environments, not just the pixi ones. 

But it comes with some risks: the ergonomics is different, the logs are swallowed by the pytest harness by default , might be harder to work with than a bash script

jandom: Also, a follow up would be how is this triggered – custom GHA workflow? Manual only for now?

jnwei:
Closing this as this is superceded by #330 and  #338

PR #318, opened 2026-07-17
Add fused kernels for fast, low-memory inference

heathcliff233:
## Summary

This PR adds seven Triton kernel modules and six supporting execution/memory optimizations for fast, low-memory OpenFold3 inference. Together they reduce HBM traffic, avoid unnecessary O(N²) intermediates, and provide `optimized` and `sweetspot` operating points, with PyTorch/cuEquivariance fallbacks. There is no additional dependencies needed.

## Current overhead

Let `U = N² × c_z × sizeof(dtype)` be the storage of one trunk pair tensor. For N=1,369, `c_z=128`, and fp32, 1U is approximately 0.894 GiB. The model must retain `z` (1U), but operator-by-operator execution creates additional short-lived tensors:

- LayerNorm output, pair projections, residual updates, and OPM output each cost another O(1U).
- The Pairformer SwiGLU transition (`n=4`) writes two 4U hidden projections plus activation/product intermediates before projecting back to 1U.
- Chunked triangle attention still produces O(C × N × channels) projections for chunk size C; materialized attention scores are O(C × heads × N²).
- Diffusion attention scores are O(samples × heads × N²), or `samples × heads / c_z` U for each score-sized fp32 tensor.
- Four legacy templates carry approximately `4 × (39 distogram + 3 vector) / 128 = 1.31U` of pair features before masks and other inputs.
- Two 64-bin sample-scaled confidence outputs (PDE and PAE) are approximately `samples × 1U` in total if retained on the GPU.

These tensors also increase HBM traffic: one operator writes an intermediate that the next immediately reads. CPU offload can reduce residency, but adds transfers and does not remove the allocation or computation.

The current path also repeats invariant work:

- every diffusion block recomputes its pair-bias projection on every one of 200 rollout steps;
- conditioning and atom representations contain rollout-invariant components;
- relative-position embedding makes four separate gather-add passes over the same pair output; and
- shape specialization can compile equivalent kernels again for each target length.

The cuEquivariance triangle-multiplication path is CUDA-specific, exposes no row/K chunking control, and uses shape-specific tuning configurations. New sequence lengths can therefore require new tuning work, while its peak workspace cannot be bounded with an explicit chunk cap.

## Proposed approach

1. **Fuse producer/consumer operations**
   - Pair transition: LayerNorm → two SwiGLU projections → activation/product → output projection → mask/residual, without storing the 4U hidden tensors in HBM.
   - Triangle attention: process a bounded row block through LayerNorm, packed Q/K/V/gate projection, flash attention, output projection, and residual write.
   - Triangle multiplication: reuse LayerNorm statistics, retain the contraction's native layout, and fuse normalization/gating/output writeback.
   - Input/template paths: fuse relative-position gather-add and construct coordinate-derived template pair features directly into projected channels.

2. **Bound unavoidable workspaces**
   - A row chunk of size C changes pair-shaped workspace from O(N² × channels) to O(C × N × channels), approximately `C/N` of the corresponding full pair tensor.
   - Flash triangle and diffusion attention keep score tiles on chip instead of writing complete score matrices to HBM.
   - Fused triangle multiplication uses about 4U without chunking and about 2.2U with the 128-row/K schedule, compared with about 7U for the previous path in local measurements.

3. **Write directly and shorten lifetimes**
   - Accumulate normalized OPM chunks directly into `z` instead of returning a second 1U update.
   - Fuse residual writes where aliasing is safe.
   - Release template/MSA features and the trunk pair tensor immediately after their last consumer.
   - Run multi-sample confidence one sample at a time and store final pair logits on CPU, keeping GPU pair scratch O(1) in sample count.

4. **Remove duplicated rollout work**
   - Prepare invariant diffusion conditioning and atom representations once.
   - Optionally cache each diffusion block's pair-bias projection across all 200 steps. For 24 blocks and 16 heads, this costs `24 × 16 / 128 = 3U` in fp32, so it is a speed-for-memory option.
   - Exclude runtime lengths/strides from Triton cache keys when one masked operation signature already supports them.

5. **Use Triton as the common kernel layer**
   - The same layer covers triangle operations, transitions, diffusion attention, relative-position embedding, and template projection.
   - Fused trimul is an independent alternative to cuEquivariance. It supports explicit row/K chunking and reuses one compiled operation signature across sequence lengths, avoiding length-specific tuning.
   - cuEquivariance remains supported and can provide higher warmed throughput on NVIDIA GPUs; the Triton path prioritizes bounded memory, predictable compilation, and implementation control.
   - Triton provides the AMD ROCm kernel path while retaining portability across supported GPU backends. The local benchmark numbers below were collected on an RTX 4090.

## Optimized and sweetspot configurations

Both enable fused LN-linear, SwiGLU, triangle attention (cap 128), triangle multiplication, diffusion attention, coordinate-template embedding, and in-place OPM. cuEquivariance handles eligible fallback triangle operations.

- `optimized`: unchunked fused trimul, diffusion pair-bias cache enabled, offload effectively disabled.
- `sweetspot`: trimul chunk cap 128, pair-bias cache disabled, offload cutoff 512.

```bash
# optimized
OPENFOLD3_FUSED_TRIMUL=1
OPENFOLD3_TRIMUL_CHUNK_CAP=
OPENFOLD3_DIFFUSION_PAIR_BIAS_CACHE=1

# sweetspot
OPENFOLD3_FUSED_TRIMUL=1
OPENFOLD3_TRIMUL_CHUNK_CAP=128
OPENFOLD3_DIFFUSION_PAIR_BIAS_CACHE=0
```

Latest local profile: MCL1, N=1,369, 10,201 atoms, one sample, RTX 4090, after an in-process warm-up. `U=0.894 GiB`. The resident baseline is measured immediately before the forward; stage deltas are above each stage's own entry baseline and are not additive.

| Configuration | Resident baseline | Overall peak | Activation above resident | Peak stage delta | Wall |
| --- | ---: | ---: | ---: | --- | ---: |
| `local_default` | 2.85 GiB | 15.14 GiB | 13.75U | MSA: 12.34U | 84.41 s |
| `optimized` | 1.68 GiB | 6.25 GiB | 5.11U | Diffusion: 4.42U | 48.74 s |
| `sweetspot` | 1.68 GiB | 4.09 GiB | 2.70U | Confidence: 2.02U | 57.98 s |

Against `local_default`, `optimized` reduced peak memory by 58.7% and wall time by 42.3%. `sweetspot` reduced peak memory by 73.0% and wall time by 31.3%. The 3U pair-bias cache explains why `optimized` is faster but peaks in diffusion; disabling it and chunking trimul makes `sweetspot` 34.5% lower-peak than `optimized` at a 19.0% wall-time cost.

## Expected long-sequence consumption

With fixed triangle-attention and trimul caps, block scratch scales as O(CN) for cap C, while `U` scales as O(N²). The scratch contribution in U therefore decreases approximately as C/N at longer sequence lengths. The remaining pair-scaled tensors dominate.

The measured N=1,369 `sweetspot` peak was 2.70U above its 1.68 GiB resident baseline. We use 2.5U only as an asymptotic planning estimate for longer sequences, where fixed-cap O(CN) scratch is a smaller fraction of U. It is not a measured N=4,097 peak.

At N=4,097:

- `1U = 4,097² × 128 × 4 bytes = 8.00 GiB`;
- `2.5U + 1.68 GiB resident` is approximately 21.69 GiB; and
- carrying forward the measured 2.70U ratio gives approximately 23.30 GiB.

The successful end-to-end 4,097-token run on a 24 GiB RTX 4090 is consistent with this range, but did not record `max_memory_allocated`, so it does not prove the 2.5U estimate. It does show actual peak consumption remained below the card limit. The pair-bias cache must be disabled at this scale: its additional 3U would raise the pair-scaled estimate to approximately 5.5U, or 44 GiB before fixed storage.

The idealized 24 GiB ceiling is N≈4,327 under the 2.5U estimate and N≈4,163 if the measured 2.70U ratio does not decline. This is not a general OOM guarantee: atom count, MSA depth, template representation, sample count, allocator fragmentation, and other GPU allocations change the margin. The 4,097-token run is the direct validation point.

## Changes

- Add fused Triton kernels and inference dispatch for triangle attention, triangle multiplication, LN-linear, SwiGLU transition, diffusion attention, relative-position embedding, and coordinate-template projection.
- Add triangle/trimul chunk caps, in-place OPM, direct residual paths, early feature eviction, per-sample confidence, and diffusion caches.
- Add deterministic inference seeding and segmented atom-to-token reduction.
- Add parity, fallback, compile-reuse, and integration tests plus profiling and benchmark scripts.
- Document kernel controls and trade-offs in `docs/source/kernels.md`.

## Related issues

- Related to #71.
- Overlaps with memory-lifecycle work in draft PR #248.

## Parity

Full-model correctness parity was run on MCL1 (N=1,369; 10,201 atoms) with eight 158-residue protein chains, three ATP chains, and one additional small-molecule ligand. The featurized input contained 1,594 MSA rows over the 1,369-token axis and four coordinate-derived templates. The run used one diffusion sample, seed 42, deterministic algorithms, `CUBLAS_WORKSPACE_CONFIG=:4096:8`, and TF32 disabled. Each configuration was constructed separately with its own flags and received a fresh fixed-seed batch.

Against `local_default`, `optimized` had 4.23e-5 Å aligned all-atom RMSD and 1.16e-3 Å maximum atom displacement; `sweetspot` had 2.56e-5 Å aligned RMSD and 6.92e-4 Å maximum displacement. Sampled output-logit RMSEs were between 7.19e-6 and 3.86e-5. This agrees with the earlier deterministic N=1,053 MCL1 check, which measured 4.46e-5 Å aligned RMSD for `optimized` versus default.

We also repeated the same deterministic, fixed-seed MCL1 comparison with TF32 enabled. Against `local_default`, the aligned all-atom RMSD was 0.176 Å for `optimized` and 0.157 Å for `sweetspot`; maximum atom displacements were 2.83 Å and 2.85 Å, respectively. The larger difference is reproducible with identical inputs and random seeds and comes from different reduction orders in the fused LN-linear and flash-attention paths under TF32, rather than different diffusion samples.

## Limitations / TODO

- Current launch configurations were selected on an RTX 4090 (SM89). They do not yet provide per-architecture tuning coverage across NVIDIA GPU generations; Ampere, Hopper, and other devices require validation and configuration tuning.
- The Triton implementation targets AMD ROCm, but the new fused schedules in this PR still need parity and performance coverage across AMD GPU architectures.
- Strict ~1e-5 Å parity currently requires TF32 to be disabled. Fused LN-linear and flash attention use different reduction orders under TF32, producing 0.16–0.18 Å same-seed aligned RMSD in the MCL1 benchmark.
- Coordinate-derived templates are inference-only and currently require the default 3.25/50.75/39 distogram settings.
- Training and autograd continue to use the existing differentiable implementations.

## Test plan

- [x] See the deterministic and TF32 parity results above.


heathcliff233:
Hi @jandom , thank you for the effort in maintaining OpenFold3. Here is a google colab notebook [demo](https://colab.research.google.com/drive/1G2DmZiIiBUSqAHmJmj1xS02aZ-xND41Z?usp=sharing) of all the current inference optimizations. Current PR serves as the reference for later breakdown PRs including #320 #321 #322 . We are currently working on the rest part as well as backward implementation to further support training. Please consider reviewing PR 320-322 first which serve as the correctness test foundation. Thank you very much.

tomgoddard:
That is amazing that you can run a 4097 residue prediction on an Nvidia RTX 4090 with 24 GB. Could you provide an exact recipe (input json, runner yaml file, environment variable settings, ...) that reproduce that?

I get out of memory with your code (cloned Sept 18, 2026) using a 3700 residue dimer. I used these environment variables

export OPENFOLD3_FUSED_LN_LINEAR="1"
export OPENFOLD3_FUSED_SWIGLU_TRANSITION="1"
export OPENFOLD3_FUSED_TRIMUL="1"
export OPENFOLD3_FUSED_DIFFUSION_ATTN="1"
export OPENFOLD3_DIFFUSION_PAIR_BIAS_CACHE="0"
export OPENFOLD3_FUSED_TEMPLATE_EMBED="1"
export OPENFOLD3_INPLACE_OPM="1"
export OPENFOLD3_FUSED_TRI_ATTN_V1="1"
export OPENFOLD3_TRI_ATTN_CHUNK_CAP="128"

and capped MSA rows to 1024 but get out of memory quickly in

File "/home/goddard/openfold3_pr318/openfold3/projects/of3_all_atom/model.py", line 260, in run_trunk
z = torch.zeros_like(z_init)
^^^^^^^^^^^^^^^^^^^^^^^^
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 6.46 GiB. GPU 0 has a total capacity of 23.49 GiB of which 5.93 GiB is free. Including non-PyTorch memory, this process has 17.04 GiB memory in use. Of the allocated memory 16.57 GiB is allocated by PyTorch, and 15.62 MiB is reserved by PyTorch but unallocated.

tomgoddard:
The previous comment was mistakenly posted to issue #322.  @heathcliff233 responded as follows:

hi @tomgoddard , thanks for your interest. PR https://github.com/aqlaboratory/openfold-3/pull/318 is an experimental branch and have not yet been fully polished. I have just updated it to make the triton behavior a bit more clear, please update to the latest commit. And there are still a few things not fully supported as they require touching the pipeline logic. As of the settings, please check the following as a reference

```
  model_update:
    presets:
      - predict
      - low_mem
    custom:
      settings:
        memory:
          eval:
            use_triton_triangle_kernels: false
            msa_module:
              swiglu_seq_chunk_size: 64
              swiglu_chunk_token_cutoff: null
  data_module_args:
    num_workers: 0
  dataset_config_kwargs:
    msa:
      max_rows: 1024
      subsample_main: false
    template:
      use_coordinate_pair_features: true
      take_top_k: true
      n_templates: 4

```
```
export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
export OPENFOLD3_FUSED_LN_LINEAR=1
export OPENFOLD3_FUSED_SWIGLU_TRANSITION=1
export OPENFOLD3_FUSED_TRIMUL=1
export OPENFOLD3_FUSED_TRI_ATTN_V1=1
export OPENFOLD3_FUSED_DIFFUSION_ATTN=1
export OPENFOLD3_FUSED_TEMPLATE_EMBED=1
export OPENFOLD3_INPLACE_OPM=1
export OPENFOLD3_DIFFUSION_PAIR_BIAS_CACHE=0
export OPENFOLD3_TRI_ATTN_CHUNK_CAP=128
export OPENFOLD3_TRIMUL_CHUNK_CAP=128

```
I think the OOM error you encountered is due to this flag: use_coordinate_pair_features: true. The current settings should work for 4096 aa cases.

tomgoddard:
Hi  @heathcliff233, I tried a 3682 residue prediction (homodimer) with the settings you suggested in the preceding comment on an Nvidia RTX 4090 (24 GB) and it fails with an out of memory error in pairformer_embedding() and embed_zij() after 15 minutes.  It seems it got pretty far since this same prediction completes in about 20 minutes on an RTX Pro 6000 (96 GB).  MSAs and templates were precomputed.  I also tried with no templates and hit the same out-of-memory error.  I include the input and output in the attached zip file.  My attempt was on Ubuntu 24.04 with 64 GB of RAM.  I used your OpenFold 3 branch for this pull-request using today's code (sha bde6d612d10a4e35670a299a33299a79fa06552b, last updated Sept 19, 2026).

I'm still interested in seeing the exact setup of the 4097 token run you report in the description, including input json file, environment variables and runner.yaml file.

[pks13_oom.zip](https://github.com/user-attachments/files/32491353/pks13_oom.zip)

heathcliff233:
hi @tomgoddard , thank you for the feedback. I just tried your sample and it should work with --num_diffusion_samples 1. The OOM is due to the additional pair feature copy in the confidence part when num_diffusion_samples > 1. Previously I fixed that in PR #321 and will push a similar fix here later.

tomgoddard:
Hi #heathclif233, with num_diffusion_samples = 1 I was able to run several large predictions using your current PR #318 code.  The following PDB entries and residue counts succeeded on my Nvidia RTX 4090 (24 GB), Ubuntu 24.04 64 GB system:

  PDB 6l53, 3661 residues
  PDB 9yqq, 3777 residues
  PDB 5oej, 3825 residues
  PDB 9yrp, 3926 residues
  PDB 22te, 3974 residues
  PDB 8e05, 4032 residues

and this one failed with out of memory

  PDB 9mfv, 4092 residues

So it is now performing close to your 4097 residue case.  That is amazing.  If I recall AlphaFold 3 on this computer only handled about 2500-3000 residues.  I develop ChimeraX and help biologists run OpenFold 3 on their low-power Mac, Windows and Linux computers.  I was sad to see that Triton does not produce fused kernels for Torch MPS on Mac.

jandom:
Thanks for the evaluation @tomgoddard – @heathcliff233 we're moving ahead with reviewing this, Jennifer will be leading the charge.

PR #280, opened 2026-06-25
ci: support custom AMD runners

jandom:
**Summary**

I had 10 minutes to spare so decided to tackle this little one: hook up to the AMD runners, self-hosted

**Changes**

Changed the integration workflows, split into CUDA and ROCM

**Related Issues**

**Testing**

@jnwei this will be challenging to test since GHA doesn't like to run actions that haven't been merged into main, I'll just do manual trigger

- Attempt 1: https://github.com/aqlaboratory/openfold-3/actions/runs/28187827468 no ability to pull code from github
- Attempt 2: https://github.com/aqlaboratory/openfold-3/actions/runs/28850216162 bad device passthrough
- Attempt 3: https://github.com/aqlaboratory/openfold-3/actions/runs/28859047319/ "The SSL connection could not be established, see inner exception."
- Attempt 4: https://github.com/aqlaboratory/openfold-3/actions/runs/28859047319/job/87393397578, fails with

```
FAILED openfold3/tests/test_kernels.py::TestKernels::test_triton_forward_bf16 - RuntimeError: Failed to find C compiler. Please specify via CC environment variable or set triton.knobs.build.impl.

```

- Attempt 5: https://github.com/aqlaboratory/openfold-3/actions/runs/29516745841

**Other Notes**

PR #319, opened 2026-07-17
feat: add openfold3-therock env for AMD APU / Strix Halo (gfx1151) support

rfilgueiras:
## What

Adds an `openfold3-therock` pixi environment that provides ROCm PyTorch +
Triton via AMD's [TheRock](https://github.com/ROCm/TheRock) nightly multi-arch
index, enabling AMD APUs — notably Strix Halo (Ryzen AI Max+ 395 / Radeon
8060S, `gfx1151`).

## Why

The existing `openfold3-rocm7` env uses the official `rocm7.2` wheels, which
do not cover `gfx1151`. In TheRock, `gfx1151` is release-ready
(SUPPORTED_GPUS.md). This slots into the existing Triton kernel path — no code
changes, only a new pixi feature/env + lock and docs.

## Verification

`validate-openfold3-rocm` passes on real gfx1151 hardware, including the
Triton evoformer kernel load:

- torch 2.14.0a0+rocm7.15.0a..., HIP 7.15.0
- GPU visible: AMD Radeon 8060S Graphics
- Triton backend HIP, evoformer kernel loaded

End-to-end inference also verified on the same hardware — `run_openfold
predict` with the `triton.yml` preset on ubiquitin (76 residues):

- Successful queries: 1, failed: 0 (~32s forward pass)
- 5 diffusion samples, valid CIF output (601 atoms each) + confidence JSONs
- Triton kernels only (DeepSpeed / cuEquivariance disabled)

One benign warning is emitted: `CK GEMM on an unsupported architecture!
Cannot set blas backend to CK` — ROCm's Composable Kernel GEMM isn't tuned for
gfx1151, so it falls back to rocBLAS. Output is correct; this is a perf note,
not an error.

## Caveats

- **Nightly wheels.** The TheRock multi-arch wheel index currently publishes
  only prerelease (`a<date>`) builds; the exact version is pinned in
  `pixi.lock`. `openfold3-rocm7` stays as the stable path.
- **Version context.** TheRock's 7.9–7.13 preview train is complete; ROCm 7.14
  is AMD's first production TheRock release and lists gfx1151 as a supported
  arch. However, 7.14 currently ships as OS packages / tarballs, not as a
  pinnable PyTorch wheel — so a stable-wheel repin is a **follow-up** once one
  is published. Until then the pinned nightly is the working path.
- Consider gating any CI for this env as experimental, given nightly drift.
- linux-64 only; Python capped `<3.14` (no cp314 wheels yet).

## Note on the `pixi.lock` diff

The lock diff is large (~1800 lines) but **additive only**. `pixi.lock` is
shared by all environments, so adding one env re-serializes the file and
reorders existing package references — hence the noisy removed/added lines.

Verified no existing environment is affected:
- No packages removed (unique packages 1412 → 1489; the +77 are all for
  `openfold3-therock`).
- Environment list is unchanged plus the one new `openfold3-therock`.
- Existing envs (`cpu`, `cuda12/13`, `*-pypi`, `msa`, `rocm7`) resolve to the
  same packages/versions/hashes as before — no re-solve, no drift.

## Note / questions (first contribution)

This is my first contribution here, opened as a **draft** to get direction
before polishing. I'd appreciate guidance on:

1. **Is TheRock the approach you want** for APU support, or would you prefer to
   wait for gfx1151 in the official PyTorch ROCm wheels?
2. **How should the nightly be handled in CI** — skip it, run it non-blocking,
   or add a self-hosted gfx1151 runner (mirroring the existing amd-gpu setup in
   #280)?
3. Any preference on **where the docs live** (I put them in `inference.md` +
   `Installation.md`) and on the `pixi.lock` diff size for a nightly env?

Happy to adjust naming, structure, or split docs/config into separate PRs.

---

*Note: this work was done with AI-assisted tooling; all changes were
reviewed and verified on real gfx1151 hardware before submitting.*

### DIFF
--- a/docs/source/Installation.md
+++ b/docs/source/Installation.md
@@ -109,6 +109,8 @@ For AMD system installation: After installation, verify your ROCm environment is
 validate-openfold3-rocm
 ```
 
+> **AMD APUs / Strix Halo (`gfx1151`)**: the `rocm7.2` wheels above don't cover APUs. Use the `openfold3-therock` pixi environment instead (AMD TheRock nightly). See the [AMD ROCm inference docs](inference.md).
+
 (installation-environment-variables)=
 ### Environment variables
 
--- a/docs/source/Installation.md
+++ b/docs/source/Installation.md
@@ -109,7 +109,7 @@ For AMD system installation: After installation, verify your ROCm environment is
 validate-openfold3-rocm
 ```
 
-> **AMD APUs / Strix Halo (`gfx1151`)**: the `rocm7.2` wheels above don't cover APUs. Use the `openfold3-therock` pixi environment instead (AMD TheRock nightly). See the [AMD ROCm inference docs](inference.md).
+> **AMD APUs / Strix Halo (`gfx1151`)**: the `rocm7.2` wheels above don't cover APUs. Use the `openfold3-therock` pixi environment instead (AMD TheRock, ROCm 7.14). See the [AMD ROCm inference docs](inference.md).
 
 (installation-environment-variables)=
 ### Environment variables
--- a/docs/source/inference.md
+++ b/docs/source/inference.md
@@ -327,6 +327,22 @@ run_openfold predict \
 
 > **Note on first-run compilation**: Triton JIT-compiles kernels on first use and caches them to `~/.triton/cache`. The compilation is a one-time cost per unique sequence length per machine; subsequent runs at the same length incur no overhead.
 
+##### AMD APUs / Strix Halo (gfx1151) via TheRock nightly
+
+The default `openfold3-rocm7` environment uses the official `rocm7.2` PyTorch wheels, which do **not** cover AMD APUs such as Strix Halo (Ryzen AI Max+ 395, Radeon 8060S, `gfx1151`). For those, use the `openfold3-therock` pixi environment, which pulls PyTorch + Triton from AMD's [TheRock](https://github.com/ROCm/TheRock) nightly multi-arch index (`gfx1151` is release-ready per TheRock's `SUPPORTED_GPUS.md`).
+
+```bash
+pixi run -e openfold3-therock validate-openfold3-rocm
+```
+
+All Triton settings above apply unchanged; TheRock's Triton JITs to HIP for `gfx1151`.
+
+> **Version note**: TheRock version strings like `rocm7.13.0a...` / `rocm7.15.0a...` are **nightly build snapshots**, not released ROCm versions (the official line is 7.0 → 7.1 → 7.2). A higher number just means a more recent nightly, not a more qualified release. The exact build is pinned in `pixi.lock` for reproducibility; `pixi update` will move it. Keep `openfold3-rocm7` as the stable fallback.
+>
+> **Host requirement**: only the `amdgpu` kernel driver (`/dev/kfd`, `/dev/dri`) is needed. The ROCm userspace ships inside the torch wheel — no system `/opt/rocm` install required.
+
+For a different AMD GPU, change the `device-gfx1151` extra in the `pytorch-pypi-therock` feature (`pixi.toml`) to your target from TheRock's device table (e.g. `device-gfx1201`).
+
 ---
 
 (inference-cif-direct-templates)=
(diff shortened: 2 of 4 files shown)


jnwei:
Thank you @rfilgueiras for the contribution and the questions!

@sdvillal @singagan do you have any thoughts about the best way to approach `gfx1151` support and the best way to organize AMD environments?

sdvillal:
Really nice contribution @rfilgueiras! Fully understand the convenience of officially suporting gfx1151.

I would need to look deeper into this. I have some concerns:

- I think supporting another non-pypi index with nightly releases is risky, from multiple angles. These urls are usually capped at corporate networks and we might unleash deep problems for people working behind firewalls. Nightly by definition means quickly changing indices and more bugs, so if we add something like this we should try as much as possible to freeze that environment (e.g., pinning versions and suggesting to use frozen environments in the docs).

- Our pixi workspace is becoming unruly already. We should follow-up with the prefix people to see how their plans on providing modularity and optionality on workspaces are coming along.

@singagan by chance do you know if there are plans to publish wheels supporting gfx1151 in the in the short term?

In any case, my preferred solution would be to help having a healthy and up-to-date rocm ecosystem in conda-forge, mostly pure-conda environments, and document recipes for users to get into the bleeding edge. The first still [feels still far](https://github.com/conda-forge/conda-forge.github.io/issues/1923#issuecomment-3339551001), even if recently good progress was made.

sdvillal:
btw, about the conda ecosystem, maybe we could try these channels:
https://github.com/gbionics/rock-the-conda

If interesting, you could give it a try in this PR too @rfilgueiras. Unfortunately I do not have access to AMD hardware so it is difficult for me to play with this.

rfilgueiras:
Following the feedback on nightlies: I looked into ROCm 7.14, which AMD announced (2026-07-15) as their new production release built on TheRock.

Findings for gfx1151:
- 7.14 is production and officially supports gfx1151, but the wheels are still **only** on TheRock's multi-arch index — no PyPI/conda-forge/stable wheel exists yet, and every build is prerelease-tagged (`+rocmX.Y.0a<date>`).
- So I can't remove the non-pypi index, but I've addressed the "freeze it" concern: repinned from the moving nightly to a **fixed, dated 7.14 build** (`torch 2.12.0+rocm7.14.0a20260624`).

Verified on my gfx1151 (Radeon 8060S): validator passes and ubiquitin inference runs end-to-end (601 atoms, avg_plddt 79.4).

The intent is exactly this: an opt-in developer env so people with this hardware can test and validate the prerelease 7.14 path now, ahead of a stable wheel landing — not a CI/production dependency. Given that, keeping it as one isolated, additive `openfold3-therock` feature seems reasonable for now, but happy to hold for a broader workspace cleanup if you'd prefer.

rfilgueiras:
Important correction after more testing on my side.

I re-tested `gfx1151` (Radeon 8060S) on the **existing** `openfold3-rocm7` env — the stable `torch 2.12.1+rocm7.2` wheels from PyPI (`download.pytorch.org/whl/rocm7.2`), no TheRock index involved. It runs OpenFold3 end-to-end across every query type I tried:

| Case | Result | Notes |
| --- | --- | --- |
| ubiquitin (monomer) | ✅ 1/1 | avg_plddt 79.4 |
| multimer (4-chain, 7cnx) | ✅ 1/1 | avg_plddt 90.6, iptm 0.85 |
| protein–ligand (mcl1, 8 chains) | ✅ 1/1 | ligand handled |
| DNA + PTM | ✅ 1/1 | valid |

Validator passes (incl. Triton evoformer kernel on HIP), no HIP/kernel errors, correct structured output.

So my original premise for this PR was wrong: I had assumed the `rocm7.2` wheels did not cover `gfx1151`, but they do. Given that the stable env already supports this hardware, the TheRock env + non-pypi prerelease index does not seem justified — it would add the firewall/nightly risk you flagged to solve a problem that is already solved.

I think this PR should be closed. Would it be helpful if I instead open a small **docs-only** PR noting that `gfx1151` works out of the box on `openfold3-rocm7` (and how to verify it)? Happy to do that if useful, otherwise I will just close this.

rfilgueiras:
Closing for now. I could validate that it is working on ROCm 7.2 (not officially documented, but in practice not a blocker)

PR #324, opened 2026-07-20
Pocket constraints

peter-s-obi:
**Summary**

Adds query-level ligand pocket constraints for inference. A user can specify a ligand chain and a set of pocket residues, and OpenFold-3(OF3) will bias ligand placement toward the chosen site.

Implementation strategy:

- Run the standard diffusion rollout once to obtain protein and ligand proposals
- Generate ligand seed poses inside the user-specified pocket
- Restart diffusion from the selected pocket-centered seeds 75% of the way through the diffusion schedule (default), allowing the model to refine the ligand and local environment. In the benchmark runs, earlier restart points allowed the ligand to return to the model-preferred orthosteric site.

**Changes**

- Adds an optional `pocket_constraint` field at the query level
- Adds featurization for ligand and pocket atom masks, pocket contact distance, VDW radii, and optional ligand conformer proposals.
- Uses OF3 processed reference molecules for conformer generation, so conformers are derived from the same ligand representation used by the model for both SMILES and CCD ligands.
- Adds pocket-seeded partial diffusion in `SampleDiffusion`.
- Keeps pocket proposal defaults centralized in `openfold3/core/config/pocket_sampling_defaults.py`.

**Related Issues**
None.

**Testing**
All 44 pocket-constraint unit tests passed.

Benchmark testing: Tested 8 allosteric ligand-placement cases from 
> Nittinger et al., “Co-folding, the future of docking – prediction of allosteric and orthosteric ligands,” Artificial Intelligence in the Life Sciences, 2025.

where unconstrained OF3 places the ligand in the orthosteric site. Each run used 16 samples with MSA enabled.

Allosteric ligand COM distance, best-of-16:

| PDB | Default OF3 | Pocket-constrained OF3 |
|---|---:|---:|
| CHEK1_3F9N_38M | 19.34 Å | 0.46 Å |
| ABL1_3PYY_3YY | 24.74 Å | 0.22 Å |
| Abl1_3K5V_STJ | 24.83 Å | 1.03 Å |
| AR_2YHD_AV6 | 18.53 Å | 0.17 Å |
| AR_2YLO_YLO | 27.16 Å | 1.45 Å |
| bla_1PZP_FTA | 20.98 Å | 0.63 Å |
| CSNK2A1_6FVF_503 | 15.25 Å | 0.54 Å |
| PTK2_4EBV_0O7 | 15.08 Å | 0.67 Å |

COM distance evaluates localization within the requested pocket; it does not measure native ligand pose accuracy.

**Other Notes**
- This is not learned-conditioning. After the initial rollout, ligand proposals are moved into the requested pocket, re-noised, and refined through standard denoising.
- The current query schema supports one ligand pocket constraint per query.
- Environment variables expose expert overrides; normal use requires only the query-level `pocket_constraint` field.
- On an A100 40 GB GPU, increasing from 16 to 32 samples did not materially improve best pocket COM, while 64 samples was OOM.


jandom:
Stellar, thank you @peter-s-obi !!

jandom:
@peter-s-obi any chance you could add me and @jnwei as collaborators to your forked repo? Otherwise we can't push to this branch with any changes – we have to stack another PR or mirror your branch (eh)

peter-s-obi:
Done. Let me know if the invite doesn't come through @jandom @jnwei

openfold3/core/data/pipelines/featurization/pocket_constraints.py
```diff
@@ -0,0 +1,278 @@
...
+    "N": 1.55,
+    "O": 1.52,
+    "F": 1.47,
+    "P": 1.80,
+    "S": 1.80,
+    "CL": 1.75,
+    "BR": 1.85,
+    "I": 1.98,
+}
+DEFAULT_VDW_RADIUS = VDW_RADII["C"]
```
jnwei, line 31: How common do you expect users to need to change the default VDW radii?

We can add another Pydantic model to support adjustements to this set if you think it will be a common feature to tune.

openfold3/core/model/structure/pocket_constraints.py
```diff
@@ -0,0 +1,262 @@
...
+    pose: torch.Tensor
+    parent_slot: int
+
+
+def _candidate_sort_key(
+    candidate: _PocketSamplingCandidate,
+) -> tuple[float, float, float, float, float]:
+    return (
+        float(candidate.pocket_com_dist),
+        -float(candidate.lig_atoms_in_pocket),
```
jnwei, line 103: Request for a double check that the signs here are correct?

 I'm not an expert, but I find that the score being inversely related to lig_atoms_in_pocket a bit surprising.

openfold3/tests/test_pocket_constraints.py
```diff
@@ -0,0 +1,765 @@
...
+        super().__init__()
+        self.calls = 0
+
+    def forward(self, *, xl_noisy, **_kwargs):
+        self.calls += 1
+        return xl_noisy
+
+
+def _pocket_sampling_batch(batch_dim: int = 1) -> dict[str, torch.Tensor]:
+    # Create a batch with 0 jitter to test guided diffusion sampling
```
jnwei: I added this docstring, but I'm not sure if this is the main motivation for this example or not. We could also change the function itself to be `pocket_sampling_batch_without_jitter`

openfold3/tests/test_pocket_constraints.py
```diff
@@ -0,0 +1,765 @@
...
+        atom_mask=batch["atom_mask"],
+        no_rollout_samples=2,
+    )
+
+    ligand = seeds[0, :, 3:5]
+    ligand_com = ligand.mean(dim=1)
+    ligand_distance = torch.linalg.vector_norm(ligand[:, 0] - ligand[:, 1], dim=-1)
+
+    assert torch.all(torch.linalg.vector_norm(ligand_com, dim=-1) < 1e-5), "ligand COM should be at origin"
+    assert torch.allclose(ligand_distance, torch.full((2,), 4.0), atol=1e-5)
```
jnwei: Could you expand a bit more on why this torch.full((2,0, 4.0) is the expected ligand distance?

It would also help to add a failure message, like the one we have on the line above.

openfold3/tests/test_pocket_constraints.py
```diff
@@ -0,0 +1,765 @@
...
+    xl_base[:, :, 4] = torch.tensor([0.2, 0.0, 0.0])
+
+    seeds = _build_pocket_sampling_seeds(
+        batch=batch,
+        xl_base=xl_base,
+        atom_mask=batch["atom_mask"],
+        no_rollout_samples=2,
+    )
+
+    assert seeds.shape == (1, 2, 5, 3)
```
jnwei: A little explanation here of why the expected value is what it is would be helpful. 

Is the expected seeds.shape intended to match xl_base?

openfold3/tests/test_pocket_constraints.py
```diff
@@ -0,0 +1,765 @@
...
+            batch=batch,
+            si_input=torch.zeros(1, 1, 1),
+            si_trunk=torch.zeros(1, 1, 1),
+            zij_trunk=torch.zeros(1, 1, 1, 1),
+            noise_schedule=torch.tensor([1.0, 0.5, 0.1]),
+            no_rollout_samples=2,
+        )
+    # The first (unconstrained) rollout already ran before the missing
+    # feature was discovered while building the pocket-sampling seeds.
+    assert sampler.diffusion_module.calls == 2
```
jnwei, line 820: Because I removed the validation tests on the pocket_sampling batch that @peter-s-obi added previously, this error is now a KeyError and the diffusion module would be called a twice before an error arose.

I had removed the validation tests on the pocket_sampling because I assumed that a valid set of features should always be given by the upstream pocket feature generation in the dataset.

However, @peter-s-obi if you prefer that the inference run fail immediately before running any diffusion steps, I can restore the validations and this test.

jnwei (changes requested):
@peter-s-obi I have now taken a pass and refactored the configuration settings to use pydantic models. 

Two requests:
- Could you please review my changes and see that these changes look good to you? I have some specific questions / suggestions that I have placed in comments below
- Could you provide an inference query example that uses pocket constraints? One of the examples you used in the table in the PR description would be perfect. We would add this examples to our [examples directory](https://github.com/aqlaboratory/openfold-3/tree/main/examples/example_inference_inputs), and write some scientific tests around this example.

openfold3/core/model/structure/pocket_constraints.py
peter-s-obi: Yes, the negative sign is intentional. Candidate tuples are sorted ascending, but a larger lig_atoms_in_pocket count is preferred, so negation converts it into an ascending sort. For example, if a ligand has 12 vs 5 atoms in the pocket, negation turns it into -12 so it sorts first. However for the COM distance, lower is better so we sort as it is. Maybe needs a short comment to clarify?

openfold3/core/data/pipelines/featurization/pocket_constraints.py
peter-s-obi: I wouldn’t expect users to change these often, so I don’t think we need another configuration model.

I also noticed OF3 already defines `PERIODIC_TABLE = Chem.GetPeriodicTable()`  from `core/data/primitives/structure/component.py`. We could reuse that here instead of maintaining a separate radii table. I'm happy to make that change

openfold3/tests/test_pocket_constraints.py
peter-s-obi: Yes, that's right. Zero jitter keeps the shared fixture deterministic without random offsets. However the helper function is used for proposal generation etc. So maybe  _Create a deterministic pocket-sampling batch with jitter disabled_ captures the scope more.

But this made me think, do we need to test nonzero jitter behavior?

openfold3/tests/test_pocket_constraints.py
peter-s-obi: The input conformer places its two atoms at -2 and +2 on the x-axis, giving a 4 Å separation. Since proposal generation only rotates and translates the conformer, both generated samples should preserve that distance. 

I’ll derive the expected value from the input conformer and add a failure message to make this clearer.

openfold3/tests/test_pocket_constraints.py
peter-s-obi: Yes, the output is intended to match xl_base: each generated seed copies a complete parent structure and replaces only its ligand coordinates. 

I’ll compare directly against xl_base.shape and add a failure message to make that contract explicit.

openfold3/tests/test_pocket_constraints.py
peter-s-obi: Makes sense to me. I don’t think we need to restore the validation, since the supported inference path always receives the complete feature set from the upstream pocket featurizer. KeyError is acceptable.

openfold3/core/model/structure/pocket_constraints.py
jnwei: Thanks for the explanation, that makes sense if we expect lig_atoms_in_pocket to be a list of atoms. A short comment, either here or in the _PocketSamplingCandidate description would be great.

openfold3/tests/test_pocket_constraints.py
jnwei: - Yes, changing the helper function name would be great here, so that it becomes obvious in line what the expectations are for the sample pocket
- It would also be great to test nonzero jitter behavior, but are there any hard guarantees we can make regarding the expected behavior? Maybe something about the average noise / RMSD of the new samples?

jnwei (commented):
Awesome, thank you so much @peter-s-obi 

I have run the example you have provided and confirmed that I get significantly different structures. This is a perfect test case for us to use for our regression tests.

openfold3/core/model/structure/pocket_constraints.py
peter-s-obi: Added a comment

openfold3/tests/test_pocket_constraints.py
peter-s-obi: I renamed the helper to make the zero-jitter setup explicit.

For nonzero jitter, the clearest hard guarantee I had previously tested is that each ligand proposal receives an independent rigid translation. Added a deterministic test that controls the random draw and verifies that all ligand atoms move together, while protein coordinates and ligand internal geometry remain unchanged by the jitter operation itself.

RMSD would be tricky because of the subsequent re-noising and denoising that can confound the values.

openfold3/tests/inference/test_pocket_constraints.py
```diff
@@ -0,0 +1,197 @@
...
+    assert off_com.best > POCKET_SAMPLING_OFF_COM_MIN_ANGSTROM, (
+        f"{QUERY_NAME}: unguided sampling unexpectedly localized the ligand — best-of-"
+        f"{NUM_DIFFUSION_SAMPLES} COM distance {off_com.best:.2f} Å is below the "
+        f"{POCKET_SAMPLING_OFF_COM_MIN_ANGSTROM} Å floor, i.e. pocket guidance may "
+        "no longer be doing anything"
+    )
+
+
+if __name__ == "__main__":
+    raise SystemExit(pytest.main([__file__, "-vv"]))
```
jandom: I don't think this is really needed

openfold3/tests/inference/test_pocket_constraints.py
```diff
@@ -0,0 +1,197 @@
...
+#   enabled  COM per-sample 0.34, 2.85, 3.23, 3.59, 3.67 -> best 0.34, mean 2.73
+#   disabled COM per-sample 21.29, 21.30, 22.26, 22.54, 30.05 -> best 21.29, mean 23.48
+# matching the PR's own best-of-16 numbers for this case (0.63 A / 20.98 A) closely
+# enough to confirm the reference and chain mapping above are right.
+# Using 5 samples in this test instead of 16 to reduce compute cost
+POCKET_SAMPLING_ON_COM_MAX_ANGSTROM = 5.0
+POCKET_SAMPLING_OFF_COM_MIN_ANGSTROM = 10.0
+
+
+def _run(*, pocket_sampling_enabled: bool, tmp_path: Path) -> list[Path]:
```
jandom: nit: I'm not a huge fan of those bools arguments, Claude really like this, instead I prefer to inject the behavior (the extra YAML lines, or construct the YAML in two different ways – on/off –  at the parameterize stage)

jandom (approved):
Just nits, approved – this is really solid, thank you @jnwei

openfold3/tests/inference/test_pocket_constraints.py
jnwei: removed

openfold3/tests/inference/test_pocket_constraints.py
jnwei: refactored, thanks!

jnwei (approved):

PR #328, opened 2026-07-21
Fix stale mask_bias shape

balvisio:
**Summary**
Fixes a stale comment on the mask shape.

**Testing**
N/A


jandom:
thanks @balvisio, very grateful for your contribution – do you know if we also need to update the docstring above, or is that fine?

balvisio:
thanks @jandom . I took a closer look and actually there were a few other stale docstring and references. Just updated.

jandom:
@balvisio this looks great, thanks so much for this contribution!!

jandom (approved):

PR #329, opened 2026-07-21
Optimize the handling of missing version_tensor

bluecake-dh:
Previously, the implementation of _warn_on_missing_version_tensor_in_load_statedict would potentially call load_state_dict twice if version_tensor was missing, causing unnecessary overhead. The new implementation guarantees that load_state_dict only be executed once.

Details of change:
load_state_dict is called once with strict=False. Then, manual checks are performed based on the missing/unexpected keys returned, and the appropriate action (continue directly, issue warning, or raise exception) is performed accordingly. Previously, the appropriate action was determined off of exception catching and exception message parsing, which was slower and less reliable.

### DIFF
--- a/openfold3/entry_points/experiment_runner.py
+++ b/openfold3/entry_points/experiment_runner.py
@@ -668,18 +668,27 @@ class InferenceExperimentRunner(ExperimentRunner):
         self, state_dict: dict
     ) -> None:
         """Load state dict, warning if only version_tensor is missing."""
-        try:
-            self.lightning_module.load_state_dict(state_dict, strict=True)
-        except RuntimeError as e:
-            if 'Missing key(s) in state_dict: "model.version_tensor".' in str(e):
-                logger.warning(
-                    "No version_tensor is found for this checkpoint."
-                    "Assuming the user knows checkpoints are parameters are compatible,"
-                    " continuing..."
-                )
-                self.lightning_module.load_state_dict(state_dict, strict=False)
-            else:
-                raise
+        missing_keys, unexpected_keys = self.lightning_module.load_state_dict(state_dict, strict=False)
+
+        # No issues with loaded state dict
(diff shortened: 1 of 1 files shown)


openfold3/entry_points/experiment_runner.py
```diff
@@ -668,18 +668,27 @@ def _warn_on_missing_version_tensor_in_load_statedict(
...
+        if not missing_keys and not unexpected_keys:
+            return
+
+        # Only version_tensor is missing
+        if missing_keys == ["model.version_tensor"] and not unexpected_keys:
+            logger.warning(
+                "No version_tensor is found for this checkpoint. "
+                "Assuming the user knows the given checkpoint parameters are compatible with the model, "
+                "continuing..."
+            )
```
jnwei, line 683:
```
       # perform the key check manually.
       model_keys = set(self.lightning_module.state_dict().keys())
        ckpt_keys = set(state_dict.keys())
        missing = model_keys - ckpt_keys
        unexpected = ckpt_keys - model_keys

        if missing == {"model.version_tensor"} and not unexpected:
            logger.warning(
                "No version_tensor found for this checkpoint. "
                "Assuming the user knows the given checkpoint parameters are compatible"
                " with the model, continuing..."
            )
            self.lightning_module.load_state_dict(state_dict, strict=False)
            return

        self.lightning_module.load_state_dict(state_dict, strict=True)
```

jnwei (changes requested):
Hey @bluecake-dh thank you for the contribution and proposed fix. Sorry for the hassle you've been experiencing with the version tensor check.

We generally prefer to make `strict=True` the default, following [pytorch lightning best practices](https://lightning.ai/docs/pytorch/stable/common/checkpointing_intermediate.html#resume-from-a-partial-checkpoint). I have a suggested fix that will help avoid the double load, and also avoid having `strict=False` be the central load behavior

PR #332, opened 2026-07-22
docs: build testing updates

jandom:
**Summary**

Small changes to our build test guide

**Changes**

**Related Issues**

**Testing**

**Other Notes**


docs/BUILD_TESTING.md
```diff
@@ -42,18 +41,18 @@ wrong dependency pins — that `python -m build` alone won't surface.
 
 ```bash
 mkdir test-sdist && cd test-sdist
-uv venv
+uv venv --python 3.14
 source .venv/bin/activate
 
-uv pip install ../dist/openfold3-<version>.tar.gz[dev]
+uv pip install "../dist/openfold3-<version>.tar.gz[dev,deepspeed]" "torch==2.10.0"
```
jnwei: why does torch need a separate install command beyond what's in the pyproject.toml dependency  list?

.gitignore
```diff
@@ -23,3 +23,6 @@ cutlass/
 
 # uv environments
 .venv/
+
+# build testing
+test-sdist/
```
jnwei: This is a generally harmless change. 

In general I typically place my test-sdist in a directory outside of openfold-3, so it's easy to inspect that i'm using the new pypi build and not a pixi / local path.

jnwei (changes requested):

docs/BUILD_TESTING.md
jandom: Fair point, i was blind-sided by pixi pinning to one version and then UV resolving another

jnwei (approved):

PR #336, opened 2026-07-23
Add Apple Silicon (MPS) inference support

### LINKED ISSUE #279: Support MacOS inference
With my bioinf knowledge from Sepp Hochreiter's JKU BioInf (now ML) institute,.. I want to do some dry-lab experimentation,...

I own a  
256GB Xeon 3435X with 2x 32GB RTX Pro 4500 Blackwell and 2x RTX 4000 Ada
128GB M5 Max MBP

Trying to install openfold-3-mlx, 1st thing I stumbled across was that install.sh is a simple wrapper / obfuscator around creating a venv, a real PITA when hooked on Anaconda.  
Took a while to realize that my shiny new MBP comes with an ancient Python 3.9 and the venv creation didn't account for that.  
Next, the venv is also hard-coded in predict.sh and predict's error handling logic is dubious...  
With Mac idiosyncrasies the code (Conda env) runs differently in the macOS vs VS Code terminal. (click module,...)  
...

Turning here to hopefully better things.  
Cheers  
G.

weitzner:
**Summary**
Ports OpenFold3 inference to run on Apple Silicon via PyTorch's MPS backend, and adds an `openfold3-mps` pixi environment. The only CUDA-only paths were hardcoded device strings, a few unconditional `torch.cuda.empty_cache()` calls, and one real memory-management bug that's specifically fatal on MPS.

`accelerator: gpu`/`auto` already resolve to MPS on a Mac in PyTorch Lightning, but several assumptions baked into the forward path meant MPS either silently no-op'd or reliably crashed on larger queries. This fixes those paths with zero behavior change on CUDA/AMD/CPU (see Testing).

**Changes**
- `device_utils.py`: `autocast_device_type()`/`empty_device_cache()` helpers, swapped in at every hardcoded-`"cuda"` autocast/empty_cache call site in the forward path.
- Root-cause fix for a reliable MPS crash on large multi-chain + ligand queries (e.g. `query_protein_ligand`): PyTorch's MPS caching allocator holds freed-but-cached memory as wired, competing with physical RAM the way CUDA's cache competes with VRAM. `TemplatePairStack` had no cache-release mechanism at all, and `msa_module` only had it in the offload path. Fixed via an opt-in `clear_cache_between_blocks` flag (off by default everywhere).
- A new `mps` preset (`model_setting_presets.yml`) for MPS-specific defaults (cache-clearing on, `msa_module` offload off, Triton disabled), auto-applied based on whether MPS will *actually* run. Covers Lightning's default `"gpu"`/`"auto"` resolution, not just the literal string `"mps"`.
- Fixed `use_triton_triangle_kernels=True` raising a bare `NameError` when
  Triton isn't installed (no Triton on mac!). Now raises the same `RuntimeError` `Attention.forward` already raises for the same situation.
- `openfold3-mps` pixi environment. It's a same-solve-group alias for`openfold3-cpu` (`"cpu"` is misleading on Apple
  Silicon, esp because the version of PyTorch that gets pulled in supports MPS out of the box).
- Updated docs: `Installation.md`'s environment table, and a new "Apple Silicon(MPS) Inference" section in `inference.md`.
- Removed `examples/example_runner_yamls/osx.yaml`: its one setting became redundant once `use_deepspeed_evo_attention` defaulted to `False` for everyone (unrelated prior commit), and the new `mps` preset makes it doubly unnecessary.

**Related Issues**
None.

**Testing**
- Every modality tested (protein, DNA, RNA, homomer, multimer, protein-ligand) yielded correct output and ~3.4–4.8x speedup over CPU.
- The specific crashing case identified now has stable memory post-fix, zero failures.
- `ruff check` / `ruff format --check` clean.
- Full `openfold3/tests/*` suite: 709 passed, 50 skipped (CUDA/ROCm/DeepSpeed-gated, expected on my hardware).
- New unit tests for the MPS-resolution logic, the `mps` preset's `clear_cache_between_blocks` behavior, and the Triton `RuntimeError`.
- CUDA/AMD/CPU-only behavior verified unchanged: the autocast/empty_cache changes are unchanged or no-ops for those backends and the new cache-clearing behavior is opt-in and off by default everywhere except the `mps` preset.

**Other Notes**
While fixing the MPS-specific Triton gap, found the same underlying issue also affects default CPU-only inference (`openfold3-cpu` has no Triton installed, and there's no `cpu` preset disabling it). Left out of scope here since it's unrelated to MPS.


jnwei:
This is awesome @weitzner ! Thank you for this contribution, I am very excited to test this out.

openfold3/core/utils/device_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    `offload_inference` frees a tensor by moving it off `device` with
+    `.cpu()`; the accelerator's allocator still holds the vacated memory in
+    its own cache until this is called. No-op for `cpu` (no allocator cache
+    to release) and for any accelerator without a matching `empty_cache`.
+    """
+    if device.type == "cuda":
+        torch.cuda.empty_cache()
+    elif device.type == "mps":
+        torch.mps.empty_cache()
```
jandom, line 44: I feel like an else is missing here, ideally one that gives something insightful (otherwise this is just a pass-through): either a log "hey you're clearning cache but you're device is really odd" or just error

openfold3/projects/of3_all_atom/config/model_setting_presets.yml
```diff
@@ -32,3 +32,28 @@ low_mem:
...
+# release that CUDA doesn't require as strictly.
+mps:
+  settings:
+    memory:
+      eval:
+        # No Triton on macOS/Apple Silicon so the base `True` default
+        # (model_config.py) always fails.
+        use_triton_triangle_kernels: false
+        offload_inference:
+          msa_module: false
```
jandom, line 50: This is probably fine

pixi.toml
```diff
@@ -110,7 +110,18 @@ openfold3-cpu = { no-default-feature = true, features = [
...
+# the same locked resolution). Exists because "cpu" is misleading on Apple
+# Silicon: the same conda-forge pytorch dependency always ships Metal (MPS)
+# support, unlike on Linux where it's truly CPU-only.
+openfold3-mps = { no-default-feature = true, features = [
+    "python",
+    "openfold3-cpu",
+    "pytorch-conda",
+    "openfold3-editable",
+    "tests"
+], solve-group = "openfold3-cpu" }
```
jandom: Ah, it's a shame we can't add an alias in pixi (to the openfold3-cpu env), this will just drift. Minor

jandom (commented):
Overall pretty positive, there are still places where raw "cuda" lurks – some of them are probably fine but this spot is probably not 

openfold3/tests/conftest.py

```python
_DEVICES = [
    pytest.param("cpu", id="cpu"),
    pytest.param(
        "cuda",
        id="cuda",
        marks=pytest.mark.skipif(
            not torch.cuda.is_available(), reason="CUDA not available"
        ),
    ),
]
```

Here we probably want to exercise MPS

jandom:
Screenshot 2026-07-27 at 19 07 44

Nice one @weitzner this is actually working :D

I'll be adding some inference tests to guard against "sudden" scientific regressions, so this will have to wait for that – but overall bloody nice work!

openfold3/core/utils/device_utils.py
weitzner: The lack of `else` is intentional to make this a no-op on a cpu - per the docstring 
```
No-op for `cpu` (no allocator cache to release) and for any accelerator without a matching `empty_cache`.
```
happy to add logging if you feel that's appropriate!

pixi.toml
weitzner: Agreed - I just couldn't abide it being called `cpu`! The `solve-group = "openfold3-cpu"` will ensure the envs are solved together and will produce identical environments. I think this may be as good as it gets with `pixi`.

weitzner:
> Overall pretty positive, there are still places where raw "cuda" lurks – some of them are probably fine but this spot is probably not
> 
> openfold3/tests/conftest.py
> 
> ```python
> _DEVICES = [
>     pytest.param("cpu", id="cpu"),
>     pytest.param(
>         "cuda",
>         id="cuda",
>         marks=pytest.mark.skipif(
>             not torch.cuda.is_available(), reason="CUDA not available"
>         ),
>     ),
> ]
> ```
> 
> Here we probably want to exercise MPS

Good callout. On my local machine, obviously, the cuda pytest is just skipped while on the CI runners, MPS would be unavailable. Happy to follow your lead to ensure we are testing this correctly.

pixi.toml
jnwei: @sdvillal what are your thoughts about renaming `openfold3-cpu` to something that is a bit more general so that we can cover MPS as well? Or do you have another suggestion for helping with this duplication?

openfold3/entry_points/experiment_runner.py
```diff
@@ -116,7 +145,10 @@ def project_entry(self) -> OF3ProjectEntry:
     @cached_property
     def model_config(self) -> mlc.ConfigDict:
         """Retrieve the model configuration."""
-        return self.project_entry.get_model_config_with_update(self.model_update)
+        model_update = self.model_update
+        if _accelerator_will_use_mps(self.pl_trainer_args.accelerator):
```
jnwei, line 149: nice usage of the ModelUpdate!

pixi.toml
sdvillal: Yes, sounds good and we already commented in that direction [here](https://github.com/aqlaboratory/openfold-3/issues/279#issuecomment-4809374586) and [here](https://github.com/aqlaboratory/openfold-3/issues/279#issuecomment-4827244236). Let's rename it to "openfold-base" or simply "openfold"?

I actually think we should also drop the "3".

In fact, let's ask pixi people to add an "alias" feature and then add openfold-mps?

Awesome work @weitzner that goes beyond improving things for metal :-)

jnwei:
Is there a canonical way to handle merges with pixi lock files?

It looks like the last merge of main with a [change where the lockfile was updated to update ruff](https://github.com/aqlaboratory/openfold-3/pull/340) has now broken the build:

```
$ pixi run -e openfold3-mps run_openfold --help
Error:   × Failed to load lock file from `/Users/jennifer/github/openfold3/pixi.lock`
  ╰─▶ environment openfold3-mps and platform linux-64 refers to a package that does not exist: https://conda.anaconda.org/conda-forge/linux-64/ruff-0.15.20-h6a952e8_0.conda
```

pixi.toml
jnwei: I would propose simplifying to `openfold3-base`. I worry that just `openfold` would be confusing compared with the original OpenFold.

If we all agree, I can go ahead and make these changes directly?

jnwei (approved):
This is awesome! With low-mem settings and mps, I can run the multimer example (7cnx) in ~25min on my 24GB M2 Mac.

jandom:
> I would propose simplifying to openfold3-base. I worry that just openfold would be confusing compared with the original OpenFold.

Agreed, i. don't like the duplication – to work on the branch directly I think we need @weitzner to add as as collaborators on his repo? (This is one of the things that make working with git a bit mental, imo)

jnwei:
No Brian already gave maintainers edit permissions to this PR. I already pushed a commit to fix the broken pixi.lock issue I mentioned [here](https://github.com/aqlaboratory/openfold-3/pull/336#issuecomment-5101683139).

pixi.toml
weitzner: Works for me!

jandom:
I've opened my inference improving PR, https://github.com/aqlaboratory/openfold-3/pull/341 once this is merged, we can update this PR and hopefully everything will look green

tomgoddard:
I tried this pull request installed with pip (not pixi) on a Mac M2 Ultra (60 GPU cores) to see how it compared for memory use and speed with the ChimeraX Mac OpenFold 3 preview 2 from 6 months ago.  Tested on a 367 amino acid protein, human mu opioid receptor (PDB 8ef5 chain R).  This pull request version is 25% faster, and takes 25% less memory, very likely because of improvement in the last several months in OpenFold 3 initialization time, and in avoiding copies of the largest tensors.  Prediction took 75 seconds and used 9 GB peak memory (according to Mac Activity Monitor) versus the preview 2 version at 98 seconds and 13 GB peak memory.

I'm preparing a class at UC San Francisco teaching structure prediction where the students all have MacBook Air M5 laptops (only 8 GPU cores) with 16 GB, so the speed and memory improvements, and the ability to run on macOS are all appeciated!

Curiously the test run output warned a few times about disabling autocast and said MPS only supports autocast to bloat16 or float16.  Maybe it was trying to autocast to float32?  At any rate inference in bfloat16 is much more memory efficient on Nvidia GPUs than float32 using Boltz 2 and Boltz 2 made it the default (bf16-mixed in pytorch_lightning).  I tested a year ago Boltz 2 bfloat16 on Mac and it did not reduce memory and decreased speed because Torch at that time did not support bfloat16 in hardware on macOS even though the ARM64 Mac chips support it.  I wonder if current Torch now supports bfloat16 in hardware with MPS.  That could make Mac OpenFold3 predictions both faster and more importantly able to run larger predictions.

I'll attach the test run output.

[openfold3_weitzner_mps_inference_8ef5_R_mac_m2_ultra_output.txt](https://github.com/user-attachments/files/30609240/openfold3_weitzner_mps_inference_8ef5_R_mac_m2_ultra_output.txt)

jandom:
This is ready to go bar 2 things 
- common pixi env, 'openfold3-base`
- merge of the inference tests https://github.com/aqlaboratory/openfold-3/pull/341

jnwei:
Updated pixi env naming from `openfold3-cpu` -> `openfold3-base`. Ran all unit tests (except the training ones) on following hardware:

- GB10
- Apple Silicon M2
- M210

Two notes for future users:
- All users who have `openfold3-cpu` hard coded in their pixi run commands will need to update to `openfold3-base`. This should happen automatically for users of other environments that consume the env fka `openfold3-cpu`

- Regenerating the lockfile requires pixi>=0.73, as in https://github.com/aqlaboratory/openfold-3/issues/352

jandom:
Bummer, we can't run integration tests on a forked PR, I propose a merge this is into a tmp branch here and then we'll get those back 

**update** okay, so these integration tests would currently be all skipped anyway, they either require triton or cueq

```
pixi run -e openfold3-base pytest -x openfold3/tests/ -m slow -vvv
openfold3/tests/test_of3_model.py::TestOF3Model::test_shape_large_eval[dtype=torch.bfloat16] SKIPPED (Requires DeepSpeed with
version ≥ 0.10.4)                                                                                                               [ 92%]
openfold3/tests/test_primitives.py::TestLMA::test_lma_vs_attention SKIPPED (Requires cuda; found mps)                           [ 94%]
openfold3/tests/test_training_full.py::test_train[smoke] SKIPPED (Requires cuda; found mps)                                     [ 97%]
openfold3/tests/test_training_full.py::test_train[full_subset] SKIPPED (Requires cuda; found mps)                               [100%]
========================================== 38 skipped, 856 deselected, 2 warnings in 16.08s ===========================================
```

But the inference tests are running, we should mark them all with "slow" so they get picked up here. 

The only failing inference test is this one but I think that's acceptable and will bump the ceil 

```
E       AssertionError: ubiquitin [no_msa-no_templates]: CA-RMSD against 1ubq.cif 1.24 Å exceeds the 0.9 Å ceiling for this mode
E       assert 1.2406350062615237 < 0.9
```

PR #330, opened 2026-07-22
add integration test for training

jnwei:
**Summary**
Introduce a pytest framework for training integration tests.

Several components are introduced:
- Scripts to generate a sample PDB training directory, which can be generated with `scripts/datasets/generate_subset_cache.py` and `scripts/datasets/download_subset.py`. These are now also available as pixi tasks (`generate-pdb-subset` and `download-pdb-subset`. This sample PDB dataset is a subset of the full PDB dataset.
- `test_training_full.py` which runs a small training job on a sample PDB dataset. 
- workflow files to test on AMD.

**Changes**
- test_training_full.py: pytest for integrated training workflow.

**Related Issues**

**Testing**
- `test_training_full.py` passes locally when tested on AMD GPU.

**Other Notes**

PR #338, opened 2026-07-27
Pytest for training integration

jnwei:
**Summary**
This PR is a subset of #330, focusing on the dataset generation and the pytest for training.

With these changes, it is now possible to locally run a test of the training pipeline in two steps:
```
# Run once to generate the training set
pixi run -e openfold3-cuda12 setup-pdb-subset

# Invoke the training test
pixi run -e openfold3-cuda12 pytest openfold3/tests/test_training_full.py
```

**Changes**
- Add scripts and helper functions to collect a subset of the full OpenFold3 PDB training set into a sample training set ( under `scripts/datasets/`). At this time, the training set is randomly selected, but the validation set is "hand-picked" to include only small examples that cover different modalities. The logic for selecting the validation set is here: `scripts/datasets/select_smoke_validation_set.py`. 
- Add pytest for training. This includes both a lightweight `smoke` version and a `full version`.
   - The smoketest runs for one epoch and uses 0 dataworkers. Time cutoff of 600s
   - The full version  runs for 2 epochs and uses 1 dataworker.

**Related Issues**
- Next step would be to incorporate into CI workflows. Ideally this would be run with a cache of the training datasets so that the training datasets would not need to be regenerated each time.

**Testing**
- [x] Run two-step training workflow on multiple systems
- [x] Verify that training tests are currently skipped when no training dataset is found.

**Other Notes**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@ openfold3/resources/
 tests/test_data/sample_feats.pickle
 cutlass/
 
+# default output of scripts/datasets/generate_subset_cache.py + download_subset.py
+# (and the generate-pdb-subset/download-pdb-subset pixi tasks, which run from here)
+/datasets/
+
 # User-specific claude settings
 .claude/
 
@@ -23,6 +27,3 @@ cutlass/
 
 # uv environments
 .venv/
-
-# build testing
-test-sdist/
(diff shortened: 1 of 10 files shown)


jandom (approved):
Ran for me no problem – i'll probably forget how to run it in 24h 

Can we update the docs, to include a description of how to run this? The relevant files could be either in installation.md, or under contribution.md?

jandom:
Amazing!!

PR #341, opened 2026-07-28
refactor: inference tests

jandom:
**Summary**

Move to a world where we can catch inference regression easy. Still not ideal, very brittle (CA-RMSD only, with a ceil only)

**Changes**

The best way to review is going commit by commit in github, happy to pair. 

The 1st moves are purely mechanical – putting inference_full into a directory, split the tests for inference into templating and regular inference, isolate out some helper classes. 

Towards the end it's just cleanup, stop parsing CIFs over and over, and tighten up the metrics. 

Challenge for the review: I can't find a good way to capture how to measure these regressions, for now we mostly do mean/min/max. There is a lot of noise in some cases from the 5 samples drawn. 

**Related Issues**

- some of our inference examples i couldn't tie to real PDBs (all the MCL1 example are the asymmetric unit, with a ligand that doesn't exist in the PDB and 3x ATP that don't bind to MCL1)
- seeing some low RMSD examples that can't be rescued by templates (MSA-free)

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/core/metrics/alignment.py
+++ b/openfold3/core/metrics/alignment.py
@@ -19,9 +19,9 @@ through :func:`~openfold3.core.metrics.quality.get_superimpose_metrics` — Kabs
 superposition followed by RMSD — so the tests measure agreement with the same primitive
 the model's own validation uses, rather than a second implementation that could drift.
 
-Everything here takes :class:`Structure`, never a path: parsing is the expensive step and
-every metric wants the same two structures, so callers parse once at the boundary and
-pass the result down.
+Everything here takes :class:`Structure`, never a path: parsing is the expensive step
+and every metric wants the same two structures, so callers parse once at the boundary
+and pass the result down.
 """
 
 import itertools
@@ -119,7 +119,10 @@ def _paired_positions(
     ref_by_chain: Mapping[str, Mapping[int, np.ndarray]],
     assignment: Sequence[tuple[str, str]],
 ) -> tuple[np.ndarray, np.ndarray]:
-    """Stack the CAs shared by each ``(pred_chain, ref_chain)`` pair, in matching order."""
+    """Stack the CAs shared by each ``(pred_chain, ref_chain)`` pair.
+
+    The two returned arrays come back in matching order.
+    """
     pred_xyz: list[np.ndarray] = []
     ref_xyz: list[np.ndarray] = []
     for pred_chain, ref_chain in assignment:
@@ -162,16 +165,22 @@ def _best_ca_assignment(
 ) -> CaAlignment:
     """Best superposition CA-RMSD (Å) of *pred* against *ref*.
 
-    ``pred_chains`` and ``ref_chains`` are matched as sets, not pairwise: every bijection
-    between them is scored and the best is returned. That is what makes a homomer
-    meaningful — with N interchangeable copies the chain labels a prediction happens to
-    emit carry no information, so pinning a specific pairing would measure label
-    agreement rather than structural agreement.
+    ``pred_chains`` and ``ref_chains`` are matched as sets, not pairwise: every
+    bijection between them is scored and the best is returned. That is what makes a
+    homomer meaningful — with N interchangeable copies the chain labels a prediction
+    happens to emit carry no information, so pinning a specific pairing would measure
+    label agreement rather than structural agreement.
 
     ``pred_chains`` defaults to every polymer chain in *pred*, which is what a
     single-chain prediction wants: the chain id the writer emits is then irrelevant.
 
-    Returns the winning assignment's ``rmsd``/``gdt_ts``/``gdt_ha`` and its superposition.
+    Returns the winning assignment's ``rmsd``/``gdt_ts``/``gdt_ha`` and its
+    superposition.
+
+    Partial duplication with
+    ``core.utils.permutation_alignment.find_greedy_optimal_mol_permutation``: that one
+    is greedy, whereas this brute-forces every assignment, which suits post-processing
+    analysis.
     """
     pred_by_chain = pred.ca_positions_by_chain
     ref_by_chain = ref.ca_positions_by_chain
@@ -275,6 +284,10 @@ def symmetry_mappings(
     For toluene this yields the two ring flips; for a fully asymmetric ligand, one. The
     predicted ligand comes from SMILES so its atom *names* are whatever the writer chose
     — matching has to go through the graph, not the names.
+
+    Partial duplication with
+    ``core.data.primitives.structure.conformer.get_cropped_permutations``, which
+    enumerates equivalent orderings but does not hand back the matches themselves.
     """
     pred_mol = _connectivity_graph(pred_selected)
     ref_mol = _connectivity_graph(ref_selected)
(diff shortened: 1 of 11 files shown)


openfold3/setup_openfold.py
```diff
@@ -208,24 +208,28 @@ def _ccd_is_stale(*, ccd_path: Path) -> bool:
...
+INTEGRATION_TEST_PATH = (
+    Path(__file__).parent / "tests" / "inference" / "test_inference_full.py"
+)
+
+
 def _run_integration_tests() -> None:
     logger.info("Running integration tests...")
     os.environ["OPENFOLD_SETUP_SCRIPT"] = "1"
-    import unittest
+    import pytest
```
jnwei: We should add a check earlier to check if pytest is installed, and prompt the user to install pytest via `pip install openfold3[dev]`

We should not add pytest directly to the core dependencies of openfold3. It is considered bad practice to do so because the large size of `pytest` creates bloat for downstream packages.

openfold3/core/metrics/alignment.py
```diff
@@ -0,0 +1,344 @@
...
+    ``transformation`` maps predicted coordinates onto the reference frame, so it can be
+    applied to anything else in the prediction — a ligand, say — to ask where it lands
+    once the protein has been placed.
+    """
+
+    metrics: dict[str, float]
+    transformation: Transformation
+
+
+def _best_ca_assignment(
```
jnwei, line 194: It is worth adding a pointer to `core.utils.permutation_alignment.find_greedy_optimal_mol_permutation`. This function does a similar task, i.e. computes the best alignment, but this function performs a brute force search over all alignments, rather than a greedy approach. The brute force approach is better for analyzing the test examples, but it is good to highlight the connection and the intentional choice to use brute force matching here.

openfold3/core/metrics/alignment.py
```diff
@@ -0,0 +1,344 @@
...
+        mol.AddAtom(Chem.Atom(str(element).capitalize()))
+    for i, j, _bond_order in selected.bonds.as_array():
+        mol.AddBond(int(i), int(j), Chem.BondType.SINGLE)
+    mol = mol.GetMol()
+    mol.UpdatePropertyCache(strict=False)
+    Chem.FastFindRings(mol)  # substructure matching needs ring perception
+    return mol
+
+
+def symmetry_mappings(
```
jnwei, line 313: Partial duplication with `openfold3/core/data/primitives/structure/conformer get_cropped_permutatinos`

Probably not worth trying to reuse `get_cropped_permutations` since that function doesn't give matches, but perhaps worth adding a comment regarding the connection.

jnwei (changes requested):
First of all, thank you so much for refactoring the inference tests. It is good to explicitly use our examples as queries, and to have some rmsd uppeer bounds to do more than just verify that output files are being generated.

I have a couple of high level comments, mostly decisions to be made about how we organize code and examples.

- Placement of new metrics functions in openfold3/tests/inference/analysis.py: The new functions are important for assessing quality, and provide functionality beyond our current analysis code. However, there is enough code here that these metrics should be placed under core.metrics.quality itself, and have separate tests written to verify these functions.

- pytest vs unittest for `setup_openfold`: See my comments in `setup_openfold.py`

- Placement of examples for pypi installation: We currently do not include the examples/ directory as part of the openfold3 package. This means that a user who installs openfold3 via `pip install openofold3` will not have the examples/ directory, and thus, all of the integration tests which use `examples` will fail because they cannot find the relevant example queries. 

To deal with the examples placement issue, we have a few options:
- Error when the examples directory is missing, prompting the user to run the tests from a clone of the repo. (easiest)
- Move the examples directory under the core `openfold3` directory which is included as the pypi build (makes examples more difficult to discover on the repo, but acceptable)
- Make a script that downloads our examples directory upon invocation. (perhaps we select a default download location is OPENFOLD_CACHE for consistent discovery)

Not all these issues need to be addressed in this PR; I think this direction is excellent overall and do not want to block over making this PR perfect.

jandom:
Hey there Jennifer, thank you – this is a great review – I think i'll address the analysis.py placement and the pytest – but not the examples/ not being included. Running the integration tests will work as currently, running only on the examples and will skip the structural alignment (remains dev-only and CI/CD).

openfold3/setup_openfold.py
jandom: This is implemented now, thanks!

openfold3/tests/core/metrics/test_alignment.py
```diff
@@ -0,0 +1,496 @@
...
+    pytest.param("1kd8", "A", "B", id="1kd8-A-vs-B"),
+    pytest.param("1kd8", "C", "D", id="1kd8-C-vs-D"),
+    pytest.param("1hf9", "A", "B", id="1hf9-A-vs-B"),
+    pytest.param("3u8v", "A", "B", id="3u8v-A-vs-B"),
+    pytest.param("4i6p", "A", "B", id="4i6p-A-vs-B"),
+]
+
+
+@pytest.mark.parametrize(("pdb_id", "pred_chain", "ref_chain"), BIOTITE_AGREEMENT_CASES)
+def test_agrees_with_biotite_superimposition(pdb_id, pred_chain, ref_chain):
```
jnwei: Is the purpose of this test to compute RMSD with biotite and with our rmsd implementation, and ensure they get equivalent results?

This was not immediately obvious to me when I was reviewing this test. I think the confusion for me is `struc`, the biotite import vs. `Structure`, the new data class introduced and being tested in this module.

Maybe renaming `expected` to `biotite_expected` or something similar would make it more explicit.

openfold3/tests/core/metrics/test_alignment.py
```diff
@@ -0,0 +1,496 @@
...
+
+def test_symmetry_equivalent_relabelling_scores_zero():
+    """A ring-flipped toluene is the same pose; only the atom order differs."""
+    reference = load_structure("7l39")
+    ligand = reference.heavy_atoms("D")
+    flip = next(
+        mapping
+        for mapping in symmetry_mappings(ligand, ligand)
+        if list(mapping) != sorted(mapping)
+    )
```
jnwei: This iterator chain logic is a bit hard for me to follow.

Is it possible to just exlicitly write the old and new atom orders (that are flips of each other), so that it becomes trivially obvious that it is a flip?

openfold3/tests/test_setup_openfold.py
```diff
@@ -9,3 +18,60 @@ def test_setup_biotite_ccd(tmp_path):
...
     has_downloaded = setup_biotite_ccd(ccd_path=ccd_path, force_download=False)
     assert not has_downloaded
+
+
+def test_require_pytest_passes_when_installed():
+    """This suite runs under pytest, so the probe must find it."""
+    _require_pytest()
+
+
+def test_require_pytest_exits_with_install_instructions(caplog):
```
jnwei: use pytest to test behavior when pytest isn't installed... *brain explosion*

jnwei (approved):
Thanks Jan! These tests for the quality metrics are super thorough and fantastic, thank you for adding them.

One overall comment: Do we want to block `setup_openfold` upon not having `pytest` installed? 

I agree that it's helpful to kill the setup at the very beginning when pytest isn't present. This behavior of early exit of the setup pipeline due to missing pytest may be annoying to some users, since pytest itself is not critical for setting up openfold. The comment does prompt the user to rerun setup without integration tests, but this would require the user to resubmit.

An alternative would be to throw a warning that the integration tests will be skipped. We can provide a prompt for the user can run the integration test after installing pytest afterwards, if desired.

Otherwise, the rest of my comments are small nits. Great work!

jandom:
> I agree that it's helpful to kill the setup at the very beginning when pytest isn't present. This behavior of early exit of the setup pipeline due to missing pytest may be annoying to some users, since pytest itself is not critical for setting up openfold. The comment does prompt the user to rerun setup without integration tests, but this would require the user to resubmit.

No i think that's a bit harsh, if we bail people who just want to pull weights on installing this, then it should come pre-installed (which we agree is not a great idea) 

> An alternative would be to throw a warning that the integration tests will be skipped. We can provide a prompt for the user can run the integration test after installing pytest afterwards, if desired.

This would be better, imo – there is no point asking to run integration tests without pytest installed, so we could warn and proceed.

openfold3/tests/test_setup_openfold.py
jandom: well it's mocked out sooo :P

openfold3/tests/core/metrics/test_alignment.py
jandom: This is a good question actually, struc.superimpose has no concept of chain matching, it just assumes they match. so this is a smoke test to confirm, on an easy case, that our alignment doesn't bork anything and agrees with biotite.

openfold3/tests/core/metrics/test_alignment.py
jandom: I kind of changed it but i don't like it very much – will revisit to tomorrow :-)

PR #343, opened 2026-07-29
fix: template_cache key'ing bug

### LINKED ISSUE #294: [BUG] CIF templates seem to get ignored
Hi, I'm struggling with CIF Direct Templates approach - it seems my templates get ignored. 

(Apologies, I can't share the protein sequence or CIF templates here due to confidentiality:) My protein A has experimental structures in the PDB where it exists in a conformation 1. From close homologs, I know that these proteins also exist in a significantly different conformation 2 (think active vs. inactive state). I want to model protein A in conformation 2, using CIF templates from its homologs. 

I added several CIF templates via "_template_cif_paths_" in the _.json_ file. But when I run the prediction, I still only get models in conformation 1 - just like when I don't use templates at all. Adding "_--use-templates=true_" seems to have no effect. Likewise, "_--use-msa-server=false --use-templates=true_" produces a diverse set of seemingly random protein structures, without indication that the provided CIF templates contributed to the structures.

**Configuration:**
 - OS: Fedora 44 
 -  NVIDIA GeForce RTX 4070 16 GB
 -  current version via pip

jandom:
**Summary**

Fixes https://github.com/aqlaboratory/openfold-3/issues/294

Two bugs 
- auto-spilling of templates between queries: template_cache uses the query sequence (and sequence only) to key its entries. Queries with the same sequence over-write it, even though they could have different template settings!
- remove sentinel -1 from idx_map, because it breaks things downstream

**Changes**

**Related Issues**

**Testing**

**Other Notes**


openfold3/core/data/io/sequence/template.py
```diff
@@ -382,6 +382,30 @@ class TemplateData(NamedTuple):
     seq: str | None
     cif_path: Path | None = None
 
+    def build_residue_idx_map(self) -> np.ndarray:
+        """Builds the query-to-template residue index map for a template cache entry.
+
+        `query_aln_pos` and `aln_pos` are per-alignment-column residue indices that use
+        -1 where the query resp. the template has a gap. The template cache contract
```
jnwei: Could we expand this abbreviation for clarity? Is "resp." short for "respective to" or something else?

openfold3/tests/core/data/io/sequence/template/test_template_parsers.py
```diff
@@ -276,3 +277,63 @@ def test_parse_templates_from_cif_files_missing_cif_skipped():
...
+        chain_id="A",
+        query_aln_pos=query_ids_hit,
+        aln_pos=template_ids_hit,
+        seq_id=1.0,
+        q_cov=1.0,
+        seq=template_aln.replace("-", ""),
+    )
+
+
+@pytest.mark.parametrize(
```
jnwei, line 300: good test! nice clear example

jnwei (approved):
Amazing detective work again @jandom ! The solution of updating the hash keys of the template caches to use the entire file contents rather than just the sequence is a good solution, and you have provided many clean unit tests and integration tests to support the changes. Great work.

My main feedback here: Can we update both the title of this PR and the description to mention something about template_caches ? It was a bit opaque to me just based on this line in the PR description: "auto-spilling of templates between queries"

I think the changes to the template cache keying / invalidation could be highlighted more in this PR so that future readers can have a better sense of what changed here.

openfold3/core/data/io/sequence/template.py
jandom: Re-worded that section completely, good catch

Issue #31, opened 2025-11-10
raw colabfold outputs always deleted even if `cleanup_msa_dir` is disabled


qurat-ul-ain95:
Dear Team,

Thank you for your hard work on this project and for releasing it opensource for everyone's benefit! :) 

I have noticed the following issue in MSA generation pipeline:

openfold3/openfold3/entry_points/experiment_runner.py currently always deletes the `raw` msa folder after conversion into .npz format regardless of how the `cleanup_msa_dir ` flag is set. 
The user might need raw colabfold outputs for analysis or downstream processing.

Submitting a PR for this issue.
Thank you!

jnwei:
Thanks @qurat-ul-ain95 looking forward to the PR

jnwei:
The issue of what to do with the raw colabfold directory has been raised in several different issues (https://github.com/aqlaboratory/openfold-3/issues/31, https://github.com/aqlaboratory/openfold-3/issues/39). I'd like to have a discussion here about what outcome is the best path forward, since the current solution seems not yet satisfactory for users who wish to preserve colabfold outputs.

AFAICT, we have 3 main user journeys:
- Case 1: [working] The user runs openfold3 inference and wishes save the raw colabfold artifacts. This is the use case @qurat-ul-ain95 initially described, that @ahmedtaha100 also wishes to support in #344 
- Case 2: [working] The user runs openfold3 inference, but wishes to save only the openfold3 alignment artifacts (organized by unique chain / paired chains) and does not need the colabfold directory
- Case 3: [error] The user hits an error, either from openfold3 testing or server issues. Because the colabfold artifacts may be corrupted, we want this directory to be cleared before running openfold3 again.

The current openfold3 behavior always deletes the raw directory, regardless of the `cleanup_msa_dir`, and raises an error when the raw directory is found before a new run. This does not satisfy the Case 1, but satisfies Case 2 and Case 3.

The approach in #344, proposes changing the behavior to combine raw colabfold cleanup with `cleanup_msa_dir`. 
- This means when `cleanup_msa_dir` is false, both raw colabfold outputs and OpenFold3 artifacts are saved.
-  When the user to goes to run a new inference job, the user receives an error and is prompted to move or remove the raw colabfold directory. This creates at best a helpful reminder to move the raw directory for Case 1, but for Case 2, requires users must delete the raw colabfold directory before every run. I think this is an unwanted hinderance for Case 2, especially for users who may have high nubmers of queries.

[...]

PR #344, opened 2026-07-30
fix: save per-run MSA outputs

ahmedtaha100:
## Summary

This updates the MSA output workflow discussed in [aqlaboratory/openfold-3#31](https://github.com/aqlaboratory/openfold-3/issues/31).

MSA preparation now uses a unique temporary workspace. During prediction, OpenFold saves both the raw ColabFold files and its formatted alignments by default, with separate settings controlling which records prediction keeps. The temporary workspace is always removed during final cleanup.

## Changes

- Save raw ColabFold files as soon as the server response has been validated, before OpenFold starts formatting it.
- Save OpenFold-formatted alignments after MSA processing finishes.
- Add `save_openfold_outputs` and `save_colabfold_outputs` settings, both enabled by default.
- Add `colabfold_output_dir` for choosing a different parent directory for raw records.
- `cleanup_msa_dir` now controls temporary template preprocessing cleanup. The new save settings control which MSA records are kept.
- Keep existing explicit `msa_output_directory` destinations separate from the temporary workspace and reject paths that cleanup could delete.
- Keep `align-msa-server` outputs and the paths written to `query_msa.json` after temporary cleanup.
- Update the configuration reference, inference guide, and examples.

## Related Issues

Related to [aqlaboratory/openfold-3#31](https://github.com/aqlaboratory/openfold-3/issues/31).

## Testing

- Focused affected suite: `96 passed, 4 deselected`. The four deselected tests require external S3 setup and are unrelated to this change.
- Ruff formatting and lint passed.
- Mypy passed against the current baseline.
- All three changed YAML examples parsed successfully.
- Strict Sphinx documentation build passed.
- `git diff --check` passed.
- CUDA 13 validation on an NVIDIA GeForce RTX 5090 passed for server-backed prediction, prediction without the MSA server, and `align-msa-server`. The server-backed and alignment-only runs kept their saved outputs and removed their temporary workspaces. The no-server run saved the dummy formatted MSA without creating raw output.

### DIFF
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -1160,6 +1160,11 @@ def preprocess_colabfold_msas(
         InferenceQuerySet:
             The updated inference query set with the MSA file paths added.
 
+    Raises:
+        FileExistsError:
+            If `output_directory/raw` already exists. Move or remove that directory
+            before retrying.
+
     Mapping:
         First, maps each sequence to a representative ID (deduplicate repeated
         sequences). Then maps each set of sequences in a structure to a representative
@@ -1200,19 +1205,21 @@ def preprocess_colabfold_msas(
     output_directory = compute_settings.msa_output_directory
     logger.warning(f"Using output directory: {output_directory} for ColabFold MSAs.")
 
+    raw_dir = output_directory / "raw"
+    try:
+        # Creating `raw/` is the reservation. A second process fails here.
+        raw_dir.mkdir(parents=True, exist_ok=False)
+    except FileExistsError:
+        raise FileExistsError(
+            f"ColabFold raw output directory already exists: {raw_dir}\n"
+            "Move or remove it before starting another MSA-server run with this "
+            "output directory. OpenFold3 will not overwrite or reuse existing raw "
+            "output because it may belong to a different query."
+        ) from None
+
     # Save mappings to file
     if compute_settings.save_mappings:
         save_colabfold_mappings(colabfold_mapper, output_directory)
-
-    # Abort early if a stale raw directory exists — it contains unvalidated
-    # out.tar.gz files that would be silently reused for the wrong query.
-    raw_dir = output_directory / "raw"
-    if raw_dir.exists():
-        raise FileExistsError(
-            f"ColabFold raw directory already exists: {raw_dir}\n"
-            "This is likely left over from a previous failed run. "
-            "Please remove it before starting a new run."
-        )
 
     # Run batch queries for main and paired MSAs
     colabfold_query_runner = ColabFoldQueryRunner(
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -12,20 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import getpass
 import json
 import logging
 import os
 import random
+import secrets
 import shutil
 import tarfile
 import time
 import warnings
 from collections.abc import Callable
+from contextlib import suppress
 from dataclasses import dataclass, field
+from datetime import datetime, timezone
 from enum import IntEnum
 from pathlib import Path
 from typing import Literal, NamedTuple
-from uuid import uuid4
 
 import numpy as np
 import pandas as pd
@@ -1104,6 +1107,12 @@ def _default_msa_output_root() -> Path:
     return get_of3_tmpdir()
 
 
+def _default_run_directory_name() -> str:
+    """Return a readable, per-run MSA directory name."""
+    timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S%fZ")
+    return f"msa-{getpass.getuser()}-{timestamp}-{secrets.token_hex(4)}"
+
+
 def _workspace_contains_output(workspace: Path, output_directory: Path) -> bool:
     """Return whether workspace cleanup would remove the output directory."""
     workspace = workspace.resolve()
@@ -1126,9 +1135,9 @@ class MsaComputationSettings(BaseModel):
     save_openfold_outputs: bool = True
     save_colabfold_outputs: bool = True
     colabfold_output_dir: Path | None = None
-    _workspace_token: str = PrivateAttr(default_factory=lambda: uuid4().hex)
-    _run_directory_name: str = PrivateAttr(default_factory=lambda: f"msa-{uuid4().hex}")
+    _run_directory_name: str = PrivateAttr(default_factory=_default_run_directory_name)
     _saved_output_root: Path | None = PrivateAttr(default=None)
+    _workspace_created: bool = PrivateAttr(default=False)
 
     @property
     def run_directory_name(self) -> str:
@@ -1204,32 +1213,17 @@ class MsaComputationSettings(BaseModel):
             raise
 
     def create_workspace(self) -> None:
-        """Create and mark the temporary workspace owned by this run."""
+        """Create the temporary workspace and record ownership for cleanup."""
         self.workspace_directory.mkdir(parents=True, exist_ok=False)
-        try:
-            (self.workspace_directory / ".openfold3-msa-workspace").write_text(
-                self._workspace_token
-            )
-        except OSError:
-            shutil.rmtree(self.workspace_directory, ignore_errors=True)
-            raise
+        self._workspace_created = True
 
-    def cleanup_workspace(self) -> bool:
-        """Remove this run's temporary workspace and report whether it was removed."""
-        marker = self.workspace_directory / ".openfold3-msa-workspace"
-        try:
-            if marker.read_bytes() == self._workspace_token.encode():
-                shutil.rmtree(self.workspace_directory, ignore_errors=True)
-                removed = not self.workspace_directory.exists()
-                if not removed:
-                    logger.warning(
-                        f"Could not remove temporary MSA directory: "
-                        f"{self.workspace_directory}"
-                    )
-                return removed
-        except OSError:
-            pass
-        return False
+    def cleanup_workspace(self) -> None:
+        """Remove the temporary workspace created by this settings instance."""
+        if not self._workspace_created:
+            return
+        with suppress(FileNotFoundError):
+            shutil.rmtree(self.workspace_directory)
+        self._workspace_created = False
 
     @classmethod
     def from_config_with_cli_override(
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -991,14 +991,6 @@ class ColabFoldQueryRunner:
                         msas_preparsed[k] = v.to_dict()
                     np.savez_compressed(npz_file, **msas_preparsed)
 
-    def cleanup(self):
-        """Remove raw colabfold MSA directory.
-
-        If the same MSA output directory is used, this directory must be removed
-        to avoid using old MSAs for new inputs.
-        """
-        shutil.rmtree(self.output_directory / "raw", ignore_errors=True)
-
 
 def add_msa_paths_to_iqs(
     inference_query_set: InferenceQuerySet,
--- a/openfold3/core/data/tools/colabfold_msa_server.py
+++ b/openfold3/core/data/tools/colabfold_msa_server.py
@@ -20,15 +20,17 @@ import shutil
 import tarfile
 import time
 import warnings
+from collections.abc import Callable
 from dataclasses import dataclass, field
 from enum import IntEnum
 from pathlib import Path
 from typing import Literal, NamedTuple
+from uuid import uuid4
 
 import numpy as np
 import pandas as pd
 import requests
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel, PrivateAttr
 from pydantic import ConfigDict as PydanticConfigDict
 from pydantic_core import Url
 from tqdm import tqdm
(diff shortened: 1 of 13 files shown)


openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1200,20 +1205,22 @@ def preprocess_colabfold_msas(
     output_directory = compute_settings.msa_output_directory
     logger.warning(f"Using output directory: {output_directory} for ColabFold MSAs.")
 
+    raw_dir = output_directory / "raw"
+    try:
+        # Creating `raw/` is the reservation. A second process fails here.
+        raw_dir.mkdir(parents=True, exist_ok=False)
+    except FileExistsError:
+        raise FileExistsError(
```
jnwei: Moving the `raw` directory test here, before the mappings are generated is a good move.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
jnwei, line 1: Good test set up with appropriate mocks / dummy values

jnwei (changes requested):
@ahmedtaha100 Thank you for this excellent contribution. I appreciate that you have added documentation and also extensive unit tests.

I will leave some minor comments here, but I'd like to have a deeper discussion about what the best approach is for handling the colab raw directory on the parent issue: https://github.com/aqlaboratory/openfold-3/issues/31. I'm afraid that the current approach in this PR does not work well for users with large numbers of queries that do not use colabfold outputs.

Let me know if you agree with the alternative approach outlined in https://github.com/aqlaboratory/openfold-3/issues/31, or if you have any suggestions on how to improve the approach.

ahmedtaha100:
> @ahmedtaha100 Thank you for this excellent contribution. I appreciate that you have added documentation and also extensive unit tests.
> 
> I will leave some minor comments here, but I'd like to have a deeper discussion about what the best approach is for handling the colab raw directory on the parent issue: #31. I'm afraid that the current approach in this PR does not work well for users with large numbers of queries that do not use colabfold outputs.
> 
> Let me know if you agree with the alternative approach outlined in #31, or if you have any suggestions on how to improve the approach.

Thank you for your feedback, @jnwei. I agree with the approach you proposed in #31.

The current PR is awkward for people who run repeated jobs against the same MSA directory and only want OpenFold's organized alignments. I agree that `colabfold_output_dir` should control whether the raw output is copied elsewhere, and that the working `raw/` directory should be removed whether or not a copy is requested.

I left a fuller reply on #31. The two points I raised there are:

1. The default behavior. Your proposal makes raw preservation opt-in by defaulting `colabfold_output_dir` to `None`. @jandom suggested preserving everything by default while keeping it configurable for large-scale workloads, which is the option I prefer.
2. When to copy the raw output. Your proposal waits until the full inference job succeeds. I think it is worth copying the raw output once MSA processing succeeds so a later model failure does not prevent it from being preserved.

I'll keep the early `raw/` reservation and the tests around it. Which default and copy point would you prefer? Once we figure out those two details, I'll update the PR accordingly.

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1088,17 +1123,113 @@ class MsaComputationSettings(BaseModel):
...
-    def create_dir(self) -> "MsaComputationSettings":
-        """Creates the output directory if it does not exist."""
-        if self.msa_output_directory is None:
-            from openfold3.core.data.tools.utils import get_of3_tmpdir
+    def create_workspace(self) -> None:
+        """Create and mark the temporary workspace owned by this run."""
+        self.workspace_directory.mkdir(parents=True, exist_ok=False)
+        try:
+            (self.workspace_directory / ".openfold3-msa-workspace").write_text(
+                self._workspace_token
```
jnwei: Is the purpose of the `_workspace_token` primarily to ensure that the user is able to write and delete the specified `workspace_directory`? Would it work to write "test" or the current date/time?

I was unfamiliar with this pattern so I did some research. It looks like this pattern is typically used in security practices to prevent (malicious) third parties from altering the input directories, e.g. TOCTOU attacks. I'm not sure if we need to follow the same practice here, but perhaps I am missing a use case?

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1088,17 +1123,113 @@ class MsaComputationSettings(BaseModel):
     save_mappings: bool = True
     msa_output_directory: Path | None = None
     cleanup_msa_dir: bool = True
+    save_openfold_outputs: bool = True
+    save_colabfold_outputs: bool = True
+    colabfold_output_dir: Path | None = None
+    _workspace_token: str = PrivateAttr(default_factory=lambda: uuid4().hex)
+    _run_directory_name: str = PrivateAttr(default_factory=lambda: f"msa-{uuid4().hex}")
```
jnwei: if possible, it would be good to include the user name in the default _run_directory_name, and possibly the date-time instead of the uuid hex

openfold3/core/data/tools/colabfold_msa_server.py
```diff
@@ -1088,17 +1123,113 @@ class MsaComputationSettings(BaseModel):
...
+                removed = not self.workspace_directory.exists()
+                if not removed:
+                    logger.warning(
+                        f"Could not remove temporary MSA directory: "
+                        f"{self.workspace_directory}"
+                    )
+                return removed
+        except OSError:
+            pass
+        return False
```
jnwei: This check of without the _workspace_token will be simpler as it would just check for the existance fo the directory.

Could we change this logic to so that rather than returning a bool, it raises the error directly?

Then the block should be wrapped in a finally-try-except block, e.g.

```
finally:
    try:
        shutil.rmtree(self.workspace_directory)
    except OSError:
        logger.warning(f"Could not remove temporary MSA directory: ...", exc_info=True)
```

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -321,6 +334,106 @@ def test_runner_on_multimer_example(
...
+        assert saved_raw.exists() is save_colabfold
+        assert (saved_output / "mappings/seq_to_rep_id.json").exists() is save_openfold
+        expected_root = saved_output if save_openfold else workspace
+        assert (
+            processed.queries["query1"]
+            .chains[0]
+            .main_msa_file_paths[0]
+            .is_relative_to(expected_root)
+        )
+        assert settings.cleanup_workspace()
```
jnwei: This was called on its own in the previous test, and with the assert here. I think it's preferrable to call without the assert, especially if cleanup_workspace() is changed to provide errors rather than a bool

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -321,6 +334,106 @@ def test_runner_on_multimer_example(
...
+        query = self._construct_monomer_query("TEST")
+        processed = preprocess_colabfold_msas(query, settings)
+        saved_main = (
+            saved_output / "main" / get_sequence_hash("TEST") / "colabfold_main.a3m"
+        )
+        saved_raw = settings.saved_colabfold_output_directory / "main/server-output.a3m"
+
+        assert saved_main.exists() is save_openfold
+        assert saved_raw.exists() is save_colabfold
+        assert (saved_output / "mappings/seq_to_rep_id.json").exists() is save_openfold
```
jnwei, line 403: The check of the `saved_main` and the `saved_raw` should happen after the cleanup step.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -707,27 +845,62 @@ def test_cli_output_dir_overrides_config(self, tmp_path):
...
-        with pytest.raises(ValueError) as exc_info:
+        with pytest.raises(ValueError, match="Output directory mismatch"):
             MsaComputationSettings.from_config_with_cli_override(
-                cli_output_dir, test_yaml_file
+                tmp_path / "cli-output", config_file
             )
 
-        assert "Output directory mismatch" in str(exc_info.value), (
-            "Expected ValueError on output directory conflict"
+    def test_workspace_and_default_records_share_one_run_id(self, tmp_path):
```
jnwei: nit: this is less about sharing one_run_id and more about sharing the saved_output_root? Perhaps the title of the test could reflect this?

It might also be good to test this at a higher level, e.g. when the `MsaComputationSettings.msa_output_directory` is specified, rather than using the lower level `set_saved_output_root` which the user will typically not use directly

Also, is there a reason to use the InferenceExperimentConfig instead of the MsaComputationSettings directly?

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -707,27 +845,62 @@ def test_cli_output_dir_overrides_config(self, tmp_path):
...
+        assert settings.workspace_directory.name == settings.run_directory_name
+        assert settings.saved_output_directory.name == settings.run_directory_name
+        assert (
+            settings.saved_colabfold_output_directory.name
+            == settings.run_directory_name
         )
+        assert other_settings.run_directory_name != settings.run_directory_name
+
+    @pytest.mark.parametrize("nested", [False, True])
+    def test_workspace_cannot_contain_openfold_output(self, nested):
```
jnwei, line 907: Nice corner case to test for!

jnwei (changes requested):
Thank you @ahmedtaha100 for taking on the daunting task of reworking the MSA saving behavior! The changes involved are quite complex and touch many different parts of the codebase, some of which had missing test coverage. 

I really appreciate your attention to detail with the test cases and covering many corner cases to ensure that that the test directories cannot be accidentally deleted. Great work!

I have a few mostly small comments, but overall, this PR is in good shape!

openfold3/core/data/tools/colabfold_msa_server.py
ahmedtaha100: You're right, we don't need the token for a security or TOCTOU use case here. I was only using it as an ownership marker so cleanup would not delete a directory this run did not create. I've removed it and now track successful workspace creation in memory instead.

openfold3/core/data/tools/colabfold_msa_server.py
ahmedtaha100: Good suggestion. I updated the default to `msa-<user>-<UTC timestamp>-<short suffix>`. The suffix prevents a collision if two runs start at the same time.

openfold3/core/data/tools/colabfold_msa_server.py
ahmedtaha100: Done. `cleanup_workspace()` now returns nothing and lets deletion errors propagate. Both cleanup paths catch `OSError` and log with `exc_info=True`. I kept a small in-memory flag so a failed workspace creation cannot cause cleanup to delete a directory this run did not create.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
ahmedtaha100: Agreed. I removed the assertion and now call `cleanup_workspace()` directly, since failures are raised instead of returned as a Boolean.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
ahmedtaha100: Done

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
ahmedtaha100: Agreed. I renamed this test and rewrote it to use `MsaComputationSettings` directly with `msa_output_directory`. This test no longer uses `InferenceExperimentConfig` or calls `set_saved_output_root()`.

openfold3/tests/core/data/tools/test_colabfold_msa_server.py
```diff
@@ -693,12 +813,37 @@ def test_obsolete_pdb_falls_back(self, mock_fetch):
...
+    def test_workspace_cleanup_is_unconditional(self, cleanup_msa_dir):
+        settings = MsaComputationSettings(cleanup_msa_dir=cleanup_msa_dir)
+        settings.create_workspace()
+
+        settings.cleanup_workspace()
+
+        assert not settings.workspace_directory.exists()
+        settings.cleanup_workspace()
+        assert not settings.workspace_directory.exists()
+
```
jnwei: Is this duplication intentional?

jnwei (commented):
This looks great, LGTM! Thanks @ahmedtaha :)

Two very small changes:
- Could you help remove the old cleanup function in ColabFoldQueryRunner, ie. https://github.com/ahmedtaha100/openfold-3/blob/44e6a866b12c3c9fcdd9dbdcd1f94df0da12cb07/openfold3/core/data/tools/colabfold_msa_server.py#L994-L1001 This is unused and redundant with the workspace handling you added to the MSAComputationSettings, and so we should remove it to avoid confusion.
- I had a small question regardina. duplicated cleanup /assert statement in the tests, see comments

jnwei (approved):

jnwei (changes requested):
Ah sorry, I forgot one more thing: We need to update the inference integration tests with the expected paths now that we've changed the defaults to include the msa directories

https://github.com/ahmedtaha100/openfold-3/blob/fix/issue-31-preserve-colabfold-raw/openfold3/tests/inference/test_inference_full.py#L545

It should be a matter of adding the expected directories for the alignment files to the list of expected directories to check. Let me know if you'd like a hand with this.

jnwei (approved):
LGTM thank you @ahmedtaha100 This was a prickly issue to resolve, thank you for your hard work on the fix!

PR #348, opened 2026-07-30
Deduplicate template hits by sequence 

etowahadams:
During inference, we allow for templates to have the same sequence. This is different from the [behavior in AF3](https://github.com/google-deepmind/alphafold3/blob/85c4d20505fd5cef05eac22b534d4e793971ae69/src/alphafold3/data/templates.py#L392-L396), where template hits with the same sequence are filtered out, such that the templates all have different sequences. This PR implements those changes.


openfold3/core/data/pipelines/preprocessing/template.py
```diff
@@ -2075,6 +2084,21 @@ def _preprocess_templates_for_query(
...
+                    if template.seq in seen_sequences:
+                        if self.create_logs:
+                            worker_logger.info(
+                                f"{template.entry_id} {template.chain_id} is a"
+                                " sequence duplicate of an already-accepted"
+                                " template."
+                            )
+                        continue
+                    seen_sequences.add(template.seq)
+
```
jandom: major: let's pull this out into a testable unit of code (a function) and putting under test

openfold3/core/data/pipelines/preprocessing/template.py
```diff
@@ -1600,6 +1606,7 @@ class TemplatePreprocessorSettings(BaseModel):
     max_release_date: datetime | None = None
     min_release_date_diff: int | None = None
     max_templates: int = 20
+    deduplicate_sequences: bool = True
```
jandom, line 1683: nits: boolean flags are typically a code smell, we should probably wrap this in a configuration object – but this is out of scope for this PR

jandom (changes requested):
Great PR, thank you @etowahadams – there are some minor linting problems

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -853,3 +854,153 @@ def test_template_sources_do_not_collide_across_queries(tmp_path, order):
...
+        pytest.param(
+            "m8",
+            id="m8",
+            marks=pytest.mark.xfail(
+                strict=True,
+                reason=(
+                    "M8Parser leaves TemplateData.seq unset, so the dedup check in "
+                    "step A2 compares None while step H records the sequence "
+                    "recovered by the step E realignment. The check never matches and "
+                    "deduplicate_sequences is a no-op on the ColabFold path."
```
jandom: @etowahadams this potentially is another bug

jandom (approved):
After my changes i'm obviously happy with this :D 

@etowahadams would be great if you could take another look?

openfold3/core/data/pipelines/preprocessing/template.py
```diff
@@ -2344,6 +2353,23 @@ def _preprocess_templates_for_query(
...
+                # so `seq` is populated even for parsers that leave it unset (.m8).
+                if self.deduplicate_sequences:
+                    if template.seq in seen_sequences:
+                        if self.create_logs:
+                            worker_logger.info(
+                                f"{template.entry_id} {template.chain_id} is a"
+                                " sequence duplicate of an already-accepted template."
+                            )
+                        continue
+                    seen_sequences.add(template.seq)
```
jandom, line 2371: this is moved significantly more down than it was: depending on the input (a3m, has sequence; m8 no sequence, needs to get seq from cif), the sequence is not available until this step

PR #349, opened 2026-07-31
Optimizing the process of cloning model weights when calculating EMA of weights

bluecake-dh:
**Summary**
Previously, `ExponentialMovingAverage` cloned all model parameters upon instantiation. However, duplicating every parameter consumes noticeable memory and time. For non-training workflows (such as inference), this eager cloning meant duplicating weights even when no EMA updates would ever occur.

To conserve resources, this PR defers weight cloning until the first training step actually executes. If there are no training steps, the weights are never cloned.

**Changes**
* `exponential_moving_average.py`:
– Move cloning model weights from `__init__` to method `init_params()`.
– `self.params` is empty until `init_params()` is called.

* `model_runner.py`:
– Now checks if the EMA params are empty during the training step; if so, initialize them with `init_params()`.
– In eval step, only load in the EMA object's weights if they are not empty.

**Testing**
– ~20% decrease in runtime
– ~35% decrease in %MEM peak (CPU RAM)


jnwei (changes requested):
This is a great contribution, thank you @bluecake-dh !

It would be awesome to have a few tests to verify that the ema loading behavior is still preserved. Would it be possible to add some? You can see some examples of state dict loading tests here: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_model_checkpoint.py

bluecake-dh:
> This is a great contribution, thank you @bluecake-dh !
> 
> It would be awesome to have a few tests to verify that the ema loading behavior is still preserved. Would it be possible to add some? You can see some examples of state dict loading tests here: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_model_checkpoint.py

I ran the following tests, and they all passed successfully; let me know if this script is sufficient.

```
import pytest
import torch
import torch.nn as nn
from unittest.mock import MagicMock
from ml_collections import ConfigDict

from openfold3.core.runners.model_runner import ModelRunner 

# ==========
# SETUP
# ==========
class DummyModel(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.linear = nn.Linear(10, 10)

    def forward(self, batch):
        return {"preds": self.linear(batch["x"])}

def get_dummy_config():
    config = ConfigDict()
    config.settings = ConfigDict()
    config.settings.ema = ConfigDict({"decay": 0.999})
    return config

@pytest.fixture
def runner():
    config = get_dummy_config()
    model_runner = ModelRunner(model_class=DummyModel, config=config)
    model_runner.loss = MagicMock(return_value=(torch.tensor(1.0), {"loss": torch.tensor(1.0)}))
    model_runner._log = MagicMock()
    return model_runner

@pytest.fixture
def dummy_batch():
    return {"x": torch.randn(1, 10)}


# ==========
# TESTS
# ==========
class TestEMALoading:

    def test_ema_no_init_inference(self, runner, dummy_batch):
        """
        Check EMA params remain empty if no training occurs, saving memory
        """
        assert not runner.ema.params, "EMA params should be empty upon initialization."

        runner.eval_step(dummy_batch, batch_idx=0)

        assert not runner.ema.params, "EMA params should remain empty during inference to save memory."
        
        runner.on_validation_epoch_end()
        assert runner.cached_weights is None, "Cached weights were not cleared."

    def test_ema_init_during_training(self, runner, dummy_batch):
        """
        Ensures EMA params are properly cloned exactly when the first training step runs.
        """
        assert not runner.ema.params, "EMA params should be empty upon initialization."

        runner.training_step(dummy_batch, batch_idx=0)

        assert runner.ema.params, "EMA params should be populated after training_step."
        assert "linear.weight" in runner.ema.params, "Model parameters were not cloned into EMA."

    def test_ema_weight_swapping_during_validation(self, runner, dummy_batch):
        """
        Ensures the model successfully swaps to EMA weights for validation, 
        and restores the live weights afterward.
        """
        runner.training_step(dummy_batch, batch_idx=0)
        
        original_weight = runner.model.linear.weight.clone()

        # Artificially alter the EMA weights to see if a swap happens
        with torch.no_grad():
            runner.ema.params["linear.weight"].add_(5.0)
        altered_ema_weight = runner.ema.params["linear.weight"].clone()

        runner.eval_step(dummy_batch, batch_idx=0)

        assert torch.allclose(runner.model.linear.weight, altered_ema_weight), "Model failed to load EMA weights during eval_step."
        assert runner.cached_weights is not None, "Original weights were not cached."

        runner.on_validation_epoch_end()

        assert torch.allclose(runner.model.linear.weight, original_weight), "Model failed to restore original weights after validation."
        assert runner.cached_weights is None, "Cached weights were not cleared."
```

jnwei:
This is a good test @bluecake-dh. Nice simple dummy model with clear expected behavior.

Go ahead and make a new test module. Perhaps the module could live under `openfold3/tests/core/runners` (creating a new folder. Thank you!

bluecake-dh:
I added the new test script, thank you!

jnwei (approved):
LGTM thank you @bluecake-dh ! This is a great contribution

jnwei (changes requested):
Hey @bluecake-dh upon running the full test suite, the new parameter saving behavior for exponential_moving_average skips the version tensor, creating this test failure:

FAILED openfold3/tests/test_model_checkpoint.py::TestOF3ModelCheckpointing::test_make_model_ckpt - AssertionError: assert 'version_tensor' in {}

Could you have a look? Basically, I think we need to copy the version tensor from the model parameters, since the [version tensor is defined downstream in `openfold3.projects.of3_all_atom`](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/projects/of3_all_atom/model.py#L50). We want to ensure that new models also have this tensor so that we can support version control.

Thanks!

bluecake-dh:
Hi @jnwei! Thank you for reviewing my changes. I see the issue with version_tensor that is causing the assertion error, so now the EMA params are initialized with a copy of the model's version_tensor instead of empty. To go along with this change, I made appropriate updates to logic and my testing script. I also realized that the runner for of3_all_atom needed checks similar to the ones I added for the core model_runner, so I added those. Please let me know if you have any other requests!

jnwei (approved):
Awesome, thank you for the fixes @bluecake-dh ! LGTM

PR #350, opened 2026-08-02
fix: two bugs in extract_alignments_to_pair (crash + off-by-one)

Pana-TsK:
**Summary**

`extract_alignments_to_pair`'s "exclude query row from subsequent MSA sources" step had two bugs:

1. A crash: `np.bool([1] * len(m.msa.shape[0]))` calls `len()` on `m.msa.shape[0]`, which is already an int, not something to `len()`. This raised `TypeError` immediately for any representative with hits in more than one of the configured `msas_to_pair` sources — with the default config (`msas_to_pair = ["uniprot_hits", "uniprot"]`), that's whenever a chain has data in both, which is the common case, not an edge case.

2. After fixing (1), a second bug surfaced: the "does this source have a real hit" check, `if len(m) > 1`, was applied unchanged to sources after the first. The first source still includes its query row at that point, so `> 1` correctly means "has at least one real hit." But sources after the first have already had their query row  stripped by the block just above, so the same `> 1` check silently requires *two* real hits to keep that source at all — a source contributing exactly one real hit gets dropped entirely, with no error or warning.

Both bugs are in code that's been present since a single commit (`63cd5a7fd`, 2025-11-17) and has zero test coverage, so neither was ever exercised.

**Changes**

- Fix the mask construction: `non_query_mask = np.ones(m.msa.shape[0], dtype=bool)` then `non_query_mask[0] = False`.
- Track whether a source is the first *actual* contributor via `is_first_pairable_source = len(msa_arrays_to_pair_i) == 0` (not loop position — if an earlier-listed source contributes nothing, the next one that does should still be treated as first), and use the correct row-count threshold (`1` vs `0`) accordingly for each case.

**Testing**

Added `openfold3/tests/core/data/primitives/sequence/msa/test_extract_alignments_to_pair.py`:
- A crash regression test (two populated pairable sources).
- A parametrized test covering a second source with 0/1/2 real hits,
  confirming it's included whenever it has at least one.
- A test confirming the "first contributor" tracking is based on
  accumulated state, not list position.

4 of 5 tests fail against the prior code (all via the crash); the 5th documents a scenario the prior code happened to get right by accident (the buggy branch was never entered in that specific case).


gnikolenyi:
Nice catch, both bugs are real, I reproduced them end-to-end, and the fix is correct. Do note though that our MSA pipeline produces `uniprot_hits.sto` only, iirc `uniprot.sto` is from an earlier iteration and we have it there for backward compatibility, but in practice we won't have both files for the same chain. It is generally recommended to have the alignments that come from one database in a single file.

A couple of things but @jnwei should confirm that these are necessary:
- Regenerate mypy-baseline.txt as the typecheck job fails as submitted. Your diff deletes the only lines producing baseline entries 1043 and 1113, and mypy-baseline filter exits 1 on stale entries. pixi run -e openfold3-base typecheck-sync.
- The new tests can't see the bug class they exist for. They only assert result["repA"].metadata. I broke subset() so it stops masking msa and deletion_matrix, metadata drops the duplicate query row, the msa array keeps it, the two silently disagree so all 5 tests still passed. Fix: give each row a distinct deletion value (the helper's np.zeros makes any assertion on it vacuous) and assert all three arrays describe the same rows.

Also some suggestions from a separate Claude-based review I ran that could be added:
- m = m.truncate(slice(1, None)) ([msa.py:65]) replaces all three mask lines. It matches the slicing idiom cap_seqs_per_species already uses, and it can't hit the IndexError that np.ones(0, dtype=bool)[0] = False raises on a zero-row source. Byte-identical for both metadata types the loaders produce.
- Rename min_rows — it's a strict bound (len(m) > min_rows), so min_rows = 1 actually means "at least 2 rows." Or switch to >= with 2/1 so the name is true.
- Add tests for 3+ sources and np.ndarray metadata. Nothing currently pins the strip behavior beyond 2 sources, and the .npz loader's ndarray metadata goes through a different subset() branch than the list metadata the tests use.
- Note the behavior change in the description. More paired rows → n_rows_paired_subsampled up → feature depth n_rows_total up → paired/unpaired boundary shifts → main-MSA budget shrinks 1:1. Not bit-reproducible for affected samples, though no default config is affected today.

Pana-TsK:
Thanks for the thorough review! I really appreciate the mutation-testing catch on the tests, that's an issue.

Also, thank you for explaining that uniprot_hits.sto/uniprot.sto never co-occurring in practice. I'll keep the multi-source handling in place since it's still correct behavior for anyone running a custom config that produces multiple pairable sources per chain, but useful context that it's not exercised by the default pipeline today.

On the two starred items:
- mypy-baseline: I confirmed the issue: this branch predates #354, so mypy-baseline.txt isn't in our branch's history at all. Regenerating it cleanly means rebasing onto current main first (picking up several unrelated commits, including a fairly large pixi.toml/pixi.lock update), then running typecheck-sync. Happy to do that once it's confirmed necessary, I didn't want to pull that churn onto an otherwise 2-file diff preemptively.
- Vacuous test assertions: agreed, will fix. Will give each row a distinct deletion value in the _msa() helper and assert msa/deletion_matrix/metadata all agree on which rows survived, instead of only checking metadata.

And I'll take the other four suggestions from the second review too:
- Swap the 3-line manual mask for m.truncate(slice(1, None)), this matches the idiom cap_seqs_per_species already uses, and sidesteps the IndexError a 0-row source could hit with the current masking approach.
- Rename min_rows (or switch to >=) so it doesn't read backwards from what it actually enforces.
- Add tests for 3+ pairable sources and np.ndarray metadata (the npz-loader path), of which neither is currently covered.
- Add a note to the PR description: fixing the off-by-one increases n_rows_paired_subsampled for affected samples, which shifts the paired/unpaired boundary and shrinks the main-MSA budget 1:1 — not bit-reproducible for those samples, though no default config hits this today.

Will push an update as soon as I have the time.

jandom:
> mypy-baseline

Yeah, this is my recent addition – I'm curious to hear your feedback? Yeah, if you want to avoid pulling in big lock changes, the rebase is the cleanest way – otherwise a merge commit will just bring all of this stuff in. 

I really think we need mypy because without it we're exposed to so many "misunderstanding" bugs that python, being type permissive, allows. 

Also, i'm delighted that you've found this – great work @Pana-TsK

Pana-TsK:
Behavior change from the off-by-one fix: for representatives with hits in a second-or-later pairing source, n_rows_paired_subsampled will increase for affected samples (previously-dropped single-hit sources are now correctly included). This shifts the paired/unpaired boundary and shrinks the main-MSA budget 1:1 for those samples — outputs for affected chains are not bit-reproducible against pre-fix runs. No default config exercises this path today (it requires >1 configured pairing source with real hits beyond the first), so this shouldn't affect existing reference outputs, but flagging it in case any downstream regression baseline picks it up.

PR #353, opened 2026-08-04
feat: MPS inference + fp32 autocast is CUDA-only

### LINKED ISSUE #352: [BUG] Regenerating pixi.lock requires pixi>=0.73
**Describe the bug**
Regenerating the pixi lockfile for this project with pixi =0.72 creates a panic and pixi cannot resolve the environment. This makes our pixi-requires setting somewhat misleading as we say the minimum version is 0.72 https://github.com/aqlaboratory/openfold-3/blob/main/pixi.toml#L101

**To Reproduce**

```
$ pixi --version
0.72
$ pixi clean
$ pixi run -e openfold3-cuda12 pytest openfold3/tests
▪ fetching repodata    [━━━━━━━━━━━━━━━━━━━━]
▪ solving              [━━━━━━━━━━━━━━━━━━━━]  2/2
⠁ updating lock file   [00:00:02] [────────────────────────────────────────]    0/6
thread 'main2' (3321861) panicked at crates/pixi_core/src/lock_file/resolve/build_dispatch.rs:477:17:
Error:   × failed to solve the pypi requirements of environment 'openfold3-rocm7' for platform 'linux-64-cuda12'
  ╰─▶ build dispatch initialization failed: there is no compatible Python interpreter for 'linux-aarch64'
```

**Expected behavior**
We should be able to rebuild the lockfile with the version specified in our pixi.toml

**Stack trace**
See above, generated on GB10 hardware after a `pixi clean` command.

jandom:
**Summary**

MPS inference, picking up from https://github.com/aqlaboratory/openfold-3/pull/336 — thank you @weitzner

The interesting part turned out to be precision. Every `torch.amp.autocast(device_type="cuda", ...)` was hardcoded, so on MPS they either blew up or quietly did nothing. Fixed, and pinned down with a test, because `autocast(dtype=float32)` is not portable: torch special-cases CUDA *before* it validates the dtype, so fp32 stays enabled there and is **silently** **disabled** on MPS and CPU. The force-fp32 blocks in `OpenFold3.forward`, `PairformerEmbedding` and `InputEmbedderAllAtom` are therefore no-ops outside of CUDA.

This is 'harmless' today inference runs `precision: "32-true"`, so operands are already fp32 on every backend. It bites the moment anyone enables bf16 on MPS: those regions would compute in bf16 instead of upcasting. `test_device_utils.py` is what fails first if that happens.

**Changes**

- `autocast_device_type()` / `empty_device_cache()` replacing hardcoded `"cuda"` and `torch.cuda.empty_cache()`
- `mps` preset (no triton kernels, no offload, cache clearing between blocks), auto-applied when the accelerator resolves to MPS
- `openfold3-mps` pixi env + docs; `openfold3-cpu` → `openfold3-base`
- inference tests recalibrated — see Testing

**Related Issues**

**Testing**

**Other Notes**


jandom:
Hmm, this is a bit concerning, everything with MPS is a bit higher for ubq, testing a 2x2 matrix with [msa, templates] on and off. 

```
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[no_msa-no_templates-ubiquitin] - AssertionError: ubiquitin [no_msa-no_templates]: CA-RMSD against 1ubq.cif 1.24 Å exceeds the 0.9 Å ceiling for this mode
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[no_msa-templates-ubiquitin] - AssertionError: ubiquitin [no_msa-templates]: CA-RMSD against 1ubq.cif 1.24 Å exceeds the 0.9 Å ceiling for this mode
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[msa-no_templates-ubiquitin] - AssertionError: ubiquitin [msa-no_templates]: CA-RMSD against 1ubq.cif 1.61 Å exceeds the 0.8 Å ceiling for this mode
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[msa-no_templates-query_single_protein_single_ligand] - AssertionError: pdb_7L39 [msa-no_templates]: CA-RMSD against 7l39.cif 0.41 Å exceeds the 0.3 Å ceiling for this mode
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[msa-templates-ubiquitin] - AssertionError: ubiquitin [msa-templates]: CA-RMSD against 1ubq.cif 1.52 Å exceeds the 0.8 Å ceiling for this mode
FAILED openfold3/tests/inference/test_inference_full.py::test_inference_writes_outputs[msa-templates-query_single_protein_single_ligand] - AssertionError: pdb_7L39 [msa-templates]: CA-RMSD against 7l39.cif 0.35 Å exceeds the 0.3 Å ceiling for this mode
```

I think it's just seeding/RNG and poor test assertion, investigating

> pl.seed_everything(42) seeds every generator, but each backend has its own stream — manual_seed(42); randn(4) gives [0.337, 0.129, 0.234, 0.230] on CPU and [0.905, 0.223, 0.146, 0.736] on MPS. ROCm reuses CUDA's philox implementation, so AMD replays the identical draw and passes.

This seems correct, unless we mock the generator we'll never get parity here. 

Running an integration test https://github.com/aqlaboratory/openfold-3/actions/runs/30914218305

openfold3/tests/core/utils/test_device_utils.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.mark.parametrize(
+    "operand_dtype",
+    [
+        # Why the divergence is harmless today: inference runs at
+        # ``precision: "32-true"`` (`InferenceExperimentConfig`), so the operands
+        # reaching these regions are fp32 and intent and effect coincide.
+        pytest.param(torch.float32, id="fp32-operands-agree-everywhere"),
+        # And what it would cost: enabling bf16 on MPS silently drops the fp32
+        # guarantee the call sites were written to obtain. This is the case that
+        # should fail first if that is attempted.
```
jandom, line 123: This is the big gotcha!

openfold3/tests/core/utils/test_device_utils.py
```diff
@@ -0,0 +1,152 @@
...
+    empty_device_cache(torch.device(device_type))
+
+
+def test_fp32_autocast_is_only_enabled_on_cuda(device_type):
+    """``autocast(dtype=torch.float32)`` is a no-op everywhere except CUDA.
+
+    This is the asymmetry behind the ``UserWarning: In MPS autocast, but the
+    target dtype is not supported. Disabling autocast.`` lines that inference on
+    Apple Silicon emits — from `OpenFold3.forward`, `PairformerEmbedding.forward`
+    and `InputEmbedderAllAtom.forward`, all of which ask for fp32 by name.
```
jandom, line 95: Same gotcha

jnwei (approved):
Nice catch and fix for the precision issue. Also good for understanding the amount of change in performance we can expect when changing dtypes.

Could you please update both the PR title and description to describe the percision changes mentioned here? 

LGTM otherwise

jandom:
Updated the PR title + desc but stumbled upon the mypy landmine you predicted :D

openfold3/projects/of3_all_atom/model.py
```diff
@@ -381,7 +383,9 @@ def _rollout(
         # Compute atom positions
         with (
             torch.no_grad(),
-            torch.amp.autocast(device_type="cuda", dtype=torch.float32),
+            torch.amp.autocast(
+                device_type=autocast_device_type(si_input), dtype=torch.float32
+            ),
```
weitzner, line 388: Would we want to adjust instances of hard-coded float32 dtypes like this?

jandom: I think these are set for a reason, for some modules the understanding is that you **really** don't want to change the precision. It'd be much simpler if we could!

jnwei: yeah, this is especially important for coordinates, otherwise if the precision is too low, you may end up with coordinates truncated to the nearest 0.5

PR #354, opened 2026-08-04
ci: mypy-baseline

jandom:
**Summary**

we want to start type annotation enforcement but not penalize folks for just touching a file that carried many mypy errors (past problems). this PR setups a new pixi task 'typecheck' that runs mypy and only regresses when users introduce new errors, relative to the current baseline (a txt file with mypy errors, sorted). 

if we improve the baseline, we can 'sync' mypy-baseline and up the bar. 

The GHA will output the regressions in the action log, so folks can fix-up

**Changes**

- review question: the comments are partially useful, could strip those?

**Related Issues**

**Testing**

**Other Notes**


jnwei (approved):
I'm fine with adding this as a trial. I have some concerns that this tool will be too persnickety and will have to be tuned.

Issue #40, opened 2025-11-13
Is there a way to provide an installation-default runner_yaml file?


roccomoretti:
**Is your feature request related to a problem? Please describe.**

My system is running afoul of the missing `evoformer_attn.so` issues described in #17 & #23 
The fix recommended is to provide a custom `-runner_yaml` file which turns off `use_deepspeed_evo_attention`. However, it's a bit inconvenient to have to remember to pass this command line parameter each time. It would be more convenient if I can set these defaults once at the install level, and have them take effect automatically.

**Describe the solution you'd like**
Provide a location (probably the `$OPENFOLD_CACHE` directory)  where you can place a central runner yaml file that would take effect for each run of that install. (And which could be overridden by the command line `-runner_yaml`, if needed.)

**Describe alternatives you've considered**
I guess you could manually edit the values in `openfold3/entry_points/validator.py`, but manually editing the official install files seems hacky to me, and would likely not persist if OpenFold was updated. It would also be much more user friendly to have the default values be set using the same syntax as the standard configuration takes.

jnwei:
I agree that it would be helpful to have some additional personal settings stored, perhaps in `$OPENFOLD_CACHE`

I think I would implement this solution in the following way

1. Add the settings for cuequivariance only as a model preset, e.g. add the following section to the [model presets for the of3_all_atom project](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/projects/of3_all_atom/config/model_setting_presets.yml)

```
cuequivariance_only:
   settings:
      memory:
        eval:
          use_cueq_triangle_kernels: true
          use_deepspeed_evo_attention: false
```
This will create the `cuequivariance_only` model preset, which can be used in conjunction with the other presets like `low_mem`, `pae_enabled`

2. Add some logic to read `runner.yml` files saved in `$OPENFOLD_CACHE`. One option could be to make an additional constructor for `InferenceExperimentConfig` that
i. Checks if there is a default `runner.yml` in `$OPENFOLD_CACHE`
ii. If there is default runner.yml loads this runner.yml as a dictionary (using `config_utils.load_yaml` or something similar)
iii. Updates these user-default settings with any provided runner.yml inputs provided by the user at the command line.

Would this make sense?

Between some infrastructure tasks and vacation I have in the next few weeks, it will take me some time to implement this change. I'd appreciate any external contributions.

[...]

PR #358, opened 2026-08-08
automatically load user-default runner.yml from cache directory

CesarPuentes:
**Summary**

This PR addresses Issue #40 by implementing the automatic detection and loading of a base runner.yml configuration file from the user's cache directory (defaulting to ~/.openfold3/runner.yml).

The primary motivation is to allow users to set defaults (such as disabling use_deepspeed_evo_attention) without the friction of passing a custom --runner-yaml flag on every single execution.

Additionally, to maintain traceability, this PR records the path of the loaded base file in the final experiment_config.json, making it easier to audit the trail of config updates.

Note: While the original issue stems from memory constraints, the manual tests below use output formats to verify the same underlying behavior.

**Changes**

1. **Automatic base configuration detection (`openfold3/run_openfold.py`)**:
   - Added logic to detect if a `runner.yml` file exists in the path specified by the `$OPENFOLD_CACHE` environment variable or in `DEFAULT_CACHE_PATH`.
   - If the file exists, its values are loaded as the base configuration.
   - If the user additionally provides a file via the `--runner-yaml` CLI flag, it is applied on top of the base configuration using `config_utils.deep_update`.
2. **Configuration traceability (`openfold3/entry_points/validator.py`)**:
   - Added the optional `user_default_runner_yaml_path` field to the `ExperimentConfig` model. This allows the path of the local `runner.yml` to be recorded in the final `experiment_config.json`, making it easier to audit the parameters applied during each run.

**Related Issues**

Solves #40 

**Testing**

**Manual Tests**
Three testing scenarios were executed, assuming the OpenFold cache is located at `~/.openfold3`:

- **Scenario 1 (No default configuration file):**
   - **Condition**: No `runner.yml` exists in the local cache.
   - **Result**: The inference pipeline ran correctly with the application's default parameters. In the generated `experiment_config.json`, the `"user_default_runner_yaml_path"` field was set to `null`.
   - **Output format**: `cif.gz` (default).

- **Scenario 2 (With configuration file in cache):**
   - **Condition**: The following `runner.yml` was placed at `~/.openfold3/runner.yml`:
     ```yaml
     # 1. Output structures directly in PDB format for PyMOL/ChimeraX
     output_writer_settings:
       structure_format: pdb
       full_confidence_output_format: npz
     # 2. Skip queries that are already generated (saves re-running time)
     experiment_settings:
       skip_existing: true
     # 3. CPU threads optimization
     data_module_args:
       num_workers: 4
     # 4. Optional GPU attention tweak (if needed)
     model_update:
       custom:
         settings:
           memory:
             eval:
     ```
   - **Result**: The parameters were successfully applied. The output structures were generated in **PDB** format. In the generated `experiment_config.json`, the field was properly recorded as `"user_default_runner_yaml_path": "/home/$USER/.openfold3/runner.yml"`.

- **Scenario 3 (Merging default configuration with explicit `--runner-yaml`):**
   - **Condition**: Keeping the base file in the cache (`~/.openfold3/runner.yml` with `structure_format: pdb`), inference was run by explicitly passing the `--runner-yaml override.yml` flag. The local `override.yml` file contained the following configuration:
     ```yaml
     experiment_settings:
       skip_existing: false
     data_module_args:
       num_workers: 8
     ```
   - **Result**: The merge was performed successfully. In the generated `experiment_config.json`, it was verified that:
     1. `"user_default_runner_yaml_path"` correctly pointed to the cached file (`/home/$USER/.openfold3/runner.yml`).
     2. `"structure_format"` was inherited as `"pdb"` (from the base configuration).
     3. `"skip_existing"` was set to `false` and `"num_workers"` to `8` (successfully overwritten/added by the explicit `override.yml` file).

### ⚙️ **Automated Tests**
The unit test suite was executed using the following command:
```bash
pixi run -e openfold3-cuda12 pytest openfold3/tests/*
```
- **Hardware used**: NVIDIA GeForce RTX 3060 (12GB VRAM).
- **Results**: 
   - The vast majority of tests completed successfully.
   - Two expected Out of Memory (OOM) failures occurred due to the VRAM limitations of the GPU used:
     - `test_shape_large_bf16_train` (requires more than ~11.6GB).
     - `test_lma_vs_attention` (requires more than ~11.6GB).

**Other Notes**

Linting and Formatting
Code validation commands were run successfully:
```bash
ruff format && ruff check --fix

### DIFF
--- a/openfold3/tests/test_entry_points.py
+++ b/openfold3/tests/test_entry_points.py
@@ -831,12 +831,11 @@ class TestUserDefaultRunnerYaml:
     def test_no_default_runner_yaml(self, tmp_path, dummy_ckpt_file):
         """Scenario 1: no runner.yml in cache → defaults, path is None."""
         cache = tmp_path / ".openfold3"
-        
+
         cfg = _build_inference_config(dummy_ckpt_file, cache_path=cache)
 
         assert cfg.user_default_runner_yaml_path is None
         assert cfg.output_writer_settings.structure_format == "cif"
-
 
     def test_default_runner_yaml_applied(self, tmp_path, dummy_ckpt_file):
         """Scenario 2: runner.yml in cache → settings applied, path recorded."""
@@ -896,10 +895,10 @@ class TestUserDefaultRunnerYaml:
     def test_corrupted_runner_yaml_raises_error(self, tmp_path, dummy_ckpt_file):
         """Scenario 4: cache runner.yml is corrupted → fails gracefully with yaml error."""
         import yaml
-        
+
         cache = tmp_path / ".openfold3"
         cache.mkdir()
-        
+
         # Write an intentionally malformed YAML (indentation error)
         (cache / "runner.yml").write_text(
             textwrap.dedent("""\
@@ -909,7 +908,7 @@ class TestUserDefaultRunnerYaml:
                 """)
         )
 
-        # Verify that when attempting to build the configuration, 
+        # Verify that when attempting to build the configuration,
         # the system raises a YAML error, preventing silent failures.
         with pytest.raises(yaml.YAMLError):
             _build_inference_config(dummy_ckpt_file, cache_path=cache)
--- a/openfold3/tests/test_entry_points.py
+++ b/openfold3/tests/test_entry_points.py
@@ -799,40 +799,12 @@ class TestRemoveQuerySetDuplicates:
         assert set(deduplicated_set.queries.keys()) == set(["query_2", "query_3"])
 
 
-def _build_inference_config(
-    ckpt_path: Path,
-    cache_path: Path,
-    runner_yaml: Path | None = None,
-) -> InferenceExperimentConfig:
-    """Mirrors the default-yaml resolution logic in ``run_openfold.predict``."""
-    default_yml = cache_path / "runner.yml"
-    user_default_runner_path = None
-
-    if default_yml.exists():
-        runner_args = config_utils.load_yaml(default_yml)
-        user_default_runner_path = default_yml.resolve()
-    else:
-        runner_args = {}
-
-    if runner_yaml:
-        config_utils.deep_update(runner_args, config_utils.load_yaml(runner_yaml))
-
-    return InferenceExperimentConfig(
-        inference_ckpt_path=ckpt_path,
-        cache_path=cache_path,
-        user_default_runner_yaml_path=user_default_runner_path,
-        **runner_args,
-    )
-
-
 class TestUserDefaultRunnerYaml:
     """Tests for the automatic loading of ``~/.openfold3/runner.yml``."""
 
-    def test_no_default_runner_yaml(self, tmp_path, dummy_ckpt_file):
+    def test_no_default_runner_yaml(self, dummy_ckpt_file):
         """Scenario 1: no runner.yml in cache → defaults, path is None."""
-        cache = tmp_path / ".openfold3"
-
-        cfg = _build_inference_config(dummy_ckpt_file, cache_path=cache)
+        cfg = InferenceExperimentConfig(inference_ckpt_path=dummy_ckpt_file)
 
         assert cfg.user_default_runner_yaml_path is None
         assert cfg.output_writer_settings.structure_format == "cif"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by jnwei — CHANGES_REQUESTED** [review]
This looks great, thank you @CesarPuentes !

Thank you also for providing your test cases. Could you please incorporate those into our test suite?

You can see some example tests for entry_point / runner yamls here: https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_entry_points.py, specifically the [model update tests](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_entry_points.py#L264)

As a suggestion, you can use the following pattern to mock the OPENFOLD_CACHE / home dir:

https://github.com/aqlaboratory/openfold-3/blob/320c115ac605e9a293de5abc90b28f2f09786a01/openfold3/tests/test_entry_points.py#L893C46-L894C81

```
        user_default_yaml_str = textwrap.dedent("...")
        test_yaml_file = tmp_path / "runner.yml"
        test_yaml_file.write_text(test_yaml_str)

       # use a patch to fake the HOME environment.
        env_patch = patch.dict(os.environ, {"HOME": str(tmp_path)}, clear=False)
        with env_patch:
            # test run openfold settings
           
```

Let me know if you have any questions / if I can provide more detailed guidance. Thank you!

**Comment by CesarPuentes** [comment]
Hi @jnwei,

Thank you for the guidance! I've added the requested tests to test_entry_points.py.

The tests simulate the cache directory using tmp_path and verify that the runner.yml is correctly detected, parsed, and merged into the InferenceExperimentConfig. I also added a 4th scenario to ensure the system fails gracefully (raising a yaml.YAMLError) if the user's cached runner.yml has malformed indentation.

All tests are passing locally after running both:

pixi run -e openfold3-base pytest openfold3/tests/test_entry_points.py -v

pixi run -e openfold3-base pytest -v


Let me know if everything looks good to merge!

File: openfold3/tests/test_entry_points.py [hunk 1]
```diff
@@ -799,6 +799,121 @@ def test_remove_duplicates(self, dummy_ckpt_file, dummy_output_path):
...
+
+    if runner_yaml:
+        config_utils.deep_update(runner_args, config_utils.load_yaml(runner_yaml))
+
+    return InferenceExperimentConfig(
+        inference_ckpt_path=ckpt_path,
+        cache_path=cache_path,
+        user_default_runner_yaml_path=user_default_runner_path,
+        **runner_args,
+    )
```
jnwei: This loading logic should be replaced with the direct loading logic from the ExperimentRunner.

jnwei (changes requested):
Hey @CesarPuentes thanks for adding the test cases, that is super helpful!

I have a further clarification to make regarding the test set up. We want the tests to exercise the loading logic that is used by the ExperimentRunner directly. So in other words, the tests should call [the exact same lines that are used in `run_openfold`,](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/run_openfold.py#L182-L186) e.g.

```
    expt_config = InferenceExperimentConfig(
        inference_ckpt_path=dummy_checkpoint_path,
        **runner_args,
    )
```

and we should expect that the default runner yaml is resolved based just on its location in the `$OPENFOLD_CACHE`. This will require [a patch of the `$OPENFOLD_CACHE` environment variable to a tmp_dir](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/test_entry_points.py#L894-L901) in order to inject a dummy `user_default_runner_yaml`.

As the test is now, the helper function `_build_inference_config` sides steps the resolution used by the InferenceExperimentConfig setup, which means the test doesn't exercise the desired default runner yaml behavior.

Let me know if I can provide any additional clarification, thank you!

CesarPuentes:
Hi @jnwei, many thanks for the guidance.

I've removed the _build_inference_config function and inlined the config loading to match the existing test patterns (e.g. test_low_mem_model_config_preset), calling **InferenceExperimentConfig** directly with 

`**config_utils.load_yaml(...).`


Does this align with what you had in mind?

All tests in test_entry_points.py passed again after this change.

Thanks

jnwei (approved):
Thank you @CesarPuentes , this test looks much better!

Originally I was also thinking that we could test the environment detection of the cache / runner yaml, but I think the way you've tested it with passing the `user_default_runner_yaml_path` suffices for this PR. 

I'll run our CI tests and if everything is green, I shall merge. Thank you again!

PR #360, opened 2026-08-10
fix: multi-gpu nodes hangs in run_openfold predict

jandom:
**Summary**

Never seen this when we run single-GPU inference but observed 1st with a multi-GPU node on an HPC cluster. The inference job starts to cleanup but then hangs indefinitely, until it's killed by walltime. 

Here is the Claude TLDR 
>On run_openfold predict with 8 ranks, every prediction completes and summary.txt is written, but the ranks never exit — each one blocks at interpreter shutdown in multiprocessing.util._exit_function, whose untimed Process.join() waits forever on a DataLoader worker, so all 8 processes sit at 0% GPU holding their memory until SLURM kills the job at the walltime. The cause is the forkserver start method hardcoded in safe_multiprocessing_context: workers are children of the forkserver rather than of the rank, so a rank can't waitpid() them itself and instead waits on a sentinel pipe for the forkserver to report the exit status — which never comes, because the forkserver sits idle in ep_poll while all ten workers pile up as unreaped zombies.

**Changes**

This PR for now just introduces a test that shows this failure mode

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/openfold3/tests/test_exit_deadlock.py
+++ b/openfold3/tests/test_exit_deadlock.py
@@ -1,0 +1,159 @@
+# Copyright 2026 AlQuraishi Laboratory
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Regression tests for the post-run exit deadlock in ``run_openfold``.
+
+Observed on a SLURM H100 node (8 ranks, ``run_openfold predict``): every
+prediction was written and ``summary.txt`` said COMPLETE, then all 8 ranks sat
+at 0% GPU utilisation holding their GPU memory until the scheduler killed the
+job. ``PYTHONFAULTHANDLER=1`` plus ``kill -ABRT`` on a rank gave::
+
+    File "multiprocessing/util.py",           line 428 in _exit_function
+    File "multiprocessing/process.py",        line 156 in join
+    File "multiprocessing/popen_fork.py",     line  44 in wait
+    File "multiprocessing/popen_forkserver.py", line 65 in poll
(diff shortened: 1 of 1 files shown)


jnwei:
I know this is still in draft, but when it's ready, could you provide a little more context about the run and the exact error message / stack trace?

For running with multiple GPUs - were you using a batch submission, i.e. a query json with multiple queries?

jandom:
All good questions, there is no stack-trace, the thing simply hangs until teh job is terminated due to timeout. 

The logs look like this 

```
==================================================
    PREDICTION SUMMARY (COMPLETE)    
==================================================
Total Queries Processed: 43
  - Successful Queries:  43
  - Failed Queries:      0
==================================================


Predicting DataLoader 0: 100%|██████████| 6/6 [02:58<00:00,  0.03it/s]
Removing empty log directory... # <--- HANGS HERE
srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
[2026-08-24T13:32:10.296] error: *** STEP 162.0 ON openfold-64-node-006 CANCELLED AT 2026-08-24T13:32:10 DUE to SIGNAL Terminated ***
[2026-08-24T13:32:10.310] error: *** JOB 162 ON openfold-64-node-006 CANCELLED AT 2026-08-24T13:32:10 DUE to SIGNAL Terminated ***

```

I have not seen this in other HPC environments so it could be some quirk of a particular cluster that I'm running on. 

submit script looks like this 

```
#!/usr/bin/env bash
#SBATCH --job-name=predict
#SBATCH --partition=h100full
#SBATCH --nodes=1
#SBATCH --gres=gpu:8
#SBATCH --ntasks-per-node=8
#SBATCH --output=slurm-%j.log

# Generated by notebook.ipynb - predicts the clean-cache validation set with the
# base weights and the combined__clean-cache epoch 100 checkpoint.

export TRITON_CACHE_DIR=/tmp/$USER/triton

eval "$(pixi shell-hook --manifest-path "/data//ofd/openfold-3-repos/openfold-3-public-main/pixi.toml" -e openfold3-cuda12)"

set -euo pipefail

echo '=== condition A ==='
srun run_openfold predict \
    --query-json ... \
    --runner-yaml ... \
    --inference-ckpt-path ...  \
    --output-dir ... \
    --num-diffusion-samples 5

echo '=== condition B ==='
srun run_openfold predict \
    --query-json ... \
    --runner-yaml ... \
    --inference-ckpt-path ...  \
    --output-dir ... \
    --num-diffusion-samples 5    
 

```

Issue #278, opened 2026-06-25
[MASTER ISSUE] dataset building toward finetuning cache


JenkeScheen:
# Points of improvement while turning custom PDB complexes into an OF3 finetuning cache

Author: Jenke Scheen (jenke.scheen@omsf.io), assisted by Claude Opus 4.8
Actioned by: @jandom 

## Background

Two groups independently built finetuning datasets by posing custom small molecules into
RCSB reference structures and running them through OpenFold-3's data pipeline:

- **Synthetic kinase analogs** — ChEMBL+REALSpace analogs posed into kinase co-crystal structures,
  assembled into protein–ligand complex mmCIFs and run through `preprocess_pdb_of3.py` to
  build a structure cache.
- **Local data cocrystals** — raw cocrystal PDBs converted to OF3-ready mmCIFs (8 documented
  transformations).

Both efforts hit the same class of friction: OF3's preprocessing is tightly coupled to the
exact shape of wwPDB mmCIF + the standard CCD, and **deviations usually fail silently or
crash with errors that don't point at the cause**. Below are the merged, de-duplicated
findings, rank-ordered by severity and grouped into three categories. This issue is intended as a
parent issue from which maintainers can spawn individual tickets.

Rows marked **⚑ both teams** were found *independently by both groups* — strong signal that
they bite any external dataset builder and are worth prioritizing within their category.

Severity order across categories: **A (silent data corruption) > B (hard crashes) > C
(robustness & docs)**.

**Configuration (synthetic data team):**
OS: Ubuntu 26.04 LTS (Linux 7.0.0-22-generic)
GPU used and memory available: NVIDIA GeForce RTX 5080, 16 GB (16303 MiB), driver 595.71.05 — note: the data-preprocessing work that surfaced these issues ran CPU-only
Python version: 3.14.5
Installation path: built from repo via pixi (env openfold3-cpu); OpenFold-3 version 0.4.2.dev123+g119962b51

---

## A. Silent data corruption — *highest priority*

These produce **wrong training data with no error and no warning**. They are the most
dangerous because a run looks successful. 

| Problem scenario | Fix | Fixed scenario |
|---|---|---|
[...]

PR #362, opened 2026-08-10
Fix issue 278 C3: Enable recursive CIF file discovery in preprocessing pipeline

Tanokri02:
## Summary
Addressed the issue raised in #278 C3. Changed the CIF file lookup from a flat directory search to a recursive tree search.

## Changes
Updated `openfold3/core/data/pipelines/preprocessing/structure.py` with a helper function (`_discover_cif_files`) that recursively finds all `.cif` files and performs checks to flag duplicate file stems across different directories.

## Testing
Added `openfold3/tests/core/data/pipelines/preprocessing/test_structure.py` with test cases for `_discover_cif_files`.

## Other Notes

### DIFF
--- a/docs/source/data_pipeline_reference.md
+++ b/docs/source/data_pipeline_reference.md
@@ -21,7 +21,7 @@ For detailed format specifications on the metadata and training/validation datas
 
 ## 1. PDB Preprocessing
 ### 1.1 Structure Download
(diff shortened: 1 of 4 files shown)


jandom (approved):
This is great @Tanokri02 – tests cover both positive and negative case and you de-dup by stems... that arguably is only an edge case because we read-in nested and write-out flat but that's the state today

PR #369, opened 2026-08-13
Fix lowercase atom name encoding

Tanokri02:
## Summary

Fix issue 278 B2. Single line change that ensures all inputs are uppercase for the model. 

## Changes
Changed a single line from:  to ```atom.GetProp("annot_atom_name").ljust(4)``` to ```atom.GetProp("annot_atom_name").upper().ljust(4)```
## Testing
Ensured the new line gives the model the uppercase text, while the original molecular data is unchanged.

### DIFF
--- a/openfold3/core/data/io/structure/cif.py
+++ b/openfold3/core/data/io/structure/cif.py
@@ -20,6 +20,7 @@ import pickle
 from pathlib import Path
 from typing import Literal, NamedTuple
 
+import numpy as np
 from biotite.structure import AtomArray, get_chain_count
 from biotite.structure.io import pdb, pdbx, save_structure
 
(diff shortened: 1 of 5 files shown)


Tanokri02:
The last commit address issue #278 B3 where two lettered elements needed to be titlecased in order to pass OF3 checks. The new version updates the copy of the CIF that is passed to Biotite, leaving the original CIF file unchanged. One example unit test has also been added.

Tanokri02:
The last commit updates the ```assign_entity_ids``` function to now only accept numeric entity ids. It parses through all the non-numeric entity ids and returns an error highlighting all those that need to be updated. Additionally, 2 tests have been added.

PR #373, opened 2026-08-17
Fixing online MSA pairing - inference 

### LINKED ISSUE #371: [BUG] Precomputed paired MSAs are dropped at inference
`create_paired_from_precomputed` never sets `n_rows_paired_subsampled`, and `vstack_pad_msa_arrays` gates the paired stack on that count, so the block it just built is discarded. It also costs main-MSA rows: `create_main` dedups main against the paired block first, so the shared rows are deleted too.

This is the default path for multi-chain inference — the ColabFold server sets `paired_msa_file_paths` for any query with two or more distinct protein sequences, and `paired_msa_order` defaults to `["colabfold_paired"]`. `examples/example_inference_inputs/query_multimer.json` hits it. Training is unaffected; precomputed paired MSAs aren't parsed there.

The same gap surfaces as a `KeyError` when a rep has no paired MSA, e.g. an RNA chain in a protein complex, since one rep having a block routes the whole sample here. `create_paired` avoids both problems via `map_row_ids_to_msa_arrays`, which allocates a block for every rep and then sets the count. Worth noting for whoever fixes this: `n_rows_paired_subsampled` is a single global split point, so the blocks have to come out at uniform depth or a shallower chain's main rows end up inside the region consumers treat as paired.

Introduced in 46805bf4, which moved the row count from the featurization consumer to the producers and wired up only `create_paired`.

Reproduced on 1861831 — a heterodimer where each chain's main MSA shares its first row with its paired block:

```python
import numpy as np

from openfold3.core.data.pipelines.sample_processing.msa import (
    create_main,
    create_paired_from_precomputed,
    create_query_seqs,
)
from openfold3.core.data.primitives.featurization.msa import vstack_pad_msa_arrays
from openfold3.core.data.primitives.sequence.msa import MsaArray, MsaArrayCollection
from openfold3.core.data.resources.residues import MoleculeType

PROTEIN = MoleculeType.PROTEIN


def msa(rows):
    a = np.array([list(r) for r in rows])
    return MsaArray(msa=a, deletion_matrix=np.zeros(a.shape, dtype=int))


collection = MsaArrayCollection(
    chain_id_to_rep_id={"A": "A", "B": "B"},
    chain_id_to_mol_type={"A": PROTEIN, "B": PROTEIN},
    rep_id_to_chain_id={"A": "A", "B": "B"},
    rep_id_to_mol_type={"A": PROTEIN, "B": PROTEIN},
)
collection.set_state_parsed(
    rep_id_to_query_seq={
        "A": np.array([list("ACDEF")]),
        "B": np.array([list("GHIKL")]),
    },
    rep_id_to_paired_msa={
        "A": {"colabfold_paired": msa(["AADEF", "ACAEF"])},
        "B": {"colabfold_paired": msa(["GGIKL", "GHGKL"])},
    },
    rep_id_to_main_msa={
        # row 0 of each main MSA also in that chain's paired block
        "A": {"colabfold_main": msa(["AADEF", "ACDEA"])},
        "B": {"colabfold_main": msa(["GGIKL", "GHIKG"])},
    },
)

query = create_query_seqs(collection)
paired = create_paired_from_precomputed(
    msa_array_collection=collection,
    max_rows_paired=8191,
    paired_msa_order=["colabfold_paired"],
)
main, profile, deletion_mean = create_main(
    msa_array_collection=collection,
    chain_id_to_paired_msa=paired,
    max_rows=16384,
    aln_order=["colabfold_main"],
    subsample_main=False,
    keep_subsampled_order=True,
)
collection.set_state_processed(query, paired, main, profile, deletion_mean)

print("paired rows built:", {k: len(v) for k, v in paired.items()})
print("n_rows_paired_subsampled:", collection.row_counts.n_rows_paired_subsampled)
for chain_id in ("A", "B"):
    stack, _ = vstack_pad_msa_arrays(collection, chain_id)
    print(f"chain {chain_id} MSA features:", ["".join(row) for row in stack.msa])
```

Both paired rows are built per chain, neither reaches the features, and `AADEF` is gone from chain A's main MSA as well:

```
paired rows built: {'A': 2, 'B': 2}
n_rows_paired_subsampled: 0
chain A MSA features: ['ACDEF', 'ACDEA']
chain B MSA features: ['GHIKL', 'GHIKG']
```

Expected `['ACDEF', 'AADEF', 'ACAEF', 'ACDEA']` for chain A, with `n_rows_paired_subsampled` at 2.

jnwei:
**Summary**
Fixes #371 to ensure MSA pairing occurs in the inference pipeline

**Changes**
- Verify paired MSAs have the same depth
- Create a dummy MSA input for reps with no paired input. This will ensure that every chain will have a value when processing `set_row_counts` instead of raising a KeyError.
- Warn when a paired MSA is not found for a chain, but not for RNA
- Add tests to check expected pairing behavior / expected rows, following the example provided in #371 

**Related Issues**
#371 

**Testing**
- [x] Passes all msa related tests, included new tests
- [x] Passes full unit tests / inference tests suite

**Other Notes**


openfold3/tests/core/data/pipelines/sample_processing/test_msa.py
```diff
@@ -0,0 +1,256 @@
...
+# A third, RNA representative with only a main MSA (no paired block) — real
+# ColabFold runs never produce paired MSAs for RNA, so a rep like this should
+# never trigger the "did you forget paired_msa_file_paths?" warning.
+RNA_MAIN_A3M = textwrap.dedent(
+    """\
+    >201
+    ACGU
+    >201_1
+    ACGA
+    """
```
jandom, line 94: good catch, yup

openfold3/core/data/pipelines/sample_processing/msa.py
```diff
@@ -219,10 +221,58 @@ def create_paired_from_precomputed(
...
+        assert truncated_msa is not None
+        processed_prepaired_msas[rep_id] = truncated_msa
+
+    # Precomputed paired MSAs come from a single pairing query per complex, which
+    # row-aligns hits across representatives (padding with gaps where a rep has no
+    # ortholog for a given species), so every cropped block is expected at the same
+    # depth. This must hold because n_rows_paired_subsampled below is a single
+    # global split point shared by every chain.
+    n_rows_per_rep = {
+        rep_id: int(msa.msa.shape[0])
```
jandom, line 235: rep_id is the same as the shape of the msa array? that's odd – I thought it was an identifier like thing, eg 1LOL_A or whatever

openfold3/core/data/pipelines/sample_processing/msa.py
```diff
@@ -219,10 +221,58 @@ def create_paired_from_precomputed(
...
+        for rep_id, msa in processed_prepaired_msas.items()
```
jandom, line 236: Exactly, here I think the meaning is slightly different for 'rep_id'?

openfold3/core/data/pipelines/sample_processing/msa.py
```diff
@@ -219,10 +221,58 @@ def create_paired_from_precomputed(
...
+    # paired MSA, so every chain has a paired block to look up below. This is
+    # expected for e.g. an RNA chain in an otherwise-paired protein complex
+    # (ColabFold never computes paired MSAs for RNA), but unexpected for a
+    # protein rep alongside other paired proteins, so that case gets a warning
+    # since it more likely means paired_msa_file_paths was left unset by mistake.
+    for rep_id, query_seq in msa_array_collection.rep_id_to_query_seq.items():
+        if rep_id not in processed_prepaired_msas:
+            if (
+                msa_array_collection.rep_id_to_mol_type.get(rep_id)
+                == MoleculeType.PROTEIN
```
jandom, line 257: The clear thing is that we're doing protein-only, do we need to have a case for PTM positions, or unresolved positions, or do these only come into play downstream?

jandom (approved):
Small questions good overall

openfold3/core/data/pipelines/sample_processing/msa.py
jnwei: This is a dict where rep_id is a key and the value is the depth of the msa.

This dict is here to track the number of rows in the msa array for each rep_id to ensure that they are all equal

openfold3/core/data/pipelines/sample_processing/msa.py
jnwei: We could add one. Do you have an example of what the a3m would look like for a case with a PTM / unresolved atom so I can add it as a test case?

PR #379, opened 2026-08-22
Add optional fast_trimul fp16 triangle-multiply backend (inference, opt-in)

tiagomonteiro0715:
**Summary**

Adds an optional, off-by-default backend for the Triangle Multiplicative Update that routes the op through `fast_trimul` (https://github.com/tiagomonteiro0715/fast_trimul), a fused fp16 CuTe DSL kernel. 

It targets inference on launch-bound small-N and memory-bound long-N regimes, and complements the existing cuEquivariance / Triton paths rather than replacing them. 

The default behaviour of OF3 is unchanged. With the flag off (or the package not installed), the normal PyTorch path runs  exactly as before.

**Changes**

- `openfold3/core/model/layers/triangular_multiplicative_update.py`
  - Add a guarded optional import (`FAST_TRIMUL_AVAILABLE`), mirroring the existing Triton/cuEquivariance import guards.
  - Add `TriangleMultiplicativeUpdate._ensure_fast_impl(...)`, which lazily builds the fast_trimul module and loads this layer's weights into it once. It is stored off the `nn.Module` registry (via `__dict__`) so it never appears in `state_dict` or pollutes checkpoints.
  - Add a branch at the top of `TriangleMultiplicativeUpdate.forward` gated by `self.use_fast_trimul` (default off). It runs only for inference (`not self.training`) and honours the existing residual/inplace contract(returns `z + update` when `_add_with_inplace`, else `update`). When the flag is off, the package is missing, or during training, this is skipped.
- `pyproject.toml`: add an optional-dependency group `optional-dependencies.fast_trimul` (`fast_trimul`, `cuda-python<13`).
- `docs/source/kernels.md` and `README.md`: document the optional backend, how to install (`pip install openfold3[fast_trimul]`), how to enable it, and its limits.

Design note: OF3's kernel choices are boolean flags threaded through many modules. 

To keep this PR small and low-risk I gated the backend on a per-module attribute (`use_fast_trimul`) instead of threading a new flag through the config/runner. 

**If you'd prefer it wired as a first-class `use_fast_trimul_kernels` flag matching the `use_cueq_/triton_triangle_kernels` convention, I'm happy to expand it!**

**Testing**

- Local: the modified module imports and byte-compiles cleanly; with the flag off the existing code path is untouched (no behavioural change).
- Existing suite: `pytest openfold3/tests/test_triangular_multiplicative_update.py` 
- Correctness: fast_trimul is separately validated against reference TriMul outputs  (fp16 tolerance) in its own repo; enabling `use_fast_trimul` on an A100 produced
- The fast path is Ampere (sm80) + `N % 8 == 0` + fp16; other hardware/shapes/dtypes   use fast_trimul's built-in pure-torch fallback.

**Other Notes**

- Inference only by design: the fast_trimul backward is a correct but slow torch recompute, and the backend holds an fp16 copy of the weights — hence the   `not self.training` guard.
- The kernel is fp16 while OF3 typically runs bf16; the fast path casts to fp16 and  otherwise falls back to torch. Framed accordingly as an inference option.
- No new required dependency; everything is behind the optional extra and the flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ A summary of our supported features includes:
 - Pipelines for generating MSAs using the [ColabFold server](https://github.com/sokrypton/ColabFold) or using JackHMMER / hhblits following the AlphaFold3 protocol
 - [Structure templates](https://openfold-3.readthedocs.io/en/latest/template_how_to.html) for protein monomers
 - Kernel acceleration through [cuEquivariance](https://docs.nvidia.com/cuda/cuequivariance) and [DeepSpeed4Science](https://www.deepspeed.ai/tutorials/ds4sci_evoformerattention/) kernels - more details [here](https://openfold-3.readthedocs.io/en/latest/kernels.html)
+- Optional fp16 triangle-multiply acceleration via the third-party [`fast_trimul`](https://github.com/tiagomonteiro0715/fast_trimul) CuTe kernel (inference only) - see [kernels docs](https://openfold-3.readthedocs.io/en/latest/kernels.html)
 - Support for [multi-query jobs](https://openfold-3.readthedocs.io/en/latest/input_format.html) with [distributed predictions across multiple GPUs](https://openfold-3.readthedocs.io/en/latest/inference.html#inference-run-on-multiple-gpus)
 - Custom settings for [memory constrained GPU resources](https://openfold-3.readthedocs.io/en/latest/inference.html#inference-low-memory-mode)
 - [Training data processing](https://openfold-3.readthedocs.io/en/latest/data_pipeline_reference.html) and [model training](https://openfold-3.readthedocs.io/en/latest/training.html)
--- a/docs/source/kernels.md
+++ b/docs/source/kernels.md
@@ -35,3 +35,36 @@ model_update:
 ```
 
 This runner.yml is specifically for inference, but similar settings can be used for training. 
+
+# fast_trimul (optional, third-party)
+
+OF3 can optionally route the triangle multiplicative update through
+[`fast_trimul`](https://github.com/tiagomonteiro0715/fast_trimul), a fused fp16
+[CuTe DSL](https://github.com/NVIDIA/cutlass) kernel. It is **inference only** (its
+backward is a correct but slow torch recompute), and it holds an fp16 copy of the
+layer weights, so it is opt-in and off by default.
+
+Install the optional dependency:
+
+```bash
+pip install openfold3[fast_trimul]
+```
+
+Enable it per `TriangleMultiplicativeUpdate` module by setting the `use_fast_trimul`
+attribute on the built model (before inference, after loading weights):
+
+```python
+from openfold3.core.model.layers.triangular_multiplicative_update import (
+    TriangleMultiplicativeUpdate,
+)
+
+model.eval()
+for m in model.modules():
+    if isinstance(m, TriangleMultiplicativeUpdate):
+        m.use_fast_trimul = True
+```
+
+The fast path targets NVIDIA Ampere (sm80) with `N` divisible by 8; other
+hardware/shapes/dtypes use fast_trimul's built-in pure-torch fallback. If the
+package is not installed, or during training, the flag is ignored and the normal
+OF3 path runs.
--- a/openfold3/core/model/layers/triangular_multiplicative_update.py
+++ b/openfold3/core/model/layers/triangular_multiplicative_update.py
@@ -43,6 +43,13 @@ try:
 except ImportError:
     TRITON_AVAILABLE = False
 
+try:
+    import fast_trimul
+
+    FAST_TRIMUL_AVAILABLE = True
+except ImportError:
+    FAST_TRIMUL_AVAILABLE = False
+
 _TRITON_UNAVAILABLE_ERROR = (
     "Triton kernels requested (use_triton_triangle_kernels=True) but "
     "openfold3.core.kernels.triton is not available. Ensure the package is "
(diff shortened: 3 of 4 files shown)


jnwei:
Hi @tiagomonteiro0715 this looks interesting! Thanks for sharing with OpenFold community.

A few additional questions regarding the benchmark results based on the report on the [source repo's README](https://github.com/tiagomonteiro0715/fast_trimul/tree/main):
- Did you also investigate benchmark speeds / memory consumption for the triton kernel or the cuequivariance kernel? 
- If you run our [inference integration tests](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/inference/test_inference_full.py), how much do the predictions differ by compared to the baseline?
- Could you also investigate the speed/ memory consumption for longer queries, e.g. >2000 residues?

Also some implementation suggestions:
- Is it possible for the fallback behavior to use either triton/cuequivariance, rather than a pure pytorch fallback? Either of these kernels will likely see improved performance over a pure pytorch kernel. 
- Another option might be to avoid falling back as frequently. Maybe instead of falling back when N %8 != 0, the kernel could perform some padding to ensure that N %8 == 0?

tiagomonteiro0715:
> Hi @tiagomonteiro0715 this looks interesting! Thanks for sharing with OpenFold community.
> 
> A few additional questions regarding the benchmark results based on the report on the [source repo's README](https://github.com/tiagomonteiro0715/fast_trimul/tree/main):
> 
> * Did you also investigate benchmark speeds / memory consumption for the triton kernel or the cuequivariance kernel?
> * If you run our [inference integration tests](https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/tests/inference/test_inference_full.py), how much do the predictions differ by compared to the baseline?
> * Could you also investigate the speed/ memory consumption for longer queries, e.g. >2000 residues?
> 
> Also some implementation suggestions:
> 
> * Is it possible for the fallback behavior to use either triton/cuequivariance, rather than a pure pytorch fallback? Either of these kernels will likely see improved performance over a pure pytorch kernel.
> * Another option might be to avoid falling back as frequently. Maybe instead of falling back when N %8 != 0, the kernel could perform some padding to ensure that N %8 == 0?

Thanks for all the feedback. I will verify all of these things as soon as possible!

jandom:
Let's drop this into draft until we have more clarity on the performance

tiagomonteiro0715:
Thanks again @jnwei 

I've worked through all five points. The padding suggestion worked out really well (details in #2). 

**Methodology note:** the `triton` and `cueq` numbers below are OpenFold-3's *own* kernels, not a reimplementation. I selected them through the real `forward` kwargs in `triangular_multiplicative_update.py` (`use_triton_triangle_kernels=True` on the `inplace_safe` path; `use_cueq_triangle_kernels=True`). 

All numbers: A100-SXM4-80GB, torch 2.14 / CUDA 13, B=1, c_z=c_hidden=128, standalone TriMul module; `fast+graph` is fast_trimul (fp16) with CUDA-graph capture. 

### 1) + 3) Triton / cuEquivariance kernels, and long queries (>2000)

Median ms/call, N = 8 → 3072:

| N | OF3 torch | OF3 triton | **OF3 cueq** | **fast+graph** |
|--:|--:|--:|--:|--:|
| 8 | 0.70 | 2.74 | 0.72 | **0.11** |
| 32 | 0.77 | 2.85 | 0.73 | **0.13** |
| 128 | 0.76 | 2.87 | 1.23 | **0.39** |
| 256 | 1.90 | 2.91 | **1.26** | 1.38 |
| 512 | 8.63 | 7.76 | **2.21** | 5.95 |
| 1024 | 36.9 | 34.8 | **7.67** | 29.2 |
| 2048 | 158 | 165 | **41.1** | 121 |
| 3072 | 344 | 388 | **116** | 262 |

Peak VRAM (GB) at N=2048 / 3072 - torch 15.4 / 36.9, triton 10.4 / 24.1, cueq 15.4 / 36.9, fast+graph 11.4 / 27.9.

**Honest conclusions:**
- **cuEquivariance is clearly the strongest kernel at scale**: 3–4× faster than fast_trimul for N ≥ 512 (41 ms vs 121 ms at N=2048). Your kernel is excellent here.
- **fast_trimul wins at small N (≤ ~200)**, where launch overhead dominates. Up to ~6× faster than cueq/torch (0.11 vs 0.72 ms at N=8). Crossover with cueq is around N=256.
- **Triton wasn't competitive in this config**: ~2.9 ms flat at small N (launch-bound), ≈ torch at large N; its benefit is lower memory (24 vs 37 GB at N=3072).
- fast_trimul keeps a **memory edge over cueq at scale** (11.4 vs 15.4 GB at N=2048), even where cueq wins on latency.

### 2) Prediction difference vs baseline

Full OF-3 Pairformer run twice on identical inputs/weights, swapping only the TriMul kernel (16 modules, 8 blocks):

| metric | value |
|---|---|
| max abs diff | 2.6e-5 |
| mean abs diff | 3.8e-6 |
| within fp16 tol (atol=rtol=1e-2) | **PASS** |

End-to-end predictions match to ~5 significant figures.

### Suggestion 2:pad instead of falling back on N % 8 ≠ 0

Instead of dropping to torch when `N % 8 != 0`, the kernel now pads N to the next multiple of 8, runs, and slices back (zeroing the padded contraction rows so the projection bias can't leak into the valid output). 

Validated at N=2001: max abs diff **1.7e-5** vs reference. 

The fast path now covers any N on Ampere.

### Suggestion 1: non-torch fallback (wired, with a caveat)

The benchmark makes the case directly: since cueq is 3–4× faster than fast_trimul at large N, falling back to cueq instead of pure torch is clearly right. I've added a guarded cuEquivariance fallback (chain `cuda → cueq → torch`) that mirrors OF-3's exact `_cueq_triangle_mult` call. Same `triangle_multiplicative_update`, same fused weight layout (`cat` of the a/b projections), returns the delta.

**Honest caveat:** I have **not** been able to run this fallback green end-to-end, because on the torch-2.14 stack I tested, `cuequivariance_ops_torch` fails to import (`cannot import name 'is_fx_symbolic_tracing' from torch.fx`; it wants `triton==3.3.0` and an older torch). The backend registers and calls the kernel correctly, and the guard behaves as designed (it falls through to torch when the kernel can't load, so the library never breaks).

But I'd need a `cuequivariance_ops_torch`-compatible stack (torch ~2.4–2.6 + `triton==3.3.0`) to validate the numbers. I'm not claiming it's proven faster in your pipeline, only that it's wired to your known-good call and safe.

---

Net: fast_trimul is best positioned as a small-N inference kernel, defers to cuEquivariance for long sequences, matches predictions to ~1e-5, and now handles any N. 

It stays fully opt-in / off-by-default as in this PR, and I'll keep any changes in line with the OF-3 contributing guidelines.

jandom:
Thanks @tiagomonteiro0715 this is helpful – this is good but I worry that this is too modest to merge in. Do you want to continue iterating on this, to make the improvement more robust across the board?

Taking a second look, the peak RAM reduction looks promising.

PR #380, opened 2026-08-24
Fix naming of openbind parameter download

jnwei:
**Summary**
The AMD Integration CI failed because of an erroneous renaming step upon downloading the OpenBind parameters. Removing the rename will fix the integration test. 

This renaming issue was not present on the pixi-cuda workflows, only the amd workflow needs to be fixed.

**Related Issues**
Example workflow failure: https://github.com/aqlaboratory/openfold-3/actions/runs/32686976308/job/97316237360


jandom (approved):
Great catch

PR #382, opened 2026-08-24
initial draft of SKILLS.md

jnwei:
**Summary**
Initial draft of SKILLS folder for OpenFold3. At this time, these skills designed to support input query set construction and inference based on a query input.

**Changes**
- Adds 2 skills, one for constructing queries and one for running inference

**Testing**
- [ ] Tested locally in a new directory outside of main dev directory.


jandom (approved):
This is good, it might help to just point CLAUDE at the inference docs (query and predict) rather than repeat? i'll prevent us for getting this duplicated and harder to maintain?

Like this change to MSA directory output changing, will need an update both here and in the docs atm

Issue #323, opened 2026-07-20
Allow residue/component identifier specification for SMILES input.


roccomoretti:
When using SMILES input for ligand specification, the residue/component identifier ("three letter code"/"CCD code") cannot currently be specified, defaulting to `LIG#` (where # is the ligand number) for mmCIF output. This may be an annoyance for use with larger pipelines, where different phases may be expecting a different residue identifier (necessitating tedious file editing to match up identifiers). 

For legacy PDB format output, the situation is a bit worse: all of the `LIG#` residue identifiers are truncated to `LIG`. This can cause issues with downstream programs or pipelines, which may special case LIG, or may still interpret it as the (obsoleted) 3-PYRIDIN-4-YL-2,4-DIHYDRO-INDENO[1,2-.C.]PYRAZOLE. It is also of particular concern for structures with multiple SMILES specified ligands, which all get the same `LIG` designation, even if they have distinct chemical identities.

Ideally, the input format specification should (optionally) allow setting the residue identifier for a ligand provided as a SMILES specification, probably as an additional option to the `"molecule_type": "ligand"` block.

_I'll also note that the PDB [has reserved](https://www.rcsb.org/news/feature/2023) the all-numeric two character codes 01-99 as placeholders. (Though apparently 98 exists as an obsoleted entry.) Those may be an option for PDB-format output, though allowing manual specification and keeping the current behavior when unspecified is probably fine._

jandom:
Thanks @roccomoretti this is a good suggestion!

PR #386, opened 2026-08-26
Allow custom residue names for SMILES ligands

ahmedtaha100:
**Summary**

Adds optional `ligand_name` for SMILES ligands without changing `LIG#` defaults.

**Changes**

- Trims and uppercases explicit names, accepts only ASCII alphanumeric characters, and rejects standard polymer residue names.
- Carries explicit names into the SMILES component mapping and rejects ambiguous SMILES mappings or explicit-name collisions with CCD and non-canonical residues.
- Documents full component names in mmCIF and three-character truncation in legacy PDB output.
- Corrects the query-construction docstring and updates the mypy baseline for the optional field.

**Related Issues**

 issue #323

**Testing**

- `ruff check --no-cache .` and `ruff format --check --no-cache .`
- Exact mypy baseline filter: 0 new and 0 fixed errors
- Full unit suite in CUDA 12, CUDA 13, and legacy Conda CUDA 12.1 images: 996 passed, 90 skipped in each
- Focused structure/query file in all three images: 13 passed in each
- Full slow suite on the RTX 5090 in CUDA 12 and CUDA 13: 55 passed, 2 skipped in each
- Strict Sphinx HTML build with warnings treated as errors
- Production CUDA 13 inference on the RTX 5090 using the current 174k checkpoint: `DRG123` was preserved in the generated mmCIF

The legacy Conda image's pinned PyTorch supports GPU architectures through `sm_90`, so its slow GPU suite cannot execute on the RTX 5090 (`sm_120`). Its full unit suite and focused tests pass.

@jandom @jnwei

### DIFF
--- a/docs/source/input_format_reference.md
+++ b/docs/source/input_format_reference.md
@@ -199,7 +199,8 @@ All chains must define a unique ```chain_ids``` field and appropriate sequence o
   {
     "molecule_type": "ligand",
     "chain_ids": "Z",
-    "smiles": "CC(=O)OC1C[NH+]2CCC1CC2"
+    "smiles": "CC(=O)OC1C[NH+]2CCC1CC2",
+    "residue_name": "DRG"
   }
   ```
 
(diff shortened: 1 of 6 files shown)


jnwei (approved):
Thank you, this is awesome @ahmedtaha100 ! I have no technical suggestions to make, well done. 

I do have one nomenclature question: Is it common practice to refer to the name of the molecule / ligand as a `residue_name`? I'm guessing this name is chosen to borrow the nomenclature used by chains.

Perhaps another option is to have the field in Chain for a ligand be called `ligand_name`. Then make an adjustment in `primitives/structure/query.py` so that `ligand_name` is reassigned as `residue_name` when constructing the atom_array. wdyt?

ahmedtaha100:
@jnwei Thank you! Yes, `residue_name` follows the PDB/Biotite structural terminology and maps to `res_name` in the atom array. Good examples here [https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html) and [https://www.biotite-python.org/latest/apidoc/biotite.structure.html#biotite-structure](https://www.biotite-python.org/latest/apidoc/biotite.structure.html#biotite-structure) . I can see how `ligand_name` might be clearer in the input, though. I'll adjust it. Thank you for the feedback!

ahmedtaha100:
Renamed! @jnwei

jnwei (approved):
This is great, thank you @ahmedtaha100 !

PR #391, opened 2026-09-01
fixes #390: citation file updates

### LINKED ISSUE #390: Update citation DOI
The "cite this repo" UI element points to an older release, if you use this version:
https://doi.org/10.5281/zenodo.17485509

Then it will always point to the latest release. I think that is better since if an author wants to cite a specific version, they can still do that but anyone do a quick drive by citation will have a reference that points to the latest.

Just a suggestion btw, few different ways to do this.

jandom:
**Summary**

**Changes**

**Related Issues**

**Testing**

**Other Notes**

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,13 +1,13 @@
 cff-version: 1.2.0
 message: "If you use this software, please cite it using the metadata below."
-title: "OpenFold3-preview"
-version: "0.1.0"
+title: "OpenFold3"
+version: "0.5.0"
 authors:
   - name: "The OpenFold3 Team"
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -1,13 +1,13 @@
 cff-version: 1.2.0
 message: "If you use this software, please cite it using the metadata below."
-title: "OpenFold3-preview"
-version: "0.1.0"
+title: "OpenFold3"
```
jnwei: We should keep the preview here and below for now

jnwei (approved):

PR #393, opened 2026-09-02
More memory efficient metrics

chelsea0x3b:
**Summary**
Some simple code changes to make the metrics/quality.py code a bit more memory efficient. These changes should be bitwise identical, so this is purely a logic change.

This code is called once per log step, and does have a significant impact on memory

**Changes**
1. Replaced `torch.where(a == b, 1, 0).bool()` with `a == b`, removing the int -> bool intermediate tensor

This one is pretty straightforward to understand.

2. Replaced `* (1 - torch.eye())` with `diagonal(...).zero_()`

The original version multiplies a tensor by a matrix where all off diagonal values are 1, and diagonal values are 0. The equivalent logic is to just get a view of the tensor's diagonal. So then we call `zero_()` to explicitly zero the diagonal elements. This means you have no additional intermediate tensors (diagonal makes a view, zero_ works in place).

**Results**

The above changes result in an average of 5% decrease (on B200s) in device memory when running the `OpenFold3AllAtom::_get_metrics()` method. The average memory saved is calculated as **~1.7GB**.

**Related Issues**
N/A

**Testing**
The same openfold training runs before & after this change.

**Other Notes**
I calculated the improvement by running the same training command twice with the below changes added in:
```diff
diff --git a/openfold3/projects/of3_all_atom/runner.py b/openfold3/projects/of3_all_atom/runner.py
index 8b6f30ef..90b54832 100644
--- a/openfold3/projects/of3_all_atom/runner.py
+++ b/openfold3/projects/of3_all_atom/runner.py
@@ -299,8 +299,28 @@ class OpenFold3AllAtom(ModelRunner):
     ):
         phase = "train" if train else "val"
 
+        torch.cuda.reset_peak_memory_stats()
+        mem_before = torch.cuda.memory_allocated()
+
         metrics = self._get_metrics(batch, outputs, train=train)
 
+        peak_alloc = torch.cuda.max_memory_allocated()
+        peak_resv = torch.cuda.max_memory_reserved()
+        mem_after = torch.cuda.memory_allocated()
+        mb = 1024**2
+
+        n_atom = outputs["atom_positions_predicted"].shape[-2]
+        n_sample = outputs["atom_positions_predicted"].shape[1]
+        spike=(peak_alloc - mem_before) / mb
+        with open(f"memprobe.csv", "a") as fp:
+            parts = [
+                ','.join(batch['pdb_id']),
+                n_atom,
+                n_sample,
+                spike,
+            ]
[...]

PR #394, opened 2026-09-03
Remove pytorch channel from conda env file and upper bound deepspeed

jnwei:
**Summary**

A bad interaction between outdated pytorch from the pytorch conda channel and deepspeed is causing the unit test failures seen here: https://github.com/aqlaboratory/openfold-3/actions/runs/33709917312/job/100507354489?pr=393

`pytorch:pytorch` effectivelyi pins pytorch to 2.5.1, whereas newer versions of deepspeed rely on newer ops available in 2.7.1

The pixi build already fetches pytorch from conda-forge. Pinning deepspeed for now as we plan to deprecate it completely

**Changes**
- Update production yaml with pytorch and deepspeed changes. Also needs cudart-dev since the conda forge pytorch no longer includes it
- Update production lock generation script.
- Update all CI workflows to use `setup_openfold --non-interactive` to streamline parameter download

**Related Issues**
This issue caused a failure on the CI tests for #393 and will likely break future conda CI tests.

**Testing**
- [x] Manual GPU workflow triggered on internal branch ran successfully.

**Other Notes**


jnwei:
Paging @sdvillal for a quick review in case I've missed anything here

docker/Dockerfile.conda
```diff
@@ -37,7 +37,12 @@ COPY environments/production-linux-64.yml /opt/openfold3/production-linux-64.yml
 # Install environment based on BUILD_MODE
 # - lock: uses conda-lock for exact reproducible builds (training/production)
 # - yaml: uses mamba env create for flexible version resolution (development/testing)
-RUN mamba install -n base -c conda-forge conda-lock --yes \
+# CONDA_OVERRIDE_CUDA bypasses GPU detection while solving: conda-forge's CUDA
+# builds of pytorch are gated on the `__cuda` virtual package, which is absent on a
+# GPU-less build machine. Advertise a CUDA 12 driver so the CUDA build is selectable
+# (same trick as docker/Dockerfile.pixi); it changes nothing at runtime.
+RUN export CONDA_OVERRIDE_CUDA=12.0 \
+    && mamba install -n base -c conda-forge conda-lock --yes \
```
jandom, line 45: Note to self: I would like to remove conda support but I suspect the researchers like it

jandom (approved):

sdvillal:
Sorry for the slow reaction.

Definitely nobody should be using the pytorch channel.

If we still want to provide a conda environment yaml, could we  simply export the environments from the pixi workspace? There are tasks for that already (I suspect they need a small update) that should efficiently produce both regular and conda-lock versions in sync with our pixi world.

But maybe we should be bold and ask people to simply install pixi if they want a good conda environment. We do not have that many resources to maintain so many things.

docker/Dockerfile.conda
sdvillal: Second this, as wrote below

environments/production-linux-64.yml
```diff
@@ -29,14 +27,15 @@ dependencies:
   - bioconda::hmmer
   - bioconda::hhsuite
   - bioconda::snakemake
-  - pytorch::pytorch
-  - pytorch::pytorch-cuda
+  - conda-forge::pytorch-gpu
```
sdvillal: minor: we should not need the channel specifier

environments/production-linux-64.yml
```diff
@@ -29,14 +27,15 @@ dependencies:
...
+  - conda-forge::cuda-cudart-dev
+  - conda-forge::deepspeed>=0.19,<0.20
   - memory_profiler
   - func_timeout
   - boto3
   - conda-forge::python-lmdb=1.6
   - conda-forge::ijson
   - pip:
-      - deepspeed
       - pdbeccdutils
```
sdvillal: we should be able to also use the conda-forge dependencies here

environments/production-linux-64.yml
```diff
@@ -29,14 +27,15 @@ dependencies:
   - bioconda::hmmer
   - bioconda::hhsuite
   - bioconda::snakemake
-  - pytorch::pytorch
-  - pytorch::pytorch-cuda
+  - conda-forge::pytorch-gpu
+  - conda-forge::cuda-version>=12,<13
+  - conda-forge::cuda-cudart-dev
```
sdvillal: just our of curiousity, is this enough to compile deepspeed's Evoformer's attention? or do we combine with system/docker instructions?

If we combine with docker/system I would be cautious to ensure compatibility

sdvillal:
In the dockerfile we are mixing system cuda with conda-forge provided cuda versions. While valid, I would consider this an antipattern - to me the best practice is to either manage build CUDA dependencies entirely in the conda world with full isolation and reproducibility or system/docker-wide with good docs for the user.

After reviewing this, I would vote for removing explicit conda environment support altogether and focus our efforts on modernizing everything where conda is involved to use pixi. Otherwise we are supporting a too broad matrix.

jnwei:
Yeah I agree with you @sdvillal

This conda environment.yml is providing a false sense of support, I really wasn't sure what was the best way to handle the cuda library issue with the pytorch packages.

@jandom if you're on board, I'll close this PR and send a PR to remove the conda environment file and associated tests? If I understand correctly, we can direct conda users to use a conda appropriate lockfile generated from pixi instead, which we can update alongside the main pixi.lock

jnwei:
Talked with @jandom IRL, closing this PR in favor of one that will remove conda

PR #385, opened 2026-08-26
Ligand chemical steering

peter-s-obi:
**Summary**

Adds query-level ligand chemical steering for inference. A user can enable `ligand_chemical_steering` for a query, and OpenFold-3 (OF3) will steer ligand geometry toward chemically valid configurations during reverse diffusion.

Implementation strategy:

- Derive flat-bottom distance, tetrahedral chirality, E/Z, planarity, and VDW-overlap restraints from OF3 processed reference molecules
- Apply analytical coordinate updates to the denoised coordinate estimate during the late diffusion window
- Continue through OF3's standard reverse-diffusion update after each steering step

**Changes**

- Adds an optional `ligand_chemical_steering` field at the query level
- Adds Pydantic steering settings under `dataset_config_kwargs.ligand_chemical_steering`
- Uses OF3 processed reference molecules to support both SMILES and CCD ligands, including small molecules and macrocycles
- Adds featurization for RDKit distance-geometry bounds, assigned tetrahedral centers, assigned E/Z bonds, double-bond and ring planarity
- Adds analytical flat-bottom coordinate updates to `SampleDiffusion`
- Adds a runnable inference example, configuration documentation, third-party attribution, and focused tests

**Related Issues**
None.

**Testing**

All general OF3 tests and ligand chemical-steering-specific tests passed.

Benchmark testing used the 100 smallest post-2025 Runs N' Poses targets with the OpenBind-174k checkpoint, one seed and five diffusion samples, with MSA and templates enabled. Rates are over 103 ligand chains, represented by the top-ranked sample for each chain.

| Metric | Default OF3 | Chemical steering |
|---|---:|---:|
| PoseBusters valid | 74/103 (71.8%) | 96/103 (93.2%) |
| Fully correct stereochemistry, ligand predictions | 156/230 (67.8%) | 230/230 (100%) |
| Correct tetrahedral centers | 515/615 (83.7%) | 615/615 (100%) |
| Median ligand RMSD | 1.78 A | 1.67 A |
| Median runtime per target | 28.18 s | 29.31 s |

Median ligand RMSD was included as a placement guardrail

**Other Notes**

- Restraints are flat-bottom: valid geometry has zero energy and zero coordinate update
- Steering uses analytical gradients and does not require autograd during inference
- Default steering added approximately 4% median wall-clock overhead on an A100 40 GB GPU
- Adapted Boltz components and parameters are documented in `THIRD_PARTY_NOTICES.md`

PR #395, opened 2026-09-04
Chemical steering: gradient guidance toward valid ligand chemistry

jnwei:
**Summary**

This PR adds stereochemistry steering to OpenFold3. In this PR the foundation is laid for adding more potentials, and also more test cases, in particular by introducing  a new top level `steering` directory. 

**Provenance**
The work here is based on @peter-s-obi #385  and @etowahadams [Foldsteer repo](https://github.com/etowahadams/foldsteer). Both of these works in turn borrowed formulations and code from Boltz and Protenix

**Changes**

- Type classes to represent Potentials and Schedules
- Functions to create features which represent bounds defined by the ligand stereochemistry
- DistanceBoundsPotential as a first example potential
- Hooks to call steering in the data pipeline (in the `InferenceDataset`) and in the diffusion module
- Steering configurations settings: `SteeringSettings`
- Documentation to include steering descriptors 
- THIRD_PARTY_NOTICES.md to record provenances in the `openfold3/steering` directory 

**Testing**

- Unit tests for all new steering pieces
- An integration test with PDB:1FKB which contains rapamycin and explicitly tests for saturated six-ring torsion (h/t [Tom Goddard's BadChem](https://www.rbvi.ucsf.edu/chimerax/data/ligchem-feb2026/badchem.html)) 
- Full unit tests / integration tests (running)
- mypy linting

**Other Notes**

The individual commits are organized by themes. It might be easiest to review this PR commit by commit.

### DIFF
--- a/openfold3/steering/config.py
+++ b/openfold3/steering/config.py
@@ -1,0 +1,108 @@
+# Copyright 2026 AlQuraishi Laboratory
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Run-level settings for chemical steering.
+
+Every numeric default is referenced from ``defaults.py`` rather than
+restated here, so the derived parameter table stays in exactly one place
+(see THIRD_PARTY_NOTICES.md).
+"""
+
+from __future__ import annotations
+
+from pydantic import BaseModel, Field, field_validator
+from pydantic import ConfigDict as PydanticConfigDict
+
+from openfold3.steering import defaults
+from openfold3.steering.potentials import CLASS_REGISTRY, DistanceBoundsPotential
+
+
+class TermSettings(BaseModel):
+    """Settings for one restraint family.
+
+    Attributes:
+        enabled (bool):
+            Whether this term contributes to the guidance gradient. A
+            disabled term emits no restraints, so the sampler never sees it.
+        weight (float):
+            Coordinate-update weight. Because the flat-bottom subgradient is
+            exactly +-1 and a distance has |dv/dx| = 1, this is literally the
+            step size: a violating atom moves `weight` Angstrom per
+            gradient-descent step.
+        interval (int):
+            Apply this term every Nth gradient-descent step. Intended for
+            terms too expensive to evaluate every step; 1 means every step.
+    """
+
+    model_config = PydanticConfigDict(extra="forbid", allow_inf_nan=False)
+
+    enabled: bool = True
+    weight: float = Field(default=defaults.DISTANCE_WEIGHT, ge=0.0)
+    interval: int = Field(default=1, ge=1)
+
+
+class SteeringSettings(BaseModel):
+    """Settings for inference-time chemical steering of ligand geometry.
+
+    Consumed by `maybe_create_steering_features` in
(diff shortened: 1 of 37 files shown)


peter-s-obi:
Thanks Jennifer for putting this together! Two things:

1. I noticed the CI unit-test commands still target only `openfold3/tests`, so they’ll miss the 98 tests under `openfold3/steering/tests.` Could we include that directory too?

2. The docs describe `num_gd_steps * weight `as a maximum atom displacement, but `grad.index_add_` sums contributions from every restraint involving an atom. Thoughts on describing weight as the contribution per violated distance restraint per gradient step instead?

openfold3/steering/types.py
```diff
@@ -0,0 +1,91 @@
...
+    num_steps: int
+    start_step: int  # non-zero in the pocket-refinement pass
+    pass_name: str  # "primary" | "pocket_refine"
+    atom_mask: Tensor
+    ctx: SteeringContext
+
+    @property
+    def steering_t(self) -> float:
+        """Normalized time, Boltz/Protenix convention: 1.0 at the start -> 0."""
+        return 1.0 - self.step_index / max(1, self.num_steps)
```
jnwei, line 79: @peter-s-obi The time convention for [the scheduler in #385 uses:(https://github.com/aqlaboratory/openfold-3/pull/385/changes#diff-63c6f46c3834559752177553d131f709310760a1dcc7f42f1532a22d9eaaffcbR351)

```
step_fraction=(tau + 1) / max(num_denoising_steps, 1)
```

Here, this was changed to match the convention used in Boltz / Protenix.  I wanted to run this change by you to ensure that this is correct?

jnwei:
Thank you @peter-s-obi for the suggestions:

1. Good point about the CI tests covering steering, I have made that change now.
2. The change you suggest to the documentation makes sense, although I'm not finding the exact documentation you're looking for.

---

I had two questions for you where the implementation here differs a bit from #385. I wanted to make sure that the important technical pieces survived.

- The definition of the time step changed a little bit to be inverted from what you had in #385; please see my inline comment
- #385  had previously implemented had a [hard schedule start at `start_fraction = 0.725`](https://github.com/aqlaboratory/openfold-3/pull/385/changes#diff-053b3cd989f6df2c367f2e044320baab84c8d222e3556a8f352bd6dd6d942a63R53) With [the new Schedule definition](https://github.com/aqlaboratory/openfold-3/pull/395/changes#diff-59a8524cb142838d3fdaef063332300922f314255dd3a601b0b99b9a772f78e7R66), this is configurable through a runner yaml, but the wiring isn't plugged in yet. Just wanted to check if this start fraction was critical in your experiments, or if we could fix this to a constant like t=0.275 (with the inverted time definition) to help reduce complexity.

openfold3/steering/config.py
```diff
@@ -0,0 +1,108 @@
...
+    individual query, which also makes a steered/unsteered ablation a yaml
+    toggle rather than a second set of query JSONs.
+
+    Attributes:
+        enabled (bool):
+            Whether chemical steering runs at all. Off by default.
+        num_gd_steps (int):
+            Gradient-descent steps applied to the denoised coordinates at
+            each denoising step. With the default weight this bounds the
+            per-step correction to `num_gd_steps * weight` Angstrom.
```
peter-s-obi, line 73: @jnwei this is the docstring I am referring to in my second point

openfold3/steering/types.py
peter-s-obi: @jnwei The inverted time convention is fine with me. I used 0.725 in my experiments but didn’t compare different start fractions, so I can’t say that exact value is critical, likely a personal preference that crossed over from the pocket constraints code. A fixed late-only cutoff at 0.275 sounds reasonable for this PR; configurable schedules can come later.

On indexing, adding 1 was to convert python's counter into actual step number. Since step_index starts at zero, the old (tau + 1) / N corresponds to 1 - (step_index + 1) / N when counting down. The old formulation also included the cutoff itself (t <= 0.275). Omitting the +1 and using strict < can make steering start slightly later.

A small test around the cutoff, including the first active step and final step, could help confirm the intended timing.

PR #398, opened 2026-09-08
Fix documentation dependency and pocket configuration example

peter-s-obi:
Nothing major here, just a couple of small docs fixes especially during testing. Building the docs after installing from GitHub with [dev] was failing because sphinxcontrib-mermaid was missing, so I added it. Also fixed a typo in the pocket constraint YAML example.

Checked that the docs build and the corrected YAML validates.


jnwei (approved):
Thanks @peter-s-obi

jnwei (approved):

PR #403, opened 2026-09-10
chore: update pixi to 0.8.0

jandom:
**Summary**

Bump pixi version to 0.8.0, since the lock format change

**Changes**

- bumped min pixi version to 0.8.0 in pixi.toml this is debatable

**Related Issues**

**Testing**

**Other Notes**


jnwei (approved):
Does this mean that all users will have to do a pixi self-update in order to use openfold3? 

It's fine to do once in a while, but we should ensure not to make this process so frequent as to be disruptive

jandom:
> Does this mean that all users will have to do a pixi self-update in order to use openfold3?

Yeah, exactly but the pixi users are a very small bunch, – not even Christina is using it

PR #399, opened 2026-09-09
Add Colabfold msa server timeouts

jnwei:
**Summary**
The previous timeout gates for the colabfold server did not properly cause a timeout upon lack of response from an downed server. This action stalled two CI GPU runners for 109 minutes on 2026-09-08.

**Changes**
- In `colabfold_msa_server.py` Introduce an overall deadline for server responses / workflows, trigger an error if the deadline has passed or max number of retries is exceeded.
- Introduced unit tests to test the timeout behavior.
- Add a timeout to all GPU CI workloads of 120 min.
- Add polling to Colabfold server calls

**Testing**
- [x] All colabfold unit tests are passing
- [x] All integration and unit tests are passing.

PR #404, opened 2026-09-11
[CI/CD] Stop the AMD and EC2 runners sharing a parameter cache

jnwei:
**Summary**
Fix parameter cache corruption between the AMD and EC2 instance runners.

On AMD the path to the parameters is stored to @jandom's home directory. On EC2, the path to the parameters is `/home/ubuntu`. If AMD is run first, the cached path saved is incorrect relative to EC2, which caused the parameters not to be found. This failure mode was seen in [this nightly run](https://github.com/aqlaboratory/openfold-3/actions/runs/34558633289/job/103136993054)

**Changes**
- Remove cache restore step from AMD CI, it is unneeded as it is a persistent machine
- Change the cache name to include the default parameter name, introducing a new env variable. `OPENFOLD_DEFAULT_CKPT`
- Add a check of the parameters to ensure the size is correct, redownload if not.

**Testing**
- Run a manual test: https://github.com/aqlaboratory/openfold-3/actions/runs/34566940737

### DIFF
--- a/.github/workflows/ci-integration-test-pixi-amd-reusable.yml
+++ b/.github/workflows/ci-integration-test-pixi-amd-reusable.yml
@@ -17,6 +17,9 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}/openfold3-docker
+  # Keep in step with DEFAULT_CHECKPOINT_NAME in openfold3/entry_points/parameters.py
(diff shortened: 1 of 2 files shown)


jandom (approved):
This code is getting more and more wild :D

jnwei:
Unfortunately, the manually triggered CI hangs on the EC2 instances for 90+ min. It appears to be stuck in a PENDING state waiting for the colabfold result. 

I have a fix in #399, will continue discussing there.

PR #405, opened 2026-09-14
ci: fix broken integration tests

jandom:
**Summary**

Hand-over from @jnwei 

The nightly CUDA integration tests have been hanging until @jnwei  cancels them by hand (https://github.com/aqlaboratory/openfold-3/actions/runs/34566940737/job/103161316202). Docker's default 64 MB /dev/shm is too small for the dataloader to pass MSA batches back to the parent process — the allocation fails, multiprocessing swallows the error, and the job then waits forever for a batch that never arrives.

**Changes**

Gives both the CUDA and AMD test containers --shm-size=8g, and runs pytest with -s and a faulthandler timeout so any future hang prints something and dies instead of burning the runner. Also adds a 60-minute job timeout and tidies up a few leftover differences between the two workflows (parameter download, --rm, nightly schedule gate).

**Related Issues**

**Testing**

Launched the integration "suite" https://github.com/aqlaboratory/openfold-3/actions/runs/34843543097 all passed, CUDA and AMD

**Other Notes**

Potentially supersedes https://github.com/aqlaboratory/openfold-3/pull/399

After merging, re-enable nightly runs via https://github.com/aqlaboratory/openfold-3/settings/variables/actions


.github/workflows/ci-integration-test-pixi-cuda-reusable.yml
```diff
@@ -265,6 +265,7 @@ jobs:
     runs-on: ${{ fromJSON(needs.start-aws-runner.outputs.instances) }}
     needs:
       - start-aws-runner
+    timeout-minutes: 60
```
jnwei, line 268: nit: 60 min overall might not be long enough in the long run, I think a normal run can take 30-40 min right now. But maybe it's better to be conservative at first.

jnwei (approved):
Very nice! I'm glad that the fix is relatively simple once the correct issue was found.

.github/workflows/ci-integration-test-pixi-cuda-reusable.yml
jandom: yeah, i probably over-did it – but let's hope this holds for a while.

PR #407, opened 2026-09-14
fix: issue 406 custom template filenames dropped

### LINKED ISSUE #406: [BUG] CIF Direct Template - Template coordinates are NEVER used, fetched from RCSB or dropped altogether
**Describe the bug**
Hello and thank you for this awesome tool. While benchmarking some aspects of OF3, I stumbled accros a pretty nasty bug in the template processing in CIF-direct mode. A comment is present at a critical part of the upstream code `.../primitives/structure/template.py:269–274`, so the team might already have some knowledge of this. The present bug report hopefully gives a self contained and precise description of the issue. 
In CIF-direct mode (a query chain declares `template_cif_paths`), the provided CIF is used
only to compute the sequence alignment. The 3D coordinates fed to the model are **not** taken
from that file — they are loaded from `<structure_directory>/<stem>.cif`, where `<stem>` is the
provided CIF's **filename**, and that file is **downloaded from the PDB** (`fetch`) when absent.
Consequences:
1. **Edited / non-deposited templates are silently ignored.** If the provided CIF's coordinates
   differ from the current PDB entry (or the structure isn't in the PDB at all), the PDB version
   is used instead. The result looks correct only when the provided file happens to be byte-identical
   to the deposited entry.
2. **A template whose filename is not a valid PDB ID is silently dropped** — even though its
   coordinates and sequence are perfectly usable. The query then runs template-free with no error
   and no effect on the success/failure count.

This contradicts the code's own stated design: a dedicated CIF-direct coordinate branch exists and
preprocessing deliberately persists the CIF path into the cache — but a cache read-back bug throws
it away. There is an in-source `TODO` acknowledging the defect.

**To Reproduce**
Take any deposited mmCIF (here PDB `6TEL`) and make a copy under a non-PDB-ID name. The copy must
carry the same coordinates+sequence but **distinct file bytes** — the template cache key hashes the
CIF *content* (`get_file_content_hash`), so a byte-identical copy is deduplicated against `6TEL.cif`
and silently inherits its (successful) result instead of being processed on its own. Relabelling the
internal entry id (coordinates untouched) is enough:

```bash
cp 6TEL.cif templates/cifs/6TEL.cif
# same coords/sequence, non-PDB-id filename, distinct content so it is processed on its own:
python -c "import gemmi; d=gemmi.cif.read('templates/cifs/6TEL.cif'); b=d.sole_block(); b.name='6TEL_notpdb'; b.set_pair('_entry.id','6TEL_notpdb'); d.write_file('templates/cifs/6TEL_notpdb.cif')"
```

`queries.json` — two queries, same sequence, same structure, differing only in the template
filename (single protein chain, no MSA, to isolate template handling):

```json
{
  "queries": {
    "hasPdbName": {
      "chains": [
        {
          "molecule_type": "protein",
          "chain_ids": ["A"],
          "sequence": "GPGEKLELRLKSPVGAEPAVYPWPLPVYDKHHDAAHEIIETIRWVCEEIPDLKLAMENYVLIDYDTKSFESMQRLCDKYNRAIDSIHQLWKGTTQPMKLNTRPSTGLLRHILQQVYNHSVTDPEKLNNYEPFSPEVYGETSFDLVAQMIDEIKMTDDDLFVDLGSGVGQVVLQVAAATNCKHHYGVEKADIPAKYAETMDREFRKWMKWYGKKHAEYTLERGDFLSEEWRERIANTSVIFVNNFAFGPEVDHQLKERFANMKEGGRIVSSKPFAPLNFRINSRNLSDIGTIMRVVELSPLKGSVSWTGKPVSYYLHTIDRTILENYFSSLKNPG",
          "template_cif_paths": ["templates/cifs/6TEL.cif"]
        }
      ]
    },
    "nonPdbName": {
      "chains": [
        {
          "molecule_type": "protein",
          "chain_ids": ["A"],
          "sequence": "GPGEKLELRLKSPVGAEPAVYPWPLPVYDKHHDAAHEIIETIRWVCEEIPDLKLAMENYVLIDYDTKSFESMQRLCDKYNRAIDSIHQLWKGTTQPMKLNTRPSTGLLRHILQQVYNHSVTDPEKLNNYEPFSPEVYGETSFDLVAQMIDEIKMTDDDLFVDLGSGVGQVVLQVAAATNCKHHYGVEKADIPAKYAETMDREFRKWMKWYGKKHAEYTLERGDFLSEEWRERIANTSVIFVNNFAFGPEVDHQLKERFANMKEGGRIVSSKPFAPLNFRINSRNLSDIGTIMRVVELSPLKGSVSWTGKPVSYYLHTIDRTILENYFSSLKNPG",
          "template_cif_paths": ["templates/cifs/6TEL_notpdb.cif"]
        }
      ]
    }
  }
}
```

Run:

```bash
run_openfold predict --query-json queries.json --use-templates true --use-msa-server false
```

**Expected behavior**
- **Expected:** in CIF-direct mode the coordinates come from the provided CIF, no network fetch is
  required, the filename is irrelevant to whether the template is usable.
- **Actual:** coordinates come from a PDB download keyed on the filename stem, a non-PDB-ID filename
  drops the template entirely, the user provided template coordinates are never used (either droped or replaced by RCSB PDB ones). 

**Qualitative and Quantitative Bug Confirmation**
Hints of the issue were found in the template preprocessing logs. Observed in the template preprocessing log (`<of3_tmpdir>/template_data/template_logs/*.log`):
```
Selected chain A from 6TEL (seq_id=1.000, q_cov=1.000)
Structure not available, fetching 6TEL.
Loading structure .../template_structures/6TEL.cif           # coordinates come from the PDB download
6TEL A added to cache.  ->  Found 1 hits.

Selected chain A from 6TEL_notpdb (seq_id=1.000, q_cov=1.000)
Structure not available, fetching 6TEL_notpdb.
 6TEL_notpdb is not a valid PDB ID. Skipping this template.  # dropped despite valid coordinates
Found no hits.
```
Note that as is, the logs are not complete, and some patching was made to write them in their entirety. Let me know if this could be useful for me to share the patching done, so you can investigate. 

To confirm the bug was real, I computed the RMSD against the ground truth for four setups: no provided template path, template path matching the PDB id, template path not matching the PDB id (but coordinates identical to the PDB deposited one), template path matching PDB but with coordinate heavily perturbed (so should fail to give a good structure if the coordinates where actually used). 

| case | template file (stem) | mean Cα-RMSD vs 6TEL (Å) | avg_plddt | ptm | template log |
|---|---|---|---|---|---|
| `no_template` | — | 20.07 ± 1.08 | 33.8 | 0.19 | no template |
| `template_pdbname` | `6TEL` (valid id) | **0.14 ± 0.01** | 89.2 | 0.93 | `fetching 6TEL` → `Found 1 hits` |
| `template_nonpdbname` | `6TEL_notpdb` (not an id) | 20.14 ± 1.37 | 33.8 | 0.19 | `6TEL_notpdb is not a valid PDB ID. Skipping` → `Found no hits` |
| `template_pdbname_perturbed` | `6TEL` (valid id), coords += 10 Å noise | **0.15 ± 0.01** | 89.2 | 0.93 | `Structure file: True` → loads fetched `6TEL.cif`, **no re-fetch** → `Found 1 hits` |

Two conclusions, each a ~20 Å vs ~0.15 Å separation:
- **`template_pdbname_perturbed` ≈ `template_pdbname` (0.15 vs 0.14 Å).** The perturbed file's
  coordinates are 16 Å RMSD from native, yet the prediction is indistinguishable from the
  clean-template run — the provided coordinates were ignored and the deposited 6TEL was used. The
  template log confirms it loaded `template_structures/6TEL.cif` (the copy fetched for the clean case)
  with `Structure file: True` and **no second fetch**. Direct measurement of **Defect 1**.
- **`template_nonpdbname` ≈ `no_template` (20.14 vs 20.07 Å).** Renaming an otherwise-valid template
  to a non-PDB-id filename reduces the prediction to the no-template baseline — silently dropped
  (`Found no hits`, no cache entry written). **Defect 2.**

 **Fuchsia: template name is proper pdb id, cyan: template name is arbitrary (but template is exactly pdb coordinates)** -> should be close and they are not
blue_nonpdbname_purple_pdbname 
**Fuschia: template name is proper pdb id, yellow: template name is proper pdb id, but coordinate were perturbed heavily** -> should be different and they are very similar
purple_pdbname_yellow_pdbnameperturbed

 - OS: Ubuntu 24.04.4 LTS Noble
 - GPU used and memory available: NVIDIA A10G, 32GB RAM
 - Python version (if relevant): 3.11.15
 - Installation path: pip

**Additional context**
Related PR and issues: #343, #294, #235.

jandom:
**Summary**

Fixes https://github.com/aqlaboratory/openfold-3/issues/406 starting with the tests

**Changes**

**Related Issues**

**Testing**

**Other Notes**


ahmedtaha100:
@jandom any help I can provide here?

aqemia-antonin-brossollet:
Hello @jandom, let me know if I can also help by testing against the same setup in the bug report.

jandom:
Thank you both for reaching out, I could definitely use some help 
-  @aqemia-antonin-brossollet if you could check this branch on your case, it should be fixed
- @ahmedtaha100 i'm not pleased with my refactored code structure, it's not like it was great before, it's just overly complicated for what should be 3-4 distinct behaviors (and not a see of if...else) – if you have a way to improve this, i'm all open to suggestions. happy to give you write access to the repo, or accept a cleanup PR from a fork

aqemia-antonin-brossollet:
I checked and indeed it seems to do the trick!! Good job! I also verified that no RCSB call was made with the fix, which is the right behavior. 


| case | buggy (`main`/0.5.0) | fixed (`issue-406`) | meaning |
|---|---|---|---|
| `no_template` | 20.07 | 20.09 | baseline, unchanged |
| `template_pdbname` | 0.15 | 0.15 | valid-name template still applied |
| `template_nonpdbname` | 20.12 (dropped) | **0.14 (applied)** | Defect 2 fixed — non-PDB filename no longer dropped |
| `template_pdbname_perturbed` | 0.15 (ignored) | **15.42 ± 2.84 (used)** | Defect 1 fixed — provided coordinates now honored (a bad template now degrades the prediction) |

PR #408, opened 2026-09-14
Avoiding duplicating the diffusion module

chelsea0x3b:
**Summary & Changes**

Using `setattr` to avoid registering diffusion module as a submodule of `SampleDiffusion`. This saves `~750MB` during training because of the EMA copies.

**Testing**

Ran the same command twice, recorded memory saved with:
```python
diff --git a/openfold3/projects/of3_all_atom/runner.py b/openfold3/projects/of3_all_atom/runner.py
index 0e1836e6..426099c5 100644
--- a/openfold3/projects/of3_all_atom/runner.py
+++ b/openfold3/projects/of3_all_atom/runner.py
@@ -794,6 +794,17 @@ class OpenFold3AllAtom(ModelRunner):
                 step=self.global_step - 1,
             )
 
+        if torch.cuda.is_available():
+            gib = 2**30
+            with open(f"mem-probe.{self.global_rank}.txt", "a") as fp:
+                fp.write(
+                    f"[peak-mem] step {self.global_step} rank {self.global_rank} | "
+                    f"alloc peak {torch.cuda.max_memory_allocated() / gib:.3f} GiB | "
+                    f"reserved peak {torch.cuda.max_memory_reserved() / gib:.3f} GiB | "
+                    f"alloc now {torch.cuda.memory_allocated() / gib:.3f} GiB\n",
+                )
+            torch.cuda.reset_peak_memory_stats()
+
     def _get_train_sampler(self):
         dl = self.trainer.train_dataloader
         # If PL uses multiple loaders, dl can be CombinedLoader etc.
```

**Other Notes**


jandom (commented):
@chelsea0x3b this is super cool, and saved a lot of memory – but I have to be honest, I don't love the __setattr__ – if this pattern was to spread around the repo, it'll make it a nightmare. Is there any other, more readable way we could eat our cake and keep it too?

PR #411, opened 2026-09-14
feauter (cli): add query validation command 

samanyuckulkarni123:
**Summary**
- adds a lightweight ommand ofr validating and summarizing inference query JSON files before running inference. 
**Changes**
- added run_openfold check-query --query-json <path>.
- reused the existing inferencequeryset schema used by predict
- Added text and JSON summary output using --format
- added readable validation errors and non-0 exit codes for invalid inputs
- added CLI tests
- documented the command in the input format and inference guides 
**Related Issues**
N/A
**Testing**
- Passed the Ruff formatting and lint checks
- All 9 focused CLI tests passed
- Manual text and JSON output checks passed
- Full suite: 581 passed, 61 baseline failures, 50 skipped 
**Other Notes**
The command does not initialize a GPU, load model parameters, contact an MSA server, or create inference outputs. It checks the query schema and supplied paths but does not parse MSA or template contents or guarantee that downstream inference will succeed.


jnwei:
Hi @samanyuckulkarni123, thanks for putting this together, and for including tests and docs.

After reviewing, we've decided not to merge this, so I'm going to close it. A few reasons:

- A dedicated query checker feels like more than we need. The query is already validated at the start of inference, which can be checked without using a gpu using the `openfold3-base` pixi environment.
- We'd prefer not to add a new subcommand to `run_openfold` for this. If we wanted a standalone check, it would be a small script or skill rather than part of the main CLI.

Thanks again for your interest in OpenFold3. For future contributions, opening an issue first is a good way to align on scope before putting in the work.

PR #412, opened 2026-09-15
Enable batch size > 1

chelsea0x3b:
**Summary**
Enables running with batch size higher than 1. This is possible **only during training** as training data loader already crops & pads things to the same size. This is not the case for validation - so validation must still use batch size 1.

**Changes**

1. Add a separate `batch_size_val` to DataModuleConfig
2. Update `get_metrics` and `get_metrics_chunked` in `metrics/quality.py` to handle batches with batch size > 1. Just process each item individually.
3. Log mean values of tensors

**Testing**

Ran with & without cueq kernels enabled. Running cueq kernels requires passing `--use_tf32 1` to the run command.

**Other Notes**


chelsea0x3b:
CLosing - discussed other avenues and this is partial work

PR #414, opened 2026-09-16
Sets `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` by default

chelsea0x3b:
**Summary & changes**
Setting an allocator flag. see [pytorch docs](https://docs.pytorch.org/docs/2.14/notes/cuda.html#optimizing-memory-usage-with-pytorch-alloc-conf) for how it works.

This saves a lot of peak reserved memory. At token_budget 384 it reduced the peak memory reserved by 3GB. At higher token_budgets, the savings are more!
**Related Issues**
N/A

**Testing**
Ran with a testing config before & after.

**Other Notes**

captured memory metrics with 
```diff --git a/openfold3/projects/of3_all_atom/runner.py b/openfold3/projects/of3_all_atom/runner.py
index 0e1836e6..426099c5 100644
--- a/openfold3/projects/of3_all_atom/runner.py
+++ b/openfold3/projects/of3_all_atom/runner.py
@@ -794,6 +794,17 @@ class OpenFold3AllAtom(ModelRunner):
                 step=self.global_step - 1,
             )
 
+        if torch.cuda.is_available():
+            gib = 2**30
+            with open(f"mem-probe.{self.global_rank}.txt", "a") as fp:
+                fp.write(
+                    f"[peak-mem] step {self.global_step} rank {self.global_rank} | "
+                    f"alloc peak {torch.cuda.max_memory_allocated() / gib:.3f} GiB | "
+                    f"reserved peak {torch.cuda.max_memory_reserved() / gib:.3f} GiB | "
+                    f"alloc now {torch.cuda.memory_allocated() / gib:.3f} GiB\n",
+                )
+            torch.cuda.reset_peak_memory_stats()
+
     def _get_train_sampler(self):
         dl = self.trainer.train_dataloader
         # If PL uses multiple loaders, dl can be CombinedLoader etc.
```


openfold3/run_openfold.py
```diff
@@ -25,6 +25,8 @@
 
 import click
 
+os.environ.setdefault("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True")
```
jandom, line 28: A cleaner way would be set this up inside `_configure_torch_backend`, maybe via some config.yaml field that finds its way here - if that's technically possible?

jnwei: IMO, it's fine to set this as an environment level default, but with a few caveats:

- I would place this in https://github.com/aqlaboratory/openfold-3/blob/main/openfold3/entry_points/import_utils.py, as @jandom suggested.

- We need to be careful of handling with AMD GPUs. In particular, I'd recommend adding a flag to ensure that the HIP/Triton behavior is respected first

```
if "PYTORCH_HIP_ALLOC_CONF" not in os.environ:
    os.environ.setdefault("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True")
```



- As of torch=2.10 is typically used, `PYTORCH_ALLOC_CONF`; using `PYTORCH_CUDA_ALLOC_CONF` might trigger a deprecation warning for pytorch=2.9, but we cannot just use `PYTORCH_ALLOC_CONF` without pinning the pytorch version.

I'd also add a logger.info statement if one isn't output by torch already so that power users who are tweaking the cuda settings know that this is being set.

jnwei (changes requested):
Cross-verified that expandable_segments is also used in the anthropic uplifiting-biomolecular-modeling repo. https://github.com/anthropics/uplifting-biomolecular-modeling/blob/f4f62fa6592ae4938d49b1757bea0cfeff9f468e/common/opt_core/opt_core/mem/allocator.py#L177

Paging @singagan for suggestions re best AMD handling / testing when using the PYTORCH_CUDA_ALLOC_CONF, please see in thread.

chelsea0x3b:
@jnwei updated!

jandom:
Here is my results table, everything replicates on GB10: allocated unchanged but reserved reduces nicely

```
step tokens |   allocated MiB |    reserved MiB | inactive split MiB | reserved/alloc
            |     off      on |     off      on |      off        on |    off      on
   0    128 |    3493    3481 |    4116    3830 |      923         0 |  1.18x   1.10x
   1    256 |    8086    8061 |    9220    8450 |     1602         0 |  1.14x   1.05x
   2    160 |    4633    4614 |    9222    8450 |     1783         0 |  1.99x   1.83x
   3    320 |   10343   10322 |   12388   10790 |     3181         0 |  1.20x   1.05x
   4    192 |    5671    5644 |   12388   10790 |     1673         0 |  2.18x   1.91x
   5    320 |   10550   10529 |   14202   11070 |     3142         0 |  1.35x   1.05x
```

@chelsea0x3b please take a look – because i've made fairly substantial changes, mutualized with similar pieces of our codebase (memory metrics helpers used previously in diffusion_loss sparse vs dense)

PR #419, opened 2026-09-18
Reduce memory use computing confidence scores

tomgoddard:
**Summary**

One line change to code that computes all-by-all atom distances used in computing confidence scores.  The current OpenFold 3 code creates large temporary arrays of size (N_atom, N_atom, 3) that cause out of memory errors.  There is a built-in Torch function cdist() that does exactly this without temporary arrays avoiding out-of-memory errors.

**Changes**

Changed 

    d = torch.sum(eps + (x[..., None, :] - x[..., None, :, :]) ** 2, dim=-1) ** 0.5

to

    d = torch.cdist(x, x, p=2.0)

The eps (epsilon = 1e-8) in the original expression avoids divide by zero in gradient back propagation when two atoms have exactly the same coordinates.  The Torch cdist() function is documented to correctly handle this gradient singularity (returning 0 gradient).  The cdist() function has been in Torch since version 1.2 (2019).

**Related Issues**

**Testing**

I made this change after a 3700 amino acid prediction with OpenFold 0.5 ran out of memory on this line of code on an Nvidia RTX Pro 6000 with 96 GBytes.  After the change to use cdist() the prediction completes successfully.  I've attached the input json file and the out-of-memory error where it failed allocating one of the (N_atom, N_atom, 3) arrays of size 8.62 GB (N_atom = 27778).

**Other Notes**

[confidence_oom.zip](https://github.com/user-attachments/files/32397795/confidence_oom.zip)


jnwei:
@christinaflo Could you have a look at this? I vaguely recall there are some parts where we intentionally do not use cdist, but I don't remember where at this point.

PR #422, opened 2026-09-19
Feature/add megafold sng attn

bluecake-dh:
**Summary**
Similar to cuEquivariance, DeepSpeed, and the Triton triangle kernels, MegaFold offers similar memory/speed efficiencies. I specifically added its EvoFlash-3D Single Attention with Pair Bias to the existing model. 

**Changes**
Like the previously mentioned examples, I have added appropriate flags that are passed down the call stack of functions, and I have written a corresponding attention function for the MegaFold option. I also created files that contain the necessary source code from MegaFold as they do not provide packages.

**Other Notes**
I was working with Etowah Adams over the summer as an intern, and this was the final task I was working on before my summer vacation ended. This work is untested and experimental. Upon completion, it seemed functional, but after merging changes from the main branch, the program throws an exception, saying "Choose at most one alternative attention algorithm." I determined that the use_high_precision flag is also enabled, though I am not sure why because I do not run into this issue when I try using other attention options. There may very well be other issues. I apologize for the sloppy work, but since university is starting I do not know if I will be able to continue working on this issue. Therefore, I just wanted to share my current status, and perhaps leave a checkpoint if someone wants to pick up this issue.


jnwei:
Thank you for sharing @bluecake-dh . I'm going to convert this PR to a draft for now as you indicated this work is not yet complete.

Issue #420, opened 2026-09-18
[BUG] Templates in inference_query_set.json are ignored


tomgoddard:
**Describe the bug**

After running a query the inference_query_set.json output contains "template_alignment_file_path" and "template_entry_chain_ids" pointing to the templates found in MSA processing using the Colabfold server.  If a second query is run with this inference_query_set.json as input then the templates are ignored.  This is with OpenFold 3 code from Sept 18, 2026 (commit 7de748b7bc93adb5af0a4032d5e9f208b6e1325f) and using a runner.yaml file that preserves the MSAs and templates.

Looking at the OpenFold 3 code it appears it no longer correctly handles the cached template alignment file in .npz format given in the inference_query_set.json as "template_alignment_file_path", and in fact it silently discards the templates.

The fact that the templates are not used in the second prediction is seen from looking at the second prediction inference_query_set.json which has "template_alignment_file_path": null.  Adding a debugging print statement to OpenFold 3 code confirms that no templates are featurized in the second prediction.

**To Reproduce**

I've attached a zip file containing an input json file, a runner.yaml file to preserve MSA and templates, and a list of commands to run the prediction twice to reproduce this bug.

**Expected behavior**

The second predition should use the templates specified in the inference_query_set.json of the first prediction and the inference_query_set.json from the second prediction should have "template_alignment_file_path" correctly set to the same path as the first prediction (not null).

**Stack trace**

No stack trace

**Configuration (please complete the following information):**
 - OS: macOS 26.6.2
 - GPU and memory: Mac M2 Ultra GPU, 64 GB memory
 - Python version (if relevant): 3.12 (not relevant)
 - Installation path (e.g. conda / pip / build from repo): pip install from cloned github repository

**Additional context**

It looks like this bug was introduced in when the template alignment .npz file name was changed to include both a hash of the query sequence and the original alignment (.m8) file.

Using "template_alignment_file_path" with a .npz file worked in OpenFold 3 preview 2.  ChimeraX was using cached templates by specifying this path to the processed template alignment .npz file and I found this problem when trying to update ChimeraX to use OpenFold 0.5.

[8rf4_test_case.zip](https://github.com/user-attachments/files/32399291/8rf4_test_case.zip)

[8rf4_results.zip](https://github.com/user-attachments/files/32399300/8rf4_results.zip)

jnwei:
Thank you for the report @tomgoddard sorry for the oversight here.

[...]

PR #425, opened 2026-09-22
Fix cached template reuse in inference query sets

ahmedtaha100:
@jnwei 
Reusing `inference_query_set.json` could discard cached templates because the preprocessor treated `.npz` entries as raw alignments and rebuilt their cache keys. Preserve their declared paths and template IDs through parsing and write-back. Addresses [#420](https://github.com/aqlaboratory/openfold-3/issues/420).

One offline regression uses existing fixtures and a JSON round trip, mixing a cached chain with a fresh raw input to exercise both guards. It fails without the fix; removing either guard also fails it.

Validation:

- Affected module: **54 passed**. Ruff and mypy checks passed.
- Full CPU run: 1,015 passed; two local packaging failures resolved by editable installation, with 59 affected tests passing afterward.
- GPU selections: 62 passed, 33 backend/data skips, one bf16 tolerance failure reproduced identically on unchanged upstream.

GPU validation used an RTX 5090 with CUDA 12.9.


openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -930,3 +930,57 @@ def test_template_sources_do_not_collide_across_queries(tmp_path, order):
...
+
+def test_requeued_query_set_keeps_cached_templates(tmp_path):
+    """A reused query set keeps its cached template entry (issue #420).
+
+    Run 1 preprocesses a ColabFold alignment and writes the cache entry path and
+    template ids into the chain. The reused batch mixes that chain, reloaded from
+    the run's serialized query set, with a raw CIF source that run 1 never
+    processed, so the write-back still runs while the cached entry is bypassed.
+    """
+    run_1 = _two_source_query_set(tmp_path, ["q_colabfold"])
```
jnwei, line 943: nit could we use a consistent naming of the variables throughout this test? In particular:

`run_1`, `requeued` are all query_sets, so maybe a naming like `run_1_qs` and `run_2_qs` might help with readability.

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -930,3 +930,57 @@ def test_template_sources_do_not_collide_across_queries(tmp_path, order):
...
+    cached_path = cached_chain.template_alignment_file_path
+    assert cached_path is not None
+    assert cached_chain.template_entry_chain_ids == ["2q2k_C"]
+
+    requeue_path = _write_file(
+        tmp_path / "inference_query_set.json", run_1.model_dump_json()
+    )
+    requeued = InferenceQuerySet.from_json(requeue_path)
+    fresh = _two_source_query_set(tmp_path, ["q_custom"])
+    requeued.queries["q_custom"] = fresh.queries["q_custom"]
```
jnwei, line 968: nit: Perhaps this could be an update of the queries to make the relationship a bit more explicitly, like `requeued.queries |= fresh.queries`

openfold3/tests/core/data/pipelines/preprocessing/test_template.py
```diff
@@ -930,3 +930,57 @@ def test_template_sources_do_not_collide_across_queries(tmp_path, order):
...
+    TemplatePreprocessor(input_set=run_1, config=settings)()
+
+    cached_chain = run_1.queries["q_colabfold"].chains[0]
+    cached_path = cached_chain.template_alignment_file_path
+    assert cached_path is not None
+    assert cached_chain.template_entry_chain_ids == ["2q2k_C"]
+
+    requeue_path = _write_file(
+        tmp_path / "inference_query_set.json", run_1.model_dump_json()
+    )
```
jnwei, line 965: nit: We could define requeued as a deepcopy of the existing query set model to avoid the file dump of the inference_query json.

```
requeued_qs = run_1_qs.model_copy(deep=True)
```

jnwei (approved):
LGTM! Thank you for the quick and simple fix @ahmedtaha100 

Just a couple of small nits but I think this is mostly ready to go.

